PR #31, opened 2026-07-08
fix(terminal): stop severing wide Nerd Font prompt glyphs (#17)

### LINKED ISSUE #17: shell prompt rendering bug
### Type

### What's going on?

zsh, the shell prompt can not show properly

Image

### tty7 version (bugs)

v0.5.0

### Platform (bugs)

macOS (Apple Silicon)

### Anything else?

_No response_

l0ng-ai:
## What

Follow-up to #19 for issue #17. Nerd Font prompt icons — and the multiline `➜`/`❯` in the original report — still rendered **sliced off on the right**.

## Root cause

`paint_glyphs` clips every single-column (`Solo`) glyph to **exactly one cell** via a `content_mask`. But a symbol face routinely draws ink *past* that cell:

- A **non-Mono Nerd Font** sets a **one-cell advance** on its icons yet fills up to **~1.9 cells** of ink. Measured on the reporter's font class (`fontTools` bounding boxes, Hasklug Nerd Font Regular, cell = 600 units):
  - `folder`  ink xMax = **1.87×** cell
  - `f303` / gitbranch = **1.67×**
  - `cog` = 1.57×, `save` = 1.46×, `bolt`/`apple` = 1.2–1.25×
- The **OS cascade** serves a proportional `➜` (U+279C) the same way when nothing in the font list covers it — the "severed arrow" in the issue screenshot.

`#19` only swapped *which face* resolves a glyph (bundled Hack fallback); it never touched the clip, so a wide glyph coming from the **primary** font or a wide fallback was still cut. Advance is no signal here — it reads one cell for exactly the icons that overflow.

## Fix

Widen a solo glyph's paint window to **two cells**. A glyph that already fits is untouched (no ink to spill); a symbol that overflows renders **whole**, bleeding into the trailing blank the way iTerm2 / Terminal.app do with non-Mono faces. Bounded at two cells so a pathological face can't smear a lone glyph across the row. Batched `Run`/`Wide` segments keep their exact column-span clip; **Mono Nerd Fonts are pixel-identical**.

```rust
fn solo_clip_width(cell_width: Pixels) -> Pixels { cell_width * 2. }
```

## Verification

Built and driven in-app on Apple Silicon with a non-Mono **"Hasklug Nerd Font"** primary (worst case):

- **Before:** apple / folder / house / cog icons all cut with a hard vertical edge one cell in; pixel-diff of the render before/after my first (advance-based) attempt was empty — confirming advance is the wrong signal.
- **After:** every icon renders whole; a colored powerline prompt (segments + native `E0B0` separators + in-segment apple/clock icons) renders cleanly; **ASCII, CJK, and the Mono-font path are unchanged**.

New unit test `solo_clip_width_gives_symbols_two_cells`; full suite green (390 passed).

Closes #17.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Nerd Font prompt icons and similar single-cell symbols from being clipped or cut off on the right.
  * Improved glyph rendering so icons with slight overflow now display fully without bleeding into neighboring text.
  * Preserved correct rendering for special separator symbols while applying the new clipping behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Nerd Font prompt icons no longer render sliced off on the right. A non-Mono
+  Nerd Font (and the proportional `➜`/`❯` the OS cascade hands back when nothing
+  in your font list covers them) gives an icon a single-cell *advance* but draws
+  ink up to ~1.9 cells wide, and tty7 clipped every lone glyph to exactly one
+  cell — severing the overflow into the half-icons and cut-off arrow from the
+  report. A single glyph now paints into a two-cell window, so it renders whole
+  (bleeding into the trailing blank the way iTerm2 and Terminal.app do), bounded
+  at two cells so a stray face can't smear across the row. Pairs with the native
+  powerline separators from #19; Mono Nerd Fonts are unchanged. (#17)
+
 - New tabs and splits no longer stall for seconds while a zsh plugin manager
   reinstalls itself. tty7 launches zsh through a throwaway `ZDOTDIR` (so it can
   layer its shell integration on top of your config), but it used to leave
--- a/src/terminal/element.rs
+++ b/src/terminal/element.rs
@@ -687,6 +687,26 @@ fn powerline_path(bounds: Bounds<Pixels>, shape: PowerlineShape) -> gpui::Path<P
     }
 }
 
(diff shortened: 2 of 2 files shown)


l0ng-ai:
Addressed the review nitpick in d5fb809: folded the clip-width decision into a single pure `seg_clip_width(solo, cells, cell_width)` so both branches are unit-tested — a batched Run/Wide segment clips to its exact column span, a solo glyph gets the two-cell window — instead of only asserting the standalone constant. No behavior change.

PR #47, opened 2026-07-10
feat(view): explain a dead Ctrl+R instead of failing silently

### LINKED ISSUE #46: Shell integration silently dies under figterm-style PTY shims (kiro-cli-term): no ghost text, no Tab menu, no Ctrl+R menu
## Symptom

On a machine with Kiro CLI (the Fig / Amazon Q CLI lineage) shell integration installed, the entire command-editor overlay silently disappears: no inline ghost text, no Tab completion menu, no Ctrl+R history menu (#45). Keystrokes fall through raw to the PTY, so Ctrl+R shows zsh's own `bck-i-search:` line instead of tty7's `(reverse-i-search)` menu. There is no error or diagnostic anywhere — it just looks like the features don't exist.

Reproduced on macOS with tty7 v0.9.0 (release dmg) and Kiro CLI 2.1.1.

## Root cause

Kiro's `zshrc.pre.zsh` block (installed at the top of `~/.zshrc`) hands the session over to `kiro-cli-term`, a figterm-style shim that interposes a **nested PTY** between the terminal and the user's real shell:

```
tty7 --daemon
└── zsh (kiro-cli-term)        ttys005   ← the shell tty7 spawned, taken over by the shim
    └── /bin/zsh --login       ttys007   ← the shell the user actually types into
```

tty7's shell integration relies on OSC 133 prompt reports reaching the daemon (`ShellState.active` / `at_prompt`, `src/terminal/remote.rs`). The shim swallows those sequences instead of forwarding them from the inner PTY, so the daemon never sees a single prompt report, `input_active()` stays false, and the whole overlay never renders (`src/terminal/view.rs` gates the input bar, completion menu, and reverse-search menu on it).

The failure also stops tty7's own history file from being written (its mtime freezes), since records are appended on OSC 133;D.

Confirmed by A/B test with the same binary and same zshrc:

- launched with `Q_TERM`/`QTERM_SESSION_ID` present in the env (Kiro thinks it's already injected and skips the shim) → everything works, Ctrl+R menu pops up;
- launched with a clean env (like Finder/Dock) → shell shows up as `zsh (kiro-cli-term)`, overlay dead.

## Why this is nasty

- Fig / Amazon Q / Kiro / Codewhisperer shims are fairly common on dev machines, and they're injected from the *user's* rc files, so tty7's ZDOTDIR redirectors run first and everything looks correctly installed.
- The feature degrades to "nothing", with zero signal. A user upgrading for the Ctrl+R menu (release notes headline feature) just concludes the release is broken.

## Ideas

1. **Detect and surface it.** The daemon already knows integration never engaged (`shell_active()` false after the first prompt would normally have reported). If the spawned shell's process title matches a known shim (`kiro-cli-term`, `figterm`, `qterm`, …) or N seconds pass with output but no OSC 133, show a one-line hint ("shell integration inactive — a terminal shim (kiro-cli-term) is intercepting it") in the pane or settings.
2. **Survive the wrapper.** The inner shell inherits tty7's `ZDOTDIR` redirectors and does emit OSC 133 — into the inner PTY. Forwarding can't be fixed from our side, but documenting the workaround helps: guard the Kiro block with tty7's marker env, e.g.

   ```zsh
   if [[ -z "$TTY7_USER_ZDOTDIR" ]]; then
     builtin source ".../zshrc.pre.zsh"
   fi
   ```

   or `kiro-cli integrations uninstall dotfiles`. (Note: Kiro's own uninstaller only removed its `.zprofile` block here; the `.zshrc` block had to be removed by hand.)
3. Possibly report upstream to Kiro CLI that OSC 133 A/B/C/D should be forwarded through the shim — iTerm2/WezTerm/Kitty semantic-prompt users hit the same wall.

## Workaround (verified)

Remove/guard the Kiro CLI pre/post blocks in `~/.zshrc`, open a new tab → shell is plain `-zsh`, overlay and Ctrl+R menu work.

l0ng-ai:
Closes #46.

## What

When shell integration never engages in a pane, pressing Ctrl+R now raises a one-shot floating notice (bottom-right) explaining that inline completion and the history menu are unavailable — and names the culprit when the daemon's foreground-process query matches a known figterm-style PTY shim (`kiro-cli-term`, `figterm`, `qterm`, `cwterm`). The chord still reaches the PTY, so the shell's own `bck-i-search`/reverse-i-search keeps working as the fallback.

## Why

A fig/Q/Kiro shim installed in the user's rc files execs over the shell tty7 spawns and re-hosts it on a nested PTY without forwarding OSC 133. The daemon then never sees a prompt report, `input_active()` stays false, and the entire command-editor overlay (ghost text, Tab menu, Ctrl+R menu) silently vanishes — the v0.9.0 headline feature just looks broken, with zero signal anywhere. See #46 for the full investigation.

## How it stays honest

- **Silent when the raw path is expected**: integration engaged and a foreground command owns the PTY; a full-screen TUI on the alt screen; or inside an 8s startup grace window (slow rc files legitimately haven't reported yet).
- **Named blame only on a known shim**: an ordinary foreground process (`ssh`) is never accused of intercepting anything — unknown names get the generic message. The refinement runs off-thread (`list_panes` control connection, best-effort) after the generic message is already up.
- **Self-correcting**: retracted if a slow shell engages late (the 300ms foreground poll notices `shell_active()` flip), dismissed by the next keystroke or a 15s timeout, and latched once per pane — a diagnostic, not a nag.

Zero protocol/daemon changes; everything is client-side off state the client already has.

## Tests

- `ctrl_r_without_integration_raises_the_notice_once` — through the real `on_key_down` dispatcher: grace window stays silent, notice raised after it, survives a real render pass (`run_until_parked` draws the frame), dismissed by the next keystroke, one-shot per pane.
- `shim_detection_names_known_wrappers_only` — `zsh (kiro-cli-term)` → named; `ssh`/`wezterm` → generic; generic message never claims interception.
- Full suite: 428 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- When shell integration never engages in a pane, pressing Ctrl+R now explains
+  why the history menu can't appear (once per pane, dismissed by the next
+  keystroke) instead of failing silently — naming the wrapper when a
+  figterm-style PTY shim (`kiro-cli-term`, `figterm`, `qterm`) is intercepting
+  the shell's OSC 133 reports. The chord still reaches the shell, so its own
+  reverse-i-search keeps working. (#46)
+
 ## [0.9.0] - 2026-07-10
 
 ### Changed
--- a/src/terminal/view.rs
+++ b/src/terminal/view.rs
@@ -224,6 +224,19 @@ pub struct TerminalView {
     /// selection into the line, Cmd+Enter runs it outright, and Escape/Ctrl+G
     /// cancels.
     reverse_search: Option<ReverseSearch>,
+    /// One-shot "shell integration didn't engage" notice (#46). Set when Ctrl+R
+    /// is pressed in a pane whose shell never reported OSC 133 — the history
+    /// menu the user is reaching for can't appear, and without this the feature
+    /// just looks broken. A figterm-style PTY shim (kiro-cli-term, qterm) that
+    /// swallowed the reports is the usual culprit, so the message names the
+    /// wrapper when the daemon's foreground query recognizes one. Cleared on
+    /// the next keystroke, after a timeout, or if integration engages late.
+    integration_notice: Option<String>,
+    /// Latch so the notice shows at most once per pane — a diagnostic, not a nag.
+    integration_notice_shown: bool,
+    /// When this view was created. Ctrl+R inside the startup grace window stays
+    /// silent: slow rc files mean integration legitimately hasn't reported yet.
+    created_at: std::time::Instant,
     /// True while a left-drag that began on the command-editor line is in progress,
     /// so mouse-move extends the editor selection rather than the terminal's.
     editor_selecting: bool,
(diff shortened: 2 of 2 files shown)


l0ng-ai:
Superseded administratively: this branch's entire diff — known_pty_shim / integration_notice_message / note_integration_gap / render_integration_notice plus both tests and the CHANGELOG entry — landed on main inside the #48 squash (0ee126a), so rebasing this branch onto main drops its commit as empty. The feature ships in v0.10.0. Closing rather than merging since there is nothing left to merge.

PR #49, opened 2026-07-11
feat(links): add file path links

ayamir:
## Summary

- add Cmd-click support for local file paths in terminal output
- resolve relative file paths against the pane cwd and expand `~/...` from the pane home when possible
- keep file-path hover tied to the platform modifier so plain mouse movement does not over-detect paths
- replay a pane's initial cwd immediately so relative paths work before the shell emits OSC 7

## Details

This keeps the existing URL behavior and extends the same link pipeline to file paths:

- OSC 8 hyperlinks still take priority.
- Plain URLs still resolve before file paths.
- Existing local files can be detected as absolute paths, relative paths, and `~/...` paths.
- `path:line` and `path:line:column` are parsed, while opening currently delegates to the system file opener.

For mouse-tracking TUIs, the platform modifier is now tracked from modifier-change events and reused by mouse hover/click handling. This avoids relying only on each mouse-move event's modifier snapshot.

## Testing

- `cargo fmt --check`
- `cargo test terminal::search`
- `cargo test daemon::pane::tests::attach_replays_initial_cwd_even_before_shell_reports_osc7`
- `cargo check`
- `cargo test`

## Effect

https://github.com/user-attachments/assets/64521dff-2da6-48e8-b884-331be95b0442

### DIFF
--- a/src/daemon/pane.rs
+++ b/src/daemon/pane.rs
@@ -423,7 +423,8 @@ impl DaemonPane {
         // client didn't pass an explicit cwd (tab-inherit / session restore still
         // win). Inherit → `forced` is `None`, so we keep the fallback as before.
         let forced = crate::core::config::working_directory_base();
-        if let Some(dir) = cwd.or(forced).or(fallback) {
+        let initial_cwd = cwd.or(forced).or(fallback);
+        if let Some(dir) = &initial_cwd {
             cmd.cwd(dir);
         }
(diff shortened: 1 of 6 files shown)


l0ng-ai:
Thanks a lot for this — clean work, and the tests + demo video are much appreciated! I'll review and follow up soon.

Issue #53, opened 2026-07-12
Auto-forward loopback URLs from SSH panes on Command-click


ayamir:
## Summary

It would be useful for tty7 to make remote development links work from SSH panes:

1. User SSHes into a remote machine inside a tty7 pane.
2. A remote process prints a loopback URL, for example `http://localhost:3000/dashboard`.
3. The user Command-clicks the URL.
4. tty7 starts a local SSH port forward in the background and opens the equivalent local URL in the browser.

For example, if the current pane is running `ssh user@devbox` and the remote output contains:

```text
http://localhost:3000/dashboard
```

tty7 could start a background forward similar to:

```bash
ssh -N -L 127.0.0.1:<local_port>:127.0.0.1:3000 user@devbox
```

Then open:

```text
http://127.0.0.1:<local_port>/dashboard
```

## Why this needs a real SSH context

The feature should not trigger just because terminal output contains `localhost`.

The same text means different things depending on pane state:

- In a local shell, `http://localhost:3000` should open the user's local machine.
- In an SSH session, `http://localhost:3000` usually means the remote machine's loopback interface.
- In nested SSH, tmux/screen, mosh, or wrapper scripts, tty7 may not be able to prove which machine produced the text.

So the safer contract is:

> Only auto-forward when tty7 has a trusted `remote_context` for the current pane and that context identifies a parseable SSH target. Otherwise, fall back to the existing link-opening behavior.

This should probably be more than a boolean `is_ssh`. A useful shape would be something like:

```rust
RemoteContext {
    kind: Ssh,
    target: "...",       // host / alias / user@host
    argv: [...],         // original or reconstructable ssh arguments
    pane_id: ...,
}
```

## Relevant current code paths

From a quick pass through the current implementation:

- Mouse handling already routes Command-click through `TerminalView::open_link_at` in `src/terminal/element.rs`.
- `TerminalView::open_link_at` in `src/terminal/view.rs` already handles OSC 8 hyperlinks, bare URLs, and local file paths.
- Link detection lives in `src/terminal/search.rs`.
- The daemon/client protocol in `src/daemon/protocol.rs` currently exposes pane state such as `Cwd`, `Prompt`, `Output`, `Title`, etc., but not foreground SSH argv or a stable remote-session identity.
- The daemon side can derive foreground process information for cwd/title, but the title-level information is not enough to reconstruct an SSH target safely.

[...]

PR #58, opened 2026-07-12
feat(links): add SSH loopback forwarding

ayamir:
## Summary

- detect foreground SSH sessions in the daemon and cache the active remote context per pane
- open Cmd-clicked loopback URLs through daemon-owned `ssh -N -L` local forwards
- add settings controls for enabling SSH loopback forwarding and viewing/closing active forwards

## Details

This keeps loopback forwarding behind an explicit `ssh_loopback_forward` setting. The daemon validates that the pane's foreground process is a plain SSH session before creating a forward, rejects unsafe SSH invocations such as remote commands or pre-existing forwarding modes, and reuses active forwards for the same pane/target/remote endpoint.

The terminal UI only broadens bare `localhost:port` link handling while a pane has an SSH remote context. URL opening is rewritten to the local forwarded port only after the daemon confirms the forward was created.

## Tests

- `rtk cargo test` -> `489 passed`

## Effect

https://github.com/user-attachments/assets/b66695c0-cd8f-4e4f-9baa-174a5032bcb5

image


l0ng-ai:
Thanks for this — the implementation is careful and well-tested.

But testing the branch surfaced a fundamental limitation. Since tty7 doesn't own the SSH connection, the forward has to be a **second, independent** `ssh -N -L`, which can only authenticate non-interactively (key/agent, or a pre-existing ControlMaster). On a password-only host it just fails:

```
$ ssh -N -o ExitOnForwardFailure=yes -o BatchMode=yes -L 127.0.0.1:…:localhost:3000 user@host
user@host: Permission denied (publickey,password).
```

`BatchMode=yes` (+ stdin → /dev/null) can't prompt for a password, so the Command-click silently does nothing. And it's not fixable in code: OpenSSH won't attach a forward to an already-running session that wasn't started as a ControlMaster — so "session already up + password auth + no master" can't work in any implementation.

This is the "best-effort by construction" concern from #53 made concrete: seamless for key/agent users, silently degraded for everyone else. The clean version only exists once tty7 owns the connection (`ssh -O forward` over a master we control — no re-auth, `RemoteContext` authoritative instead of inferred).

So I'd rather keep #53 open and blocked on that foundation than merge a heuristic V1. The strong parts here (the parser, the `RemoteContext` type, the settings UI) port straight into it once the foundation lands.

ayamir:
Added SSH profiles to command palette. The ssh connection opened from SSH profiles will auto inherit the behavior, e.g. cmd+click to auto forward.

image

ayamir:
Updated the implementation to make loopback forwarding depend on tty7-owned SSH sessions rather than trying to infer or duplicate arbitrary user-started `ssh` commands.

  Current shape:

  1. tty7 can now open managed SSH tabs.
     - Settings has a manual “Open managed SSH tab” entry.
     - Command palette has an “SSH Profiles…” entry.
     - SSH profiles are discovered from local OpenSSH `Host` aliases in `~/.ssh/config` and included files.
     - tty7 does not try to resolve `HostName`, `User`, `Port`, `ProxyJump`, `Match`, etc.; OpenSSH remains the source of truth when running `ssh <alias>`.

  2. Managed SSH tabs are represented explicitly in the daemon protocol.
     - `ShellSpec` now carries `ssh: Option<SshSpec>`.
     - `SshSpec` contains the target and safe SSH args.
     - When present, the daemon launches:
       `ssh -tt -o ControlMaster=yes -o ControlPersist=no -o ControlPath=<tty7-runtime-socket> ... <target>`
     - The daemon records a trusted `RemoteContext` with the ControlPath.

  3. Cmd-click loopback forwarding reuses the existing SSH master connection.
     - The UI only enables SSH loopback forwarding when the pane has `RemoteContext.control_path`.
     - When a remote `localhost` / `127.0.0.1` URL is opened, tty7 asks the daemon to ensure a local forward.
     - The daemon uses:
       `ssh -S <control_path> -O forward -L ... <target>`
     - So password-based SSH works after the user has logged into the managed SSH tab once; forwarding does not require a second authentication.

  4. This intentionally does not support arbitrary user-typed `ssh host` sessions.
     - Those sessions do not give tty7 an authoritative ControlPath.
     - Supporting them would require unreliable process/config inference or spawning a second SSH connection, which breaks password-auth cases.

  5. I also fixed the GUI/daemon version-skew case.
     - Managed SSH spawn now uses a distinct protocol kind instead of the older shell-spawn kind.
     - This prevents an older daemon from silently ignoring the new `ShellSpec.ssh` field and launching a plain SSH tab without ControlMaster.
     - If the daemon is too old, the client now gets a protocol failure and can restart/retry instead of silently degrading.

  Validation:
  - `cargo fmt --check`
  - `cargo check`
  - full `cargo test`
  - Added protocol coverage for managed SSH spawn not using the legacy shell-spawn message kind.

PR #64, opened 2026-07-13
fix(settings): make theme preview fill the picker card width

l0ng-ai:
The mini-terminal preview in the theme picker only filled the left ~55% of each card, leaving a large empty strip on the right.

**Two issues:**

1. **Width.** The preview uses `w_full` (percent width), which only resolves against a *definite* parent — the stretched flex ancestors in the picker column were not definite, so it collapsed to content width. Pin each card to the panel's content width (`300` − `px_4` gutters = `268px`), the same explicit-width pattern the search box already uses for the same reason.

2. **Proportions.** The "lines of code" bars were fixed pixel widths tuned for the narrow "Current theme" card, so on the wider picker card they clustered at the left edge. Switch them to width *fractions* (`relative()`) so the shape scales to any card width and stays ragged-right like real terminal text — the narrow current-theme card and the wide picker now share one adaptive shape.

No functional change; visual only.

### DIFF
--- a/src/ui/settings.rs
+++ b/src/ui/settings.rs
@@ -7,7 +7,7 @@
 
 use gpui::{
     AnyElement, Context, Div, Entity, FontWeight, Image, ImageFormat, KeyDownEvent, SharedString,
-    Stateful, Subscription, Window, div, img, prelude::*, px, rgb,
+    Stateful, Subscription, Window, div, img, prelude::*, px, relative, rgb,
 };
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Superseded: this exact change (fractional preview bars via relative() + the 268px card width) already landed on main as part of #65 (commit b9b1924, which this branch also carried). Closing as already included.

PR #81, opened 2026-07-14
feat(ssh): support Unix GSSAPI auth

ayamir:
## Summary

Add native SSH `gssapi-with-mic` authentication support for the connection manager.

This lets tty7 connect to Kerberos/GSSAPI-only SSH servers through the native russh session engine instead of requiring users to fall back to a shell-launched system `ssh`.

## Changes

- Add a `GSSAPI` auth mode to saved SSH profiles and the settings UI
- Try `gssapi-with-mic` first when auth mode is `Auto`
- Add a Unix `libgssapi` authenticator that drives russh's `GssapiAuthenticator` trait
- Request the mutual-authentication and integrity flags expected by OpenSSH-style GSSAPI server configurations
- For IP-literal hosts, try the original `host@IP` service first and then a PTR hostname fallback for environments where host principals are registered under canonical names
- Temporarily patch russh to the upstream GSSAPI branch until the protocol support lands in a released russh version
- Add an ignored live integration test for real SSH servers with local GSSAPI redentials

## Notes

The russh protocol support is proposed upstream in: https://github.com/Eugeny/russh/pull/738

This tty7 PR keeps platform Kerberos/GSSAPI integration on the tty7 side while russh owns the SSH protocol exchange.

## Verification

```text
cargo fmt --check
cargo test auth_mode_kebab_case_serialization
cargo test method_order_restricts_by_mode
cargo test gssapi_service_hosts
cargo check
```

`cargo check` passed with the existing future-incompatibility warning for transitive dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,9 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y pkg-config cmake clang libxkbcommon-dev \
             libxkbcommon-x11-dev libfontconfig1-dev libfreetype6-dev \
-            libwayland-dev libx11-dev libxcb1-dev libzstd-dev libssl-dev
+            libwayland-dev libx11-dev libxcb1-dev libzstd-dev libssl-dev \
+            libkrb5-dev
+          echo "LIBGSSAPI_IMPL=mit" >> "$GITHUB_ENV"
 
       - uses: dtolnay/rust-toolchain@stable
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,8 @@ jobs:
           sudo apt-get install -y pkg-config cmake clang libxkbcommon-dev \
             libxkbcommon-x11-dev libfontconfig1-dev libfreetype6-dev \
             libwayland-dev libx11-dev libxcb1-dev libzstd-dev libssl-dev \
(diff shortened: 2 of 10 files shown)


l0ng-ai:
Thanks for the contribution — the GSSAPI implementation looks solid and the test coverage is appreciated.

I pushed a small maintainer commit (fe095c7) to unblock merging:

- **Pinned the `[patch.crates-io]` russh override to the exact rev** (`0d1d073`) instead of tracking the branch. Since russh handles user credentials, a moving branch would let a future `cargo update` pull unreviewed code. Documented the removal condition (Eugeny/russh#737 releasing).
- **`gssapi_step`: a stalled context now errors** instead of returning `GssapiStep::Complete { mic: None }` — claiming completion without a MIC gets an opaque server-side rejection; a named error is debuggable.
- Updated the `auth.rs` module doc to include gssapi-with-mic in the Auto ordering.

Two non-blocking follow-ups worth considering later:
1. In Auto mode, an IP-literal host does a blocking PTR lookup before we know whether local Kerberos credentials even exist — a cheap default-credential probe first would avoid a DNS-timeout stall for non-Kerberos users whose servers advertise gssapi-with-mic.
2. The settings UI shows the GSSAPI option on Windows where it can only fail; a cfg-gate or an SSPI implementation would close that gap.

PR #102, opened 2026-07-16
feat(terminal): support shell vi mode

ayamir:
## Summary

- report shell edit mode from zsh/bash/fish integration via `OSC 133;V;0/1`
- disable tty7's local prompt editor while the shell reports vi-mode, letting zle/readline/fish own normal/insert display and control keys
- keep edit-mode metadata out of daemon `Prompt` state, restore it from replay snapshots, and preserve existing `zle_reading` semantics

## Verification

- `cargo fmt --check`
- `git diff --check`
- `cargo test -q shell_vi_mode -- --test-threads=1`
- `cargo test -q terminal::view -- --test-threads=1`
- `cargo test -q terminal::remote -- --test-threads=1`
- `cargo test -q daemon::pane:: -- --test-threads=1`

### DIFF
--- a/src/daemon/pane.rs
+++ b/src/daemon/pane.rs
@@ -1835,8 +1835,9 @@ impl OscSniffer {
             if let Some(path) = parse_osc7(payload) {
                 signals.cwd = Some(path);
             } else if let Some(rest) = payload.strip_prefix(b"133;") {
-                handle_osc133(shell, rest);
-                signals.shell = Some(shell.clone());
+                if handle_osc133(shell, rest) {
+                    signals.shell = Some(shell.clone());
+                }
             } else if let Some(event) = crate::core::cli_agent::parse_agent_event(payload) {
                 signals.agent_events.push(event);
             } else if let Some((title, body)) = crate::core::osc::parse_notification(payload) {
@@ -1852,7 +1853,7 @@ impl OscSniffer {
 }
 
 /// Fold one OSC 133 marker into the running shell state.
-fn handle_osc133(shell: &mut ShellState, rest: &[u8]) {
+fn handle_osc133(shell: &mut ShellState, rest: &[u8]) -> bool {
     shell.active = true;
     // `at_prompt` means "no foreground command is running" — i.e. the shell is
     // drawing or sitting at its prompt, so tty7's local line editor should own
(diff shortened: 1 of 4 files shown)


ayamir:
## Design note: shell vi-mode ownership

This PR intentionally changes ownership when the shell reports vi editing mode.

In that case, tty7 disables its local prompt editor for the current prompt and lets the shell line editor own input and display. That fixes the display/control-character mismatch because zle/readline/fish are then responsible for normal/insert mode behavior instead of tty7 trying to mirror or replay the prompt state.

The trade-off is that tty7's prompt-editor features are not active while shell vi-mode owns the prompt, including:

- local prompt overlay editing
- tty7 completion UI
- ghost suggestions
- prompt-local history navigation
- tty7 Ctrl-R history search UI
- tty7's local submit/history recording path for that line

Terminal-level features still work as before: rendering, scrollback/search, copy/paste, tabs/splits, cwd/prompt tracking, command status, links, SSH, notifications, and the shell's own vi-mode history/completion.

This is probably the main design point need to be discussed. There are two possible product directions:

- shell-owned vi mode: this PR's approach. It prioritizes correctness by letting the shell handle vi normal/insert semantics.
- tty7-owned vi keymap: keep tty7's prompt editor features, but implement vi normal/insert behavior inside tty7 instead of handing off to the shell. That is a larger feature, not a small fix.

l0ng-ai:
Thanks — the shell-owned direction is the right call here: vi users already bring their own completion/history tooling (zsh-autosuggestions, fzf, atuin), so handing the prompt back to the shell is what they want. Reviewed and pushed two fixes on top:

**1. `fix(terminal)`: gap hold vs vi prompt left a stale typeahead record**

Text typed during a command gap is held for the next prompt's editor, but a vi prompt never engages the editor, so the hold sat until the 150 ms timer dumped it *and recorded it as typeahead*. That record could never drain during vi prompts (`flush_typeahead` is gated on `input_active()`), so it lingered and flushed at the next emacs-mode prompt — a spurious `^U` plus the long-consumed gap text resurrected into the editor. Repro: vi mode → `sleep 1` → type `echo hi` during the sleep → run it → `bindkey -e` → phantom `echo hi` appears in the editor. The vi prompt now releases held gap input raw and drops any pending record without its wipe (test: `shell_vi_mode_prompt_releases_gap_hold_without_stale_typeahead`).

**2. `fix(daemon)`: detection missed zsh-vi-mode and inputrc-configured bash**

- zsh: `bindkey '^['` never says `vi-cmd-mode` under zsh-vi-mode (it rebinds Esc to `zvm_readkeys_handler`/`zvm_exit_insert_mode`), so the plugin's users — the biggest zsh vi cohort — kept the old conflict. Switched to the `main` keymap link (`bindkey -lL main` → `viins`), which both plain `bindkey -v` and zsh-vi-mode establish.
- bash: `[[ -o vi ]]` misses vi mode set only via `~/.inputrc` (`set editing-mode vi`). Switched to `bind -v`, which reports readline's actual mode either way.

Both verified against real zsh/bash for all six mode/config combinations. Full suite green (726 tests). Merging.

PR #130, opened 2026-07-19
feat(shell-integration): support Git Bash on Windows

l0ng-ai:
PowerShell was the only Windows shell getting OSC 133 integration. Git Bash was excluded by two unrelated gates, neither of which looks deliberate.

> **Stacked on #129.** Based on `fix/windows-git-probe-console-flash` so the diff shows only this change; GitHub will retarget to `main` once #129 merges.

## The two gates

**`shell_kind` stripped `.exe` only for PowerShell**, so the `bash.exe` that `core::shells::find_git_bash` resolves matched nothing — even though the comment directly above it reasons about exactly that suffix. Now stripped for every shell, which also collapses the PowerShell arm's four spellings into two.

**The Git Bash dropdown row ships `-i -l`**, which tripped `has_custom_args` and made `setup` decline bash outright. But that guard exists to protect args the *user* configured — these are tty7's own, from `detect_shells`. `ShellSpec` now records who authored its args, so integration may respell tty7's (`--rcfile … -i` plus the replayed login-file chain means the same thing) while leaving the user's alone. The `-i -l` remain as the fallback for when integration doesn't apply or fails to set up, which matters: `setup_bash` is best-effort, and a bare non-login `bash.exe` would skip `/etc/profile`, where Git Bash's PATH and HOME come from.

`#[serde(default)]` is `false`, so an older client's frame — which can only carry user-configured args — keeps them untouched.

## One msys2 detail

The rcfile path is now spelled with forward slashes. The msys runtime accepts `C:\...` too, but the backslashes survive into any bash string context the path later reaches, where they *are* escape characters.

## Verification

Not just by construction — there's a new live-PTY test that spawns the real Git Bash through the real `setup` output and asserts the full `A`/`B`/`C`/`D;1` cycle plus OSC 7 come back. It skips when Git for Windows isn't installed. Green in 0.6s locally.

Getting it green surfaced two ConPTY-isms, both noted at the call sites: the master doesn't reliably EOF when its child exits, and closing its input side raises a console control event that kills the shell with `STATUS_CONTROL_C_EXIT`.

691 tests pass; clippy warning count unchanged from baseline.

**Still worth a human pass:** the test asserts the marks are emitted, not that dotfiles survive intact. Since this switches Git Bash from a login shell to non-login + a replayed startup chain, the thing to eyeball is that the MINGW64 prompt, aliases, and PATH are unchanged.

## cmd and WSL

Both stay unintegrated, now documented as decisions rather than gaps. **cmd**'s only hook is the `PROMPT` env var, which can emit `A`/`B` but not `C` or `D` — it has no preexec/postexec, and `PROMPT` is expanded when set, so even `%ERRORLEVEL%` is out of reach. Since only `C` clears `at_prompt` (`pane::handle_osc133`), an A/B-only shell would leave the line editor holding the keyboard for the whole of every command — worse than no integration. **WSL** is spawned as `wsl.exe`, the Windows-side launcher; reaching the distro's own shell needs per-distro shell detection plus `WSLENV` path translation, which is its own piece of work.

### DIFF
--- a/src/core/shells.rs
+++ b/src/core/shells.rs
@@ -278,7 +278,11 @@ fn detect_windows() -> Vec<DetectedShell> {
         out.push(DetectedShell {
             label: "Git Bash".into(),
             program: bash.to_string_lossy().into_owned(),
-            // Interactive login shell — matches Git Bash's own launcher.
+            // Interactive login shell — matches Git Bash's own launcher. These
+            // are tty7's args, not the user's, so shell integration may replace
+            // them with its own spelling of the same thing (see
+            // `protocol::ShellSpec::args_are_tty7_defaults`); they stand as the
+            // fallback for when integration doesn't apply or fails to set up.
             args: vec!["-i".into(), "-l".into()],
         });
     }
@@ -294,6 +298,14 @@ fn detect_windows() -> Vec<DetectedShell> {
     }
 
     out
+}
+
+/// Git Bash's `bash.exe`, if Git for Windows is installed. Exposed only to
+/// tests, so `daemon::shell_integration`'s live-PTY check can spawn the same
+/// binary the dropdown does (and skip itself when there is none).
+#[cfg(all(windows, test))]
+pub fn git_bash_path() -> Option<PathBuf> {
+    find_git_bash()
 }
 
 /// Git Bash from the usual Git-for-Windows install roots (machine-wide x64,
--- a/src/daemon/pane.rs
+++ b/src/daemon/pane.rs
@@ -2051,7 +2051,9 @@ fn strip_uri_drive_slash(path: &str) -> &str {
 }
 
 /// Parse an OSC 7 `file://HOST/PATH` (or bare absolute path) payload.
-fn parse_osc7(payload: &[u8]) -> Option<PathBuf> {
+/// `pub(crate)` so the shell-integration tests can round-trip what the snippets
+/// actually emit through the parser that consumes it.
+pub(crate) fn parse_osc7(payload: &[u8]) -> Option<PathBuf> {
     let rest = payload.strip_prefix(b"7;")?;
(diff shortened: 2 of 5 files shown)


l0ng-ai:
Pushed `caec744` — a regression this PR introduced, found while scoping WSL support.

**Git Bash's `$PWD` is an msys path**, and this PR started reporting it over OSC 7. Measured on a real machine:

```
$ echo $PWD
/c/Users/thoma/repo/tty7-clean
```

The daemon runs Windows-side, where that is not absolute but **drive-relative** — `strip_uri_drive_slash` only un-prefixes the `/C:/…` shape, so it resolved to `C:\c\Users\thoma\repo\tty7-clean`.

Impact, all silent:

- git-status probe returned `None` on the `!cwd.exists()` guard — the sidebar's git line never appeared in a Git Bash pane
- path completion and link resolution stopped matching
- **split / new tab broke outright**: an inherited cwd wins over every fallback in `initial_working_directory`, so the next shell was spawned with a working directory that does not exist

Before this PR, Git Bash had no integration, so it reported no cwd at all and the fallbacks applied. `pane.rs` even states the intent: *"a shell without integration reports no cwd **rather than a wrong one**"*. This PR gave it one, and it was wrong.

**Fix:** `pwd -W`, msys's own translation to the real Windows path. It also maps msys-only mounts correctly:

```
$ cd /tmp && pwd -W
C:/Users/thoma/AppData/Local/Temp
```

Falls back to `$PWD` if it fails or returns empty; the branch is taken once at install time, not per prompt.

The live-PTY test now round-trips what Git Bash actually emits through the daemon's own `parse_osc7` and asserts the resulting path exists — so the emitter and the parser are verified to agree rather than each checked in isolation. That test fails without the fix.

692 tests pass, clippy warning count unchanged.

Note the drive-relative hazard is not specific to Git Bash — a remote pane's cwd has the same problem, and nothing gates it. That is #133, which is independent of this PR and based on `main`.

PR #134, opened 2026-07-19
feat(shell-integration): support WSL

l0ng-ai:
Closes the last Windows shell gap. `wsl.exe` panes now get OSC 133 / OSC 7, so the inline input editor works in them.

## Reaching through the launcher

`wsl.exe` is not a shell — it starts one *inside* a distro, so injecting into it directly would never reach the thing that draws the prompt.

We write the integration rcfile on the Windows side and pass its path in via `WSLENV`, whose `/p` flag rewrites it to the distro's view of the filesystem. That is why the `/mnt` automount root is never hardcoded — it is configurable in `/etc/wsl.conf`.

The argv becomes:

```
wsl.exe --distribution <d> --cd ~ -- sh -c '<script>'
```

rather than `-- /bin/bash --rcfile <path>`, because the path only exists as an env var *inside* the distro after translation, and `wsl.exe` execs its command directly with no shell to expand it.

The script picks the shell and execs it, in-distro:

```sh
case "${SHELL:-}" in
  */bash) exec "$SHELL" --rcfile "$TTY7_RC" -i ;;
  *)      exec "${SHELL:-/bin/sh}" -l ;;
esac
```

**The decision is deliberately not made on the Windows side.** Spawning a second `wsl.exe` to ask which shell a distro uses blocks the whole spawn path: the client waits synchronously for the daemon's `Spawn` reply, so a cold WSL start freezes the entire window for seconds. Folding it into the one invocation we were always going to make costs nothing and cannot block. (The first revision of this PR did probe, and did freeze — see the second commit.)

`$SHELL` rather than `getent passwd`: WSL populates it from the user's passwd entry, so inside the distro it already *is* the login shell of record — the same source `shell_kind` trusts on Unix.

**No new shell code.** A bash distro reuses the existing snippet verbatim. zsh and fish distros fall through to the second arm and launch as a plain login shell — what every WSL pane did before this. They are integrable the same way, but each needs its own verification pass.

## The other half: tagging the pane

Not optional. A distro reports `/home/me/proj`, and on Windows that is not an *invalid* path but a **drive-relative** one:

```
PS> [System.IO.Path]::GetFullPath("/home/thomas/proj")
C:\home\thomas\proj
```

A new `RemoteKind::Wsl` marks these panes so `TerminalView::local_cwd` (from #133) declines their cwd. Without it the git probe, path completion, link resolution and cwd inheritance would all consume that path — and on a machine that happens to have such a directory, silently consume the wrong one.

Nothing matches exhaustively on `RemoteKind`, so adding a variant compiles clean everywhere and every miss is a silent fall-through. Two needed handling:

- **the foreground-`ssh` poll** suppressed itself only for `NativeSsh`, so it would have blanked the WSL tag — and the pane's cwd with it — **twice a second**. It now only replaces the kind it authors.
- **the tab status dot and `active_ssh_pane`** read "has a `RemoteContext`" as "is an SSH pane". Both now test the kind.

## Rename

`Injection::force_non_login` -> `replaces_argv`. bash needed it because `--rcfile` is ignored for login shells; WSL needs it because the launch flags and command must be reordered around `--`. The mechanism was always "these args replace rather than extend" — only the name was bash's.

## Verification

A live test drives a real ConPTY into a real distro and asserts the full A/B/C/D cycle comes back — the only way to show `WSLENV` translation, `wsl.exe`'s argv passing (the script contains quotes Windows command-line quoting could mangle), the in-distro `case`, and the distro's own startup chain all survive together. It shares its harness with the Git Bash test.

A second test names a distro that **cannot exist** and asserts setup still succeeds — if anything asked the distro a question, it could not. That is the deterministic form of the no-blocking guarantee; a timing bound would only fire on a cold machine, which is exactly the blind spot that let the freeze ship.

736 tests pass, clippy warning count unchanged (124).

## Still worth a human pass

- **Only tested against one distro** (Ubuntu-24.04, bash 5.2). Distro-agnostic by construction, not by measurement — in particular the non-bash `exec "$SHELL" -l` arm has never been run.
- **The `$TTY7_RC` quoting** is verified for a temp path without spaces. The quotes are there because the Windows temp dir contains the account name and so may contain spaces, but that case is reasoned, not run.
- **Splits of a WSL pane no longer inherit the directory** — they land in the distro's home. Inheriting would need `wslpath` translation both ways; not attempted here.

### DIFF
--- a/src/core/shells.rs
+++ b/src/core/shells.rs
@@ -308,6 +308,14 @@ pub fn git_bash_path() -> Option<PathBuf> {
     find_git_bash()
 }
 
+/// Installed WSL distributions. Exposed only to tests, for the same reason as
+/// [`git_bash_path`]: the live-PTY check needs a real distro to launch into,
+/// and skips itself when there is none.
+#[cfg(all(windows, test))]
+pub fn wsl_distros() -> Vec<String> {
+    list_wsl_distros()
+}
+
 /// Git Bash from the usual Git-for-Windows install roots (machine-wide x64,
 /// x86, and the per-user installer's home).
 #[cfg(windows)]
--- a/src/daemon/pane.rs
+++ b/src/daemon/pane.rs
@@ -184,9 +184,14 @@ fn build_spawn_config(
     shell: Option<ShellSpec>,
 ) -> anyhow::Result<SpawnConfig> {
     let initial_cwd = initial_working_directory(cwd);
-    // Computed before `build_shell_command` consumes the spec.
-    let remote = wsl_remote_context(shell.as_ref());
-    let (cmd, integration_dir) = build_shell_command(shell, &initial_cwd)?;
+    // Resolved here rather than inside `build_shell_command` because the WSL tag
+    // must be read off the shell we *actually* launch. `shell` is only the
+    // per-spawn override; `config.json` supplies the program when it is `None`,
+    // and a `wsl.exe` configured there is just as much a WSL pane as one picked
+    // from the dropdown.
+    let configured = choose_shell(shell, crate::core::config::shell_command());
+    let remote = wsl_remote_context(configured.as_ref());
+    let (cmd, integration_dir) = build_shell_command(configured, &initial_cwd)?;
     Ok(SpawnConfig {
         cmd,
         initial_cwd,
@@ -196,7 +201,7 @@ fn build_spawn_config(
 }
 
 /// Tag a `wsl.exe` pane as living in another filesystem namespace, from the
-/// spawn spec rather than the process table — `wsl.exe` is exactly what tty7
+/// resolved shell rather than the process table — `wsl.exe` is exactly what tty7
 /// launched, so there is nothing to detect.
 ///
 /// This is what makes `TerminalView::local_cwd` decline the distro's cwd, and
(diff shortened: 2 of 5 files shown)


l0ng-ai:
Pushed `f91707e` — a freeze in the first revision of this PR, found in real use rather than by the tests.

`setup_wsl` resolved the distro's login shell with a synchronous `wsl.exe` call. The client waits for the daemon's `Spawn` reply (`terminal::remote::spawn`), so on a **cold** WSL start — seconds, while the distro boots — the whole window locked up.

Caching per distro would not have fixed it: the first WSL pane after launch is exactly when the distro is cold, so the freeze hit precisely the case a cache cannot cover.

The fix removes the probe entirely rather than moving it off-thread — the decision now happens inside the single `wsl.exe` invocation we were always going to make, so there is no second call to block on.

**Why the tests missed it**, since that is the more useful part: the `--cd`/`-d` unit tests never reach the probe, and the live-PTY test only ever ran against an already-warm distro. The new regression test names a distro that cannot exist and asserts setup still succeeds — if anything contacted the distro, it could not. A timing bound would only have fired on a cold machine, which is the same blind spot.

Removes `wsl_login_shell` and `inner_shell_kind`, both now unreachable.

l0ng-ai:
Review pass — one real gap, fixed in c780799.

**`wsl_remote_context` was reading the wrong shell.** It took the per-spawn `ShellSpec` override, but the program actually launched comes out of `choose_shell`, which falls back to `config.json`'s `shell` when there's no override. So:

```json
{"shell": {"program": "wsl.exe", "args": []}}
```

...reaches `setup_wsl` (empty args → nothing custom to preserve), the distro starts reporting `/home/me/proj` over OSC 7 — and `wsl_remote_context(None)` leaves the pane untagged. `local_cwd` accepts it and the local git probe runs in `C:\home\me\proj`.

That's the drive-relative collision the "other half" section argues is not optional, and this PR is what makes it reachable: before, such a pane reported no cwd at all.

Fix moves `choose_shell` up into `build_spawn_config` so its result feeds both the tag and `build_shell_command` — they can no longer describe different shells. `wsl_remote_context` now takes `ChosenShell` rather than `ShellSpec`, which is what keeps it that way, since the resolved type only exists after the fallback.

Also folded the duplicate distro parser into `shell_integration::wsl_distro` — the tag's copy missed `--distribution=NAME`, so a user configuring that spelling got a correctly-integrated pane with an empty `target`. Display-only, but two parsers for one flag are free to drift.

736 pass, clippy still 124, including the live-PTY test against a real distro.

---

Rest of the design holds up. I checked the four `RemoteContext` consumers the description doesn't mention — `server.rs:378`, `forwards.rs:37`, `view.rs:2630`, `view.rs:3386` — and all four are either already kind-gated or treat any `RemoteContext` as non-local, which is the safe direction here. The `!= Ssh` inversion on the foreground poll is right, and `wsl_setup_never_contacts_the_distro` is a genuinely good way to pin the no-blocking guarantee.

Of the self-declared gaps, the one I'd weight highest is the non-bash `exec "$SHELL" -l` arm never having run — it's the fallback keeping zsh/fish distros working at all, so a quoting error there breaks panes that work today.

PR #143, opened 2026-07-20
feat(links): open file links with a configurable command

ayamir:
## What

Adds an optional `link_file_command` config that runs a user-defined command
when ⌘/Ctrl-clicking a detected file-path link, instead of tty7's built-in
"open in the default app" behavior. The default (`None`) keeps the current
open-in-default-app behavior unchanged.

The command template is tokenized on whitespace and the placeholders `{path}`,
`{line}`, and `{column}` are substituted per argument. An argument that
references a placeholder with no value (e.g. `--line={line}` on a link with no
line number) is dropped whole, so flags disappear cleanly rather than dangling.

A "Open files with" text field is added to the Links section of Settings;
clearing it restores the built-in open.

## Why

File-path link detection (#49) always opens via the OS default app. Users who
live in a terminal editor want a clicked `path:line` to open in their editor at
that line — for example, driving a terminal multiplexer to open the file in an
existing editor pane:

```
herdr edit {path} --line={line} --column={column}
```

## How

- `core/config.rs`: new `link_file_command: Option<String>`; empty/whitespace
  normalizes to `None`.
- `terminal/view.rs`: on a `File` link click, a configured template takes
  precedence over `open_file_path`; `expand_file_command_template` /
  `expand_file_command_token` do per-token placeholder substitution and drop
  tokens whose placeholder has no value. The command is spawned detached — tty7
  does not wait on or read from the launched editor.
- `ui/app.rs`, `ui/settings.rs`: "Open files with" settings input, committing on
  Enter/blur; empty clears the override.

## Testing

- `cargo fmt --check` — clean
- `git diff --check upstream/main...HEAD` — clean
- `cargo test file_command` — 3 passed (template substitution, absent-value
  token drop, path-only + unknown-placeholder passthrough)

### DIFF
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -87,6 +87,14 @@ pub struct Config {
     /// Detect URLs (OSC 8 hyperlinks + bare URLs in the text), underline them on
     /// hover, and open them on ⌘/Ctrl-click. On by default.
     pub link_url: bool,
+    /// Optional command template run when ⌘/Ctrl-clicking a detected file-path
+    /// link, instead of tty7's built-in "open in the default app" behavior. The
+    /// template is tokenized on whitespace and the placeholders `{path}`,
+    /// `{line}`, and `{column}` are substituted per argument; an argument that
+    /// contains a placeholder with no value (e.g. `{line}` on a link that has no
+    /// line number) is dropped. `None` (the default) keeps the built-in open.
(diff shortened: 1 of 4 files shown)


ayamir:
https://github.com/ogulcancelik/herdr/discussions/1639

PR #235, opened 2026-07-28
feat(remote): remote workspaces

l0ng-ai:
Pick a dev box and tty7 gives you a window that *is* that machine — every tab, pane, repo group, branch line, diff and worktree in it lives there, and it keeps running after you close the lid.

## Before / After

| | Before | After |
|---|---|---|
| Working on a remote box | SSH pane from the command palette | Home page → **Connect to Host** → a window bound to a workspace on that machine |
| Closing the lid / losing the network | The agent in that SSH pane dies with the connection | Sessions keep running on the far side; reconnecting lands back in place |
| Repo grouping, branch line, diff overlay, right-panel Changes, worktrees | Gone the moment you SSH — all of it reads the client's disk | All present, answered by the machine the pane is actually on |
| File tree / editor over SSH | SFTP browsing only | Full `Host`-backed tree, editor, search and watch |
| Workspace picking | Three partial surfaces — title-bar chip dropdown (no machine on the rows), home-page picker (closed workspaces only, empty windows only), the connect wizard | One switcher overlay, grouped by machine, this computer included |
| Where the daemon lives | `src/daemon` inside the GUI crate | `crates/tty7-core`, with a headless `crates/tty7-server` binary built on it |

Two features, not one: an SSH pane is still "connect for a minute and look at a log". A remote workspace is "develop there". A window is never a mix of both — that is enforced, not documented.

## Structure

```mermaid
flowchart LR
  subgraph client["Client machine (GUI)"]
    views["ui:: / terminal:: views"] --> ops["ui::host_ops<br/>(background executor)"]
    ops --> reg["HostRegistry<br/>HostId → SharedHost"]
    reg --> local["LocalHost"]
    reg --> remote["RemoteHost"]
    local --> disk[("local fs / git")]
  end
  remote -- "control conn (kind 60-63)" --> srv
  subgraph far["Remote machine"]
    srv["tty7-server<br/>(static musl)"] --> rlocal["LocalHost"]
    rlocal --> rdisk[("remote fs / git")]
    srv --> panes["PTYs, sessions,<br/>workspace store"]
  end
```

Both sides run the *same* `LocalHost` — the server is that code with no GUI attached, which is what keeps local and remote from drifting apart.

## Why this shape

**`Host` blocks, callers go async.** An `async` trait would need `Arc<dyn Host>` boxing on the 99% path (`LocalHost`), and the server's RPC handlers are blocking thread-pool work anyway. So the trait stays synchronous and `ui::host_ops` is the single door that moves every call onto the background executor — de-duplication, staleness checks and error reporting live there once instead of being re-derived per call site. A CI grep keeps `ui::`/`terminal::` from reaching `std::fs`, `git`, `canonicalize` or `is_absolute` behind its back (allowlisted per (file, pattern), with a reason on each entry).

**Three crates, one dependency edge.** `tty7-server` depends on nothing but `tty7-core`; if it ever needs a second crate the GUI also needs, that crate belongs in core. `keyring` and `gssapi` are deliberately *not* in core's defaults — a static musl binary pushed onto arbitrary machines should not be linking zbus/secret-service or a system krb5 for code it can never reach.

## CI

| Job | What | Required? |
|---|---|---|
| `tty7-server musl (x86_64 / aarch64)` | Static musl builds via cargo-zigbuild, asserted loader-free by `assert-static.sh` | **No** — deliberately |
| `host boundary (§10.6)` | The fs/git grep above | **No** — deliberately |

Both stay non-required on purpose: main's required checks are `rustfmt` plus the three `build & test (<target>)` names, and adding to that set — or reshaping that matrix — would wedge every open PR on a check it has never run. `docs/remote-server-assets.md` has the promotion plan.

## Testing

- `cargo test --workspace`: **1415 passed, 0 failed** (2 ignored are the GSSAPI cases, whose feature is off unless the GUI turns it on).
- `cargo fmt --all --check` clean.
- 50-case `Host` conformance suite runs against a **real `tty7-server --stdio` child process** (`crates/tty7-server/tests/stdio_conformance.rs`) — same suite `LocalHost` passes in-process, so a divergence between local and remote fails the build.
- 5 workspace-store cases over the control wire, including takeover: a second client claiming a workspace cuts the first one off.
- Static-link assertion verified locally on both musl targets; binaries run under Alpine.

## Docs

`docs/2026-07-27-remote-workspace-design.md` is the direction; `impl-contract.md` is the interface — signatures, wire byte layouts, kind numbers, module ownership. Where the two disagree the contract wins, and its §11 records each divergence with the reason. `docs/remote-server-assets.md` covers how the server binary is built, shipped and promoted.


shihuaidexianyu:
I think we need to come up with a user-friendly interaction method. WezTerm has a similar feature, but I find it a bit cumbersome. Users would probably prefer us to hide some of the internal implementation details from them.

l0ng-ai:
> I think we need to come up with a user-friendly interaction method. WezTerm has a similar feature, but I find it a bit cumbersome. Users would probably prefer us to hide some of the internal implementation details from them.

Yes, I'm also considering how to design an intuitive UX.

PR #248, opened 2026-07-28
feat(cli): session CLI for scripting and agent orchestration

yetone:
## What

The `tty7` binary now doubles as a session CLI:

```
tty7 ls --json            # every pane + its agent's status (idle/working/waiting/done)
tty7 spawn --cwd ~/app    # create a detached pane, prints its id
tty7 send 3 'claude -p "fix the failing tests"'
tty7 wait 3 --until waiting,done --timeout 600
tty7 capture 3 --lines 100
tty7 kill 3
```

## Why

The daemon already tracks, per pane, which coding agent runs there and its rich status — idle / working / **waiting on the user** / done, plus the agent's native session id — but the only consumer was the GUI's status dots. This PR hands the same view to programs: one agent can spawn a pane for another, send it a prompt, sleep until its peer blocks on a permission prompt or finishes a turn, and read its output. That `wait` primitive is exactly what tmux-based agent teams have to fake with `capture-pane` screen-scraping; here it's one structured poll, and the tty7 GUI doubles as the team's monitoring surface for free.

## Wire changes (backward-compatible)

- `PaneInfo` carries `agent` + `agent_session` (serde-defaulted, omitted when `None`) so `List` answers `ls`/`wait` in one round trip
- New control messages `SendInput` (kind 51, binary-framed like `Input`) and `CaptureScrollback` (kind 52) with `InputSent`/`Scrollback` replies — control-side by design: panes are single-subscriber, and an `Attach` would steal the GUI's live stream
- An old daemon rejects the unknown kinds with a hangup; the CLI diagnoses that as "daemon too old, restart it". No `PROTOCOL_VERSION` bump — the GUI is unaffected
- `Pane::scrollback_tail` slices the replay ring byte-wise

## Design notes

- `capture` reduces raw PTY bytes to plain text with a page-long extractor (SGR/OSC/DCS stripped; `\r` overwrites in place so a progress bar reads as its final frame). Deliberately not an emulator — cursor addressing is ignored, `--raw` gives the bytes
- Exit codes are script-friendly: `0` ok · `1` error · `2` usage · `3` daemon unreachable · `124` wait timeout
- `spawn` auto-starts the daemon (entry point of any orchestration script); the other subcommands error when none is running
- `--config-dir` works in every subcommand (it selects the socket); a bare `tty7` still launches the app

## Testing

- Protocol round-trips extended (new kinds, agent-bearing `PaneInfo`, binary `SendInput` framing incl. empty/NUL payloads); 5 new text-extractor tests (SGR/OSC stripping, CR overwrite, tabs/backspace, CJK, blank trimming). Full suite: **981 passed, 0 failed**
- E2E against a scratch daemon (`--config-dir` isolated): spawn → send → capture round-trip runs real commands; `ls --json`; kill; all five exit codes
- Agent-status path verified with real sentinel OSC events: `permission-request` flips `ls` to `waiting` and wakes `wait --until waiting`; `stop` wakes `wait --until done` within one poll interval, reply carrying the event's message + session id


yetone:
Pushed a second commit (`60b1b75`) covering the discovery half — how agents *learn* the CLI exists (the termio-style flow):

- **`core::agent_note`**: a marked, idempotent block describing the CLI, installed into `~/.claude/CLAUDE.md` (always, `CLAUDE_CONFIG_DIR` honored) and `~/.codex/AGENTS.md` (only when `~/.codex` exists). Uninstall strips exactly the marked block; an unterminated block is never truncated at a guess.
- **One-time prompt** — "Let your agents coordinate?" (Not Now / Enable) the first time a pane detects a coding agent; never re-asks either way.
- **Settings → Agents** gains an "Agent coordination" switch above the hook rows; state derives from the files themselves.

3 new unit tests (upsert idempotency + user-text preservation, byte-identical removal round-trips, unterminated-block safety). Full suite: 984 passed.

l0ng-ai:
Thanks — this is exactly the direction we want (the daemon already knows everything the GUI shows; handing that view to programs is the right call), and the `wait --until` primitive on agent status is a genuinely useful building block. CI is green and the wire changes look safely additive.

Heads-up on timing: I'm going to hold this until after the upcoming **remote workspace** release. The plan after that release is to sink the workspace/tab/pane structure into the daemon (the same store `tty7-server` already serves for remote machines, applied to localhost), and I'd like the CLI to land against that shape in one pass instead of shipping a pane-only contract first and breaking it later. Two things I'll want to adjust when we pick it up:

1. **`ls --json` should print an object, not a bare array** — e.g. `{"panes": [...]}` — so a `workspaces` dimension can be added later without breaking scripts. This is the one contract that's hard to change once released.
2. **`spawn` will grow workspace awareness** (a `--workspace` flag or a default record) so CLI-spawned panes become visible in the GUI rather than living as orphans only the CLI can see.

Nit: the PR body says kinds 51/52, the code uses 53/54.

Nothing for you to do right now — I'll ping here once the remote workspace release is out and we can rebase and finalize. Sorry for the wait, and thanks again for the thorough test coverage.

yetone:
Rebuilt per your comment — thanks for landing the daemon-owned tree and the full CLI first, it made this much smaller.

The branch is now **one commit on top of main** (`8ec4f1c`): everything superseded by `tty7-cli` (my ls/spawn/send/capture, the protocol messages, the PATH injection) is dropped, keeping only `tty7 wait` (built on `ControlRequest::AgentStates` + the machine tree, in the clap/backend/mock idiom) and the agent-coordination note with its one-time prompt and Settings switch. Your two contract notes are moot in the new shape — `ls --json` is yours now, and panes here are created through `tab new`, so nothing orphaned; the 51/52 body nit died with the old protocol changes. PR body rewritten to match. Full workspace: 1852 passed.

l0ng-ai:
I wanted to discuss whether we need the global instruction injection here.

I really like `tty7 wait/send/capture` as low-level orchestration primitives. But it may be a better fit for the primary agent to discover and use them explicitly through a tty7 orchestration skill, rather than writing guidance into `~/.claude/CLAUDE.md` / `~/.codex/AGENTS.md` whenever tty7 detects an agent.

Global instructions affect every project and future session, and they effectively encourage every agent to discover and orchestrate other agents. For the more common primary-agent → worker-agent model, the primary agent can own decomposition, dispatch, waiting, and aggregation; workers can simply receive bounded tasks without needing persistent awareness of, or authority to orchestrate, other panes.

So I’d lean toward keeping `tty7 wait` and the other primitives, while moving the workflow guidance into a skill that the user or primary agent invokes explicitly. Are there scenarios you have in mind that a skill would not cover here?

yetone:
Agreed — and done in `20e92ce`. You're right that the global note hands *every* agent, workers included, ambient orchestration authority, and taxes every session's context to boot. I went looking for scenarios a skill wouldn't cover and found only two, neither compelling:

1. **Agents without a skill system** (Codex et al.): the old note covered them via `~/.codex/AGENTS.md` — but that's precisely the global-instruction shape you're objecting to. A user who wants a non-Claude primary agent to orchestrate can point it at the skill file or paste the loop into a *project's* AGENTS.md, which keeps the blast radius chosen rather than ambient.
2. **Spontaneous worker-side delegation** (a worker discovering it can fan out): that's not a gap — it's the failure mode. The skill's own guidance now says it explicitly: keep workers bounded, don't hand them orchestration duties.

So: `core::agent_note` is gone (with its first-agent-detected prompt and config flag). `core::orchestration_skill` installs `~/.claude/skills/tty7-orchestration/SKILL.md` — one-line description in context until explicitly invoked, full delegation loop (tab new → send → wait → answer-or-capture → pane close) in the body, wholly tty7-owned (overwrite-to-install doubles as version refresh; uninstall keys on an ownership marker so a hand-written skill under the same name is refused, not deleted). The Settings → Agents switch drives it. `tty7 wait` is untouched.

Full workspace: 1851 passed, 0 failed.

PR #250, opened 2026-07-28
fix(terminal): prevent fullwidth punctuation overlap and speed OSC mark scanning

GreenOolong:
Fixes CJK/fullwidth punctuation overlap by shaping each wide glyph independently. Reuses interned single-character strings after this split and changes MarkScanner text scanning to SIMD-backed memchr. Reconciled with upstream Thai/Lao SARA AM clustering.

---

Maintainer edits (2026-07-29):

1. Rebased onto current main and re-declared `memchr` in the root crate's Cargo.toml — the dependency left with the OSC tokenizer's move down to `tty7-core`, so the branch compiled on its old base but its merge with main did not (the first three red `build & test` runs).
2. **Fixed an infinite loop the unbatching introduced**: removing the batching loop also removed its `col += 2`, so `segment_row`'s wide-glyph arm re-processed the same column forever and grew `segs` until allocation failed — the 6 GiB abort on the Windows runner and the hung macOS job in the second CI round. The original "980 tests passed" validation could not have run these tests; any row with a wide glyph loops.
3. Commit messages reworded to English per repo convention.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9553,6 +9553,7 @@ dependencies = [
  "ksni",
  "libc",
  "log",
+ "memchr",
  "notify 8.2.0",
  "notify-rust",
  "objc2 0.6.4",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,11 @@ log.workspace = true
 # Smart double-click selection patterns (URL/email/path). Already in the tree
 # transitively, so pinning it here adds no new native code.
 regex = "1"
+# SIMD byte search for `MarkScanner`'s Text-state fast path (`terminal::marks`),
+# which scans every output batch the client receives — the same skip-ahead the
+# tokenizers in `tty7-core` already use. Already in the tree via `tty7-core`,
+# so this pins no new code.
+memchr = "2"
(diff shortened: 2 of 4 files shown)


l0ng-ai:
Reproduction verified on macOS (isolated dev instance, seeded scrollback):

- **Default setup (Hack + PingFang SC fallback): the bug does NOT manifest.** Pixel-level comparison of main vs this branch showed identical glyph positions — PingFang's fullwidth punctuation advances a full 1 em, which clears `apply_force_width_to_layout`'s half-slot test.
- **Narrow-punctuation CJK fonts DO trigger it.** With a test face whose `（）「」【】` advance is 0.6 em (the Sarasa-Term-style half-width-punctuation shape), main paints the glyph after each such character on top of it and compresses the rest of the line; this branch renders every glyph in its own two-column slot. `，。：？` (still 1.2 em in the test face) stayed correct on both builds, matching the misclassification model exactly.

So the fix is real but its blast radius is font-dependent: default macOS fallbacks are unaffected; half-width-punctuation fonts (a popular choice among CJK terminal users) hit it on every affected punctuation mark.

Issue #269, opened 2026-07-30
Bug Report: SSH/Linux terminal issues, key bindings, paste behavior, and remote Git status in tty7


majunbo:
### Type

Bug — something doesn't work as expected

### What's going on?

Hello, this is my first time submitting a bug report on GitHub.

First of all, tty7 does have many advantages, such as fast startup speed, beautiful themes, and low memory usage. However, I have also encountered some issues:

* When remotely connecting to a Linux (Ubuntu) system via SSH, pressing `Ctrl+R` to search command history sometimes does not return any results, or the displayed results may come from the local Windows command history instead.

* When connected to a remote Linux system, it would be helpful to add a copy function when selecting text in the current command line or terminal window. In other words, selecting text should automatically copy it to the clipboard. For pasting, I suggest supporting the standard Linux shortcut `Shift+Insert`.

* When typing `cd` in a Linux terminal, the first suggested directory name requires pressing the right arrow key to accept it. Normally, pressing `Tab` for auto-completion is the most common workflow. Currently, when pressing `Tab` in tty7, it shows other folders in the current directory, which is not the expected behavior.

* Many default keyboard shortcuts are not suitable for Linux users. For example:

  * `Ctrl+D` is used for page-down scrolling in Vim, but in tty7 it is mapped to "split right".
  * In Linux, `Alt+1/2/3/4` is commonly used to switch between different windows or panes, but tty7 uses `Ctrl+1/2/3/4` by default.

* The most serious issue is the paste problem. For example, when I copy a piece of code and paste it into Vim on a Linux system, tty7 inserts an additional blank line after every line of code. This is a very serious bug because it affects code editing.

* After connecting to a Linux system, the "Git Changes" section in the Detail Panel cannot display the Git status of the remote Linux system. tty7 only shows the Git status of the local Windows system.

Here is the code I used for testing:

```json
{
  "env": {
    "ANTHROPIC_BASE_URL": "https://api.moonshot.cn/anthropic",
    "ANTHROPIC_AUTH_TOKEN": "sk-xxx",
    "ANTHROPIC_MODEL": "kimi-k3[1m]",
    "ANTHROPIC_DEFAULT_OPUS_MODEL": "kimi-k3[1m]",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "kimi-k3[1m]",
    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "kimi-k3[1m]",
[...]

Issue #163, opened 2026-07-25
快捷键的行为被改变了，ctrl+j 和 ctrl+r 失效了


sunznx:
### Type

Bug — something doesn't work as expected

### What's going on?

1. 默认的 ctrl+j 是回车，现在按 ctrl+j 没反应了
2. zsh 安装了 percol，唤起的快捷键是 ctrl+r，现在被覆盖了

### Platform (bugs)

None

PR #270, opened 2026-07-30
fix(input): stop the Windows/Linux keymap and CRLF pastes from breaking the shell

l0ng-ai:
Two problems reported in #269 by a Windows user driving a remote Linux box, both from code that only ever had macOS in mind.

## Pasting into vim doubled every line break

A bracketed paste forwarded the clipboard byte for byte (`src/terminal/view.rs`). Text copied on Windows carries `\r\n`, and vim counts CR and LF as two line breaks, so pasting a block of code left a blank line under every line of it.

Bracketed pastes now fold `\r\n` down to `\n`, which is exactly what the same paste already produced on Linux and macOS — the fix makes Windows take the path that already works everywhere else rather than inventing a third behaviour.

The non-bracketed branch is untouched. Without a paste mode to tell pasted text from typing, a line break still has to arrive as the CR the Return key sends.

## The default keymap took keys the shell needs

`secondary` resolves to Cmd on macOS and **Ctrl** everywhere else, and the default table was carried over from the macOS one unchanged. That put window actions directly on top of terminal control codes:

| binding | key off macOS | what it swallowed |
|---|---|---|
| FocusPrevPane | Ctrl+[ | **ESC** |
| FocusNextPane | Ctrl+] | 0x1D |
| SplitRight | Ctrl+D | EOF |
| CloseActiveTab | Ctrl+W | delete-word |
| ClearScrollback | Ctrl+K | kill-line |
| TogglePalette | Ctrl+P | previous-history |
| ToggleRightPanel | Ctrl+J | LF (#163) |
| NewTab | Ctrl+T | transpose-chars |
| Quit / EditorSave | Ctrl+Q / Ctrl+S | XON / XOFF |

These are window-level bindings with no key context, and GPUI matches bindings before it dispatches `on_key_down`, so the terminal never saw the key. `apply_readline_ctrl` already sends `0x04` for `^D` on an empty line — that code was simply unreachable off macOS.

### The rule going in

Off macOS, `ctrl-<letter>`, `ctrl-[`, `ctrl-]`, `ctrl-\` and `ctrl-space` belong to the terminal; window actions live on `ctrl-shift-*`. That is the convention GNOME Terminal, Konsole, Windows Terminal and WezTerm already share, so it should read as familiar rather than as tty7 inventing a layout.

| action | was | now |
|---|---|---|
| Focus previous / next pane | Ctrl+[ / Ctrl+] | Ctrl+Shift+[ / Ctrl+Shift+] |
| Split right / down | Ctrl+D / Ctrl+Shift+D | Ctrl+Shift+D / Ctrl+Alt+Shift+D |
| Close tab | Ctrl+W | Ctrl+Shift+W |
| New tab | Ctrl+T | Ctrl+Shift+T |
| Reopen closed tab | Ctrl+Shift+T | Alt+Shift+T |
| Clear scrollback | Ctrl+K | Ctrl+Shift+K |
| Command palette | Ctrl+P | Ctrl+Shift+P |
| Toggle right panel | Ctrl+J | Ctrl+Shift+J |
| Toggle left panel | unbound | Ctrl+Shift+B |
| Quit | Ctrl+Q | Ctrl+Shift+Q |
| Fullscreen | Ctrl+Enter | F11 |
| Activate tab 1-9 | Ctrl+1-9 | Alt+1-9 |
| Focus pane by direction | Ctrl+Alt+arrow | Alt+arrow |

Three of those are worth explaining:

- **Ctrl+S stays Ctrl+S.** It only means anything in the code panel, so instead of moving it the handler now calls `cx.propagate()` when the editor does not have focus, and the keystroke falls through to the terminal as XOFF.
- **Alt+1-9 and Alt+arrow** cost readline its `M-<digit>` numeric argument. GNOME Terminal makes the same trade, and it is what the reporter asked for.
- **Ctrl+Alt+arrow had to go** regardless — it is workspace switching under GNOME, so tty7 was never going to win that one.

Ctrl+C, Ctrl+V and Ctrl+X are unchanged: they are handled in `on_key_down` rather than as bindings, and Ctrl+C still copies only when there is a selection and sends SIGINT otherwise. **Shift+Insert** is added as a paste binding, also from #269.

macOS defaults are byte-for-byte identical. Anything a user rebound themselves still wins — only the defaults moved.

## Guards

`no_default_binding_sits_on_a_terminal_control_code` walks the default table, parses every chord, and fails on any bare `ctrl-<letter>` or `ctrl-[ ] \ space`. The whole class of bug cannot come back silently.

`every_default_chord_is_claimed_by_exactly_one_action` fails when two actions claim the same chord. It earned its place during the reshuffle: moving NewTab onto Ctrl+Shift+T collided with ReopenClosedTab, which is how ReopenClosedTab ended up on Alt+Shift+T.

## Testing

`cargo test --bin tty7` — 854 passed, 0 failed. `cargo fmt --check` clean.

The guard tests are vacuous on macOS by construction (`secondary` parses to Cmd there), so the Windows and Linux CI runners are the ones that actually enforce them.

Refs #269

---

# ⌃R searched the wrong machine

Third item from #269, added after the first review pass.

`src/terminal/history.rs` had no notion of *where* a command had run. `load()` read one file, `append()` wrote one file, so an ssh session was offered the commands typed on the local machine. That is worse than an empty menu — the entries look plausible until one of them runs against the wrong box. Path completion had already learned about remote hosts; the history layer never did.

## Scoping

History is now keyed by machine. Local keeps its existing path so nobody loses what they have. Each remote gets a file under `history.d/`, named after the target with a hash appended — a label that is not a filename (`me@box:/srv/../weird`) can neither escape the directory nor collide with another host.

A pane derives its scope from the innermost remote it can see: the ssh target reported by shell integration, otherwise its host id, otherwise local. When that changes, the pane clears the menu and reloads in the background rather than showing the previous machine's entries while it waits, and a reload that lands after the pane has moved on again is discarded.

## Reading the far end's own history

On a remote workspace there is a `Host` impl, so `~/.zsh_history` and `~/.bash_history` are read through the same channel git status and file listings already use (`HostLinks::home` supplies the remote home). The first ⌃R on a freshly connected box therefore has something in it rather than starting empty. Reads are capped at 4 MiB and happen off the UI thread.

Bare `ssh` has no `Host` to read through and falls back to what tty7 recorded itself for that target. That is less than a remote workspace gives you, but it is the right machine's commands.

## Tests

- `a_remote_scope_never_serves_the_local_machine_s_history` — the reported bug, in both directions.
- `two_remotes_keep_their_own_stores` — no cross-host leakage.
- `a_remote_scope_reads_the_far_end_s_own_shell_history` — zsh and bash files from the far end parse and land, with timestamps.
- `a_label_that_is_not_a_filename_still_gets_its_own_file` — path traversal, collisions, length.
- `an_empty_label_falls_back_to_local_rather_than_a_nameless_file`.

`cargo test --bin tty7` — 859 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Windows and Linux stop taking keys the shell needs** — `secondary` means
+  Cmd on macOS and Ctrl everywhere else, and the default keymap was carried over
+  from macOS unchanged. That put window actions straight on top of terminal
+  control codes: Ctrl+D could not send EOF (it split the pane), Ctrl+[ could not
+  send ESC (it cycled panes), and Ctrl+W, Ctrl+K, Ctrl+P, Ctrl+J, Ctrl+T,
+  Ctrl+Q and Ctrl+S were all spoken for. These are window-level bindings with no
+  context, so they matched before the terminal ever saw the key — the code that
+  sends EOF was there, just unreachable.
+
+  Off macOS the rule is now that `ctrl-<letter>`, `ctrl-[`, `ctrl-]`, `ctrl-\`
+  and `ctrl-space` belong to the terminal, and window actions live on
+  `ctrl-shift-*` — the convention GNOME Terminal, Konsole, Windows Terminal and
+  WezTerm already share. A test enforces it, so the next binding added cannot
+  quietly reintroduce the problem.
+
+  | action | was | now |
+  |---|---|---|
+  | Focus previous / next pane | Ctrl+[ / Ctrl+] | Ctrl+Shift+[ / Ctrl+Shift+] |
+  | Split right / down | Ctrl+D / Ctrl+Shift+D | Ctrl+Shift+D / Ctrl+Alt+Shift+D |
+  | Close tab | Ctrl+W | Ctrl+Shift+W |
+  | New tab | Ctrl+T | Ctrl+Shift+T |
+  | Reopen closed tab | Ctrl+Shift+T | Alt+Shift+T |
+  | Clear scrollback | Ctrl+K | Ctrl+Shift+K |
+  | Command palette | Ctrl+P | Ctrl+Shift+P |
+  | Toggle right panel | Ctrl+J | Ctrl+Shift+J |
+  | Toggle left panel | unbound | Ctrl+Shift+B |
+  | Quit | Ctrl+Q | Ctrl+Shift+Q |
+  | Fullscreen | Ctrl+Enter | F11 |
+  | Activate tab 1-9 | Ctrl+1-9 | Alt+1-9 |
+  | Focus pane by direction | Ctrl+Alt+arrow | Alt+arrow |
+
+  Ctrl+S keeps saving in the code panel but now falls through to the terminal
+  when the editor does not have focus, so a shell still receives XOFF.
+  Ctrl+C, Ctrl+V and Ctrl+X are unchanged — Ctrl+C still copies only when there
+  is a selection and sends SIGINT otherwise — and **Shift+Insert** now pastes.
+  macOS bindings are untouched. Anything you rebound yourself still wins; only
+  the defaults moved. (#269)
+
 - **The machine that runs your panes now owns their layout** — the workspace,
   tab and pane tree has moved out of the app and into the background service, so
   one machine has one tree that every client of it reads: the window on it, a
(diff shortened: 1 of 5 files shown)


l0ng-ai:
Acceptance review of the two commits, and two fixes pushed on top:

1. **The Linux runner was already failing** — `no_default_binding_sits_on_a_terminal_control_code` contradicted the PR's own design decision to keep `EditorSave` on Ctrl+S (the handler propagates when the editor lacks focus, so the terminal still gets XOFF — verified against gpui's dispatch: after all matched action handlers propagate, the raw key-down is delivered to the tree). The guard now carries a named exemption list with the earn-a-spot condition documented.

2. **Wrong-box seeding** — on a remote workspace, a pane that ssh'ed onward is scoped to the *inner* target, but `remote_shell_history_sources` still read the workspace host's `~/.zsh_history` into that scope. Seeding is now reserved for the host's own scope; inner-ssh panes fall back to tty7's own records for that target, same as bare ssh.

`cargo test --bin tty7` on Windows: 742 passed, 0 failed (the guard tests are live here, secondary = Ctrl). `cargo fmt --check` clean.

Non-blocking notes for follow-up: the new `alt-left`/`alt-right` defaults also cost the prompt editor's Alt+arrow word motion off macOS (Ctrl+arrows still work); `shift-insert` is installed outside the rebindable table so Settings can't see or reassign it; the ctrl-chord guard could also reject ctrl-2…8 (NUL, ESC, FS, GS, RS, US, DEL).

Issue #213, opened 2026-07-27
Support Kitty Image Protocol or iTerm Image Protocol


fcharlie:
### Type

Idea — suggest an improvement or new capability

### What's going on?

I gave it a try; tty7 does not support the Kitty or iTerm image protocols.

### tty7 version (bugs)

26.7.4

### Platform (bugs)

None

ayamir:
@l0ng-ai  If so, tty7 can support almost all of graphics capability which make web browser possible in tty7 by using: https://github.com/zenbu-labs/terminal-browser.

Demo in kitty here, really cool :>

<img width="3024" height="1898" alt="Image" src="https://github.com/user-attachments/assets/e692a1b0-9458-4ed9-9d60-8217cb56f05e" />

I can try to realize it in my spare time.

l0ng-ai:
> [@l0ng-ai](https://github.com/l0ng-ai) If so, tty7 can support almost all of graphics capability which make web browser possible in tty7 by using: https://github.com/zenbu-labs/terminal-browser.
> 
> Demo in kitty here, really cool :>
> 
[...]

PR #272, opened 2026-07-31
feat(graphics): render kitty graphics with shared-memory transport (#213)

ayamir:
## What

Add [kitty graphics protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/) support to tty7, so image-emitting TUIs — notably [`terminal-browser`](https://github.com/zenbu-labs/terminal-browser) — render inside a pane (issue #213).

## How

- **Daemon-side interception.** Kitty graphics (`\x1b_G…`) sequences are lifted out of the PTY stream in the pane reader *before* they reach the replay ring or the client's VT parser. On the common no-graphics chunk the sniff is zero-copy; only a chunk that actually carries `_G` allocates. `a=q` query replies are written straight back to the PTY so probing senders see support.
- **Out-of-band frames.** Images and deletes are forwarded to the client as compact binary `Image`/`DeleteImage` daemon messages, interleaved with `Output` in stream order so an image lands at the same cursor cell the sender drew it at.
- **Off-thread decode.** The client inflates + BGRA-swaps + places on a dedicated worker with newest-frame-per-id coalescing, so a full-window browser frame (~42 ms to inflate) never blocks PTY output or scrolling. The paint path reads the resulting sprite atlas.
- **Local fast path.** For local panes, file (`t=f`/`t=t`) and POSIX shared-memory (`t=s`) transfer are honored: the daemon reads (and unlinks) the object and hands the client raw pixels, skipping the client-side zlib inflate the compressed-inline path would force. This is the path senders like `terminal-browser` prefer and closes the smoothness gap with native terminals. Remote (SSH) panes keep refusing shm/file, so pixels ride the tunnel compressed-inline.

## Testing

- `cargo test -p tty7-core kitty_graphics` — 27 passed
- reader interception/forwarding tests (`reader_strips_graphics_and_forwards_them_out_of_band`, `reader_forwards_graphics_deletes`) — passing
- Verified end-to-end by running `terminal-browser` in a locally built/installed tty7: pages render full-pane at correct scale with smooth scrolling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9587,6 +9587,7 @@ dependencies = [
  "libgssapi",
  "log",
  "memchr",
+ "miniz_oxide",
  "notify 8.2.0",
  "portable-pty",
  "russh",
--- a/crates/tty7-core/Cargo.toml
+++ b/crates/tty7-core/Cargo.toml
@@ -112,6 +112,12 @@ memchr = "2"
 # roughly 4 MB. Base64 costs 1.33× instead. Already in the tree via russh.
 base64 = "0.22"
 
+# zlib inflate for the kitty graphics protocol's `o=z` (compressed) pixel
+# payloads (`core::kitty_graphics`). Pure-Rust, no new native code, and already
+# in the tree transitively — the same crate `terminal-browser` uses for the
+# deflate side, so the two ends agree on the wire format.
+miniz_oxide = "0.8"
+
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
 # GSSAPI/Kerberos SSH auth — see the `gssapi` feature below for why it is
--- a/crates/tty7-core/src/core/kitty_graphics.rs
+++ b/crates/tty7-core/src/core/kitty_graphics.rs
@@ -1,0 +1,1558 @@
+//! Streaming kitty-graphics-protocol extractor (APC `ESC _ G … ST`).
+//!
+//! The daemon-side counterpart to [`crate::core::osc`]: a tiny, resumable state
+//! machine that pulls kitty graphics commands out of the raw PTY byte stream
+//! *without* running a full VT parser. It exists for the same reason the OSC
+//! sniffer does — the client's `alacritty_terminal` fork silently discards APC
+//! sequences (`ESC _` routes to vte's `SosPmApcString`, whose payload bytes hit
+//! a `_ => ()` arm), so an image transmitted this way never surfaces as a `Term`
+//! event. We tap the bytes here instead.
+//!
+//! This lives daemon-side so the pixel payload can be lifted out of the stream
+//! *before* it enters the replay ring — a full-window RGBA frame is hundreds of
+//! KB, and letting it accumulate in the ring would make reattach replay
+//! catastrophic. The daemon decodes here and forwards a compact out-of-band
+//! frame to the client; the base64 text never rides the socket, and the VT
+//! parser never has to chew through it.
+//!
+//! Scope: this handles the subset the wire actually carries in practice
+//! (transmit-and-display, query, delete), with the two transmission media that
+//! survive tty7's socket + SSH topology — direct (`t=d`, base64 inline) only;
+//! shared-memory (`t=s`) and file (`t=f`) probes are answered *unsupported* so a
+//! sender like `terminal-browser` falls back to `t=d` on its own.
+//!
+//! Protocol reference: <https://sw.kovidgoyal.net/kitty/graphics-protocol/>
+
+use base64::Engine as _;
+use base64::engine::general_purpose::STANDARD as BASE64;
+use serde::{Deserialize, Serialize};
+
+/// Cap on a single APC payload (one chunk) we'll buffer before abandoning it.
+/// `terminal-browser` chunks direct transmissions into 4 KiB base64 pieces;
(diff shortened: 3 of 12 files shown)


ayamir:
Image: 

image

Terminal browser showcase:


https://github.com/user-attachments/assets/3055f7c7-7336-4bcc-accd-f20049e99347

ayamir:
Thanks for the review — both addressed in the amended commit (`768ae00`).

**1. Unbounded zlib inflate.** `Image::to_rgba8()` now inflates with `decompress_to_vec_zlib_with_limit`, bounded by the pixels the sender *declares* (`width * height * bytes_per_pixel` for the wire format). A payload that decompresses past its own claimed dimensions is malformed and gets dropped (`None`) before it can allocate, so the ~1000:1 deflate → tens-of-GB OOM is closed. Added `to_rgba8_bounds_the_inflate_by_declared_dimensions` covering both the rejected bomb and a well-formed compressed image still decoding.

**2. Image frame exceeding `MAX_FRAME`.** Both push sites (inline `Segment::Image` and the resolved shm/file `ImageFromMedium`) now go through a `push_image_frame` helper that drops any encoded frame larger than `MAX_FRAME` instead of queueing it. So the single oversized image is dropped rather than reaching `write_frame`, whose error the writer loop treats as fatal and would use to disconnect the client from a healthy pane. Added `push_image_frame_drops_over_max_frame`.

Tests: `cargo test -p tty7-core kitty_graphics` (28 passed) and the pane reader/graphics tests all green.

l0ng-ai:
Thanks — both of those look right, and `push_image_frame` is a better spot for the frame cap than where I'd have put it.

A second pass turned up more, so rather than send you another round trip I pushed the fixes to this branch as `2a2771c`. Happy to split any of it out or drop it if you disagree with a call.

**One gap left in the inflate bound.** `decoded_len()` is the sender's own declared size, so it's attacker-chosen too: `s=65535,v=65535` works out to a ~17 GB budget, which hands the bomb back the allocation the bound was meant to deny. Now clamped to `MAX_IMAGE_BYTES` as well.

**Filesystem, local pane only:**

- `t=t` unlinked whatever path the escape named. `name` is arbitrary and attacker-reachable — `cat`ing a hostile file is enough to emit one — so a single sequence could delete `~/.ssh/id_ed25519`. Now restricted to a temp dir, compared after `canonicalize()` so `/tmp/../home/...` can't dress a path up as one. This is the one the kitty spec calls out explicitly.
- `t=f` used `fs::read`, unbounded and on any file type. Pointed at `/dev/zero` it never returns — on the daemon's reader thread, so the pane's whole output path wedges. Regular files only, size-capped now.
- `t=s` passed the name straight to `shm_open`/`shm_unlink`; now requires a well-formed POSIX name so `/../../etc/passwd` can't escape the shm namespace where a platform resolves it against the filesystem.

**Silent drops:**

- Windows local panes answered `t=s`/`t=f` probes `OK` while `MediumTransfer::resolve` is `#[cfg(unix)]` and returns `None` — so a sender took the offer and every frame was discarded, pane blank. `query_reply` and `finalize` now share one predicate that includes the platform axis. The two local-pane tests are `#[cfg(unix)]` now, which is why they weren't catching it.
- `resize()`'s early-out only compared cols/rows. Since this PR reports cell size in *device* pixels, moving a window between a 2x and a 1x display changes `ws_xpixel`/`ws_ypixel` with the grid unchanged — the resize was skipped and the child kept rendering for the old framebuffer.
- `a=p` fell through `accept_chunk` and emitted an `Image` with an empty payload. Dropped instead, since placement needs a stored-image table this parser deliberately doesn't keep.
- A foreign APC that never sent its ST swallowed every following `ESC _G` (forwarded as APC text, which the client's vte then discards). And in `ApcGraphicsEsc`/`ApcDropEsc` the `_ =>` arm ate the ESC along with the byte, turning a following `\x1b[31m` into literal `31m`. All three states resync the same way now.

**Smaller:** `q=` was parsed and never honored, so a sender asking for silence got a reply written back to its pty as if typed; `S=` past the end discarded the frame where kitty truncates; real-alpha frames were blended as if premultiplied; and `snapshot()`/`take_retired()` were ordered so the paint path could evict an image from the atlas in the same frame it painted it. `MAX_APC_PAYLOAD`/`MAX_TRANSMISSION_BASE64` are now pinned to `MAX_FRAME` by a test instead of by comment.

Also ran `cargo fmt` — the rustfmt job was red on `768ae00`.

l0ng-ai:
Correction on one item from my last comment — the premultiply change was wrong and I've dropped it (`55a2ef7`, force-pushed over `2a2771c`).

`RenderImage` holds **straight** alpha, not premultiplied. gpui's own producers make that explicit: [`swap_rgba_pa_to_bgra`](https://github.com/zed-industries/zed/blob/main/crates/gpui/src/color.rs) divides the color channels back out by alpha, and both the CoreGraphics text rasterizer and the SVG renderer run their premultiplied output through it on the way into the atlas. So the plain R↔B swap this PR had was already correct, and premultiplying would have darkened every translucent pixel of an `f=100` PNG. `decodes_rgba_and_swaps_to_bgra` caught it on Linux and Windows; the comment there is now the only thing I changed, to say why it's a swap and nothing more.

Everything else in that commit stands — full workspace suite is green locally (`cargo test --workspace`), which is what I should have run the first time instead of just `-p tty7-core`.

PR #276, opened 2026-07-31
fix(completion): escape inserted candidates, close spent menus, keep PATH local

l0ng-ai:
Fixes three defects in the inline completion menu, each of which produced something *wrong* rather than merely unhelpful.

| | Before | After |
|---|---|---|
| `cd My<Tab>` (dir is `My Documents`) | `cd My Documents/` — resplits into two arguments, command breaks | `cd My\ Documents/` |
| `cd ~/My<Tab>` | `cd ~/My Documents/` | `cd ~/My\ Documents/` — the `~/` stays expandable |
| Two candidates sharing the prefix `My ` | Writes `cd My ` — a broken line, and the space closes the menu on the next keystroke | Leaves the line alone and steps through the candidates |
| `git ckout<Tab>` | Menu opens empty waiting on git's alias generator, never closes, swallows every later Tab | Last generator to answer closes a menu that still has nothing in it, so the next Tab reaches the shell |
| `system_prof<Tab>` in an SSH pane to Linux | Offers macOS's `system_profiler` | Offers shell builtins only; the position falls through to the remote's own compsys |

## Why

**Escaping.** Candidates were inserted verbatim. `shell_escape_path` already existed for drag-and-drop paths but completion never reached for it. `escape_candidate` wraps it and leaves a leading `~/` alone — that prefix is the user's own text, and escaping it would defeat the home expansion it was typed for.

The same check now gates the common-prefix step. A prefix that needs escaping can't be written raw, and writing `My ` would also put a space inside the open word — which closes the menu on the next keystroke and leaves the user worse off than before they pressed Tab.

**Spent menus.** A position with no static candidates but a live generator opens an empty session on purpose and waits (this is what makes `git checkout <Tab>` work). The callback that would have closed it returned early on an empty result, so a menu whose generators all came back with nothing never learned they were done. Sessions now count their in-flight generators; the last one to answer closes a menu that is still empty.

**Remote PATH.** The remote isolation in `08ca3a3` covered paths and generators but deliberately kept command completion running — right for the builtins half (`echo` is true on any POSIX shell), wrong for the PATH half. It also failed *inconsistently*: with no local match the position already fell through to the remote's compsys and answered correctly, so the bug only surfaced when this machine happened to have a match, which made it hard to notice.

## Scope

Quoting is fixed on the **insert** side only. The parser still splits words on bare whitespace, so `cd "My Doc<Tab>`, `cat a b.t<Tab>`, `cd ~roo<Tab>` and `cd $HOME/<Tab>` still yield nothing and hand the line to the shell — unchanged behaviour, and a proper lexer is a separate change. Wrapper commands (`sudo git <Tab>` offering filenames instead of git's subcommands) are also left for a follow-up.

## Testing

- `escape_candidate_quotes_what_the_shell_would_resplit` — plain text and flags pass through, spaces/metacharacters escape, `~/` prefix survives, a bare `~foo` does not
- `accepting_a_candidate_escapes_it_for_the_shell` — insertion for a dir, a `~/` dir, and a flag
- `a_candidate_needing_escapes_is_never_half_applied` — the `My ` common prefix is not written
- `a_generator_that_supplies_no_match_closes_the_menu` / `generator_results_that_match_nothing_close_the_menu` — an empty menu does not stay armed
- `a_menu_waits_while_another_generator_is_still_in_flight` — one empty generator does not close a menu another may still fill
- `a_remote_pane_offers_builtins_but_never_this_machines_binaries` — remote candidates are all builtins; a local pane still scans PATH. Before the fix this failed with 200+ names from this machine (`lazygit`, `langchain`, …)
- Full workspace suite green (882 + 757 + …), `cargo fmt --check` clean, no new clippy warnings

### DIFF
--- a/src/terminal/completion.rs
+++ b/src/terminal/completion.rs
@@ -84,16 +84,12 @@ fn current_command(chars: &[char], word_start: usize) -> Option<String> {
 pub fn complete(line: &str, cursor: usize, cwd: Option<&Path>) -> Option<Completion> {
     let chars: Vec<char> = line.chars().collect();
     let cursor = cursor.min(chars.len());
-
-    let mut word_start = cursor;
-    while word_start > 0 && !chars[word_start - 1].is_whitespace() {
-        word_start -= 1;
-    }
+    let word_start = shell_word_start(&chars, cursor);
     let word: String = chars[word_start..cursor].iter().collect();
 
     let is_command = chars[..word_start].iter().all(|c| c.is_whitespace());
     let (word_cands, pending) = if is_command && !word.contains('/') {
-        (complete_command(&word), Vec::new())
+        (complete_command(&word, cwd.is_some()), Vec::new())
     } else {
         match complete_signature(&chars, word_start, &word, cwd) {
             Some(sig) => (sig.cands, sig.pending),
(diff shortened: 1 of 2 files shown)


l0ng-ai:
Added escaped-word parsing and path lookup unescaping, so Tab can complete inside a directory after inserting an escaped candidate such as `My\\ Documents/`. This is covered for local completion and also shared by remote path requests.

PR #279, opened 2026-07-31
fix(shell): a forwarding .bash_profile must not source .bashrc twice

l0ng-ai:
Move `~/.bashrc` into the login-shell chain the bash rcfile replays, so a `~/.bash_profile` that forwards to it no longer runs the user's whole `~/.bashrc` twice.

| | Before | After |
|---|---|---|
| `~/.bash_profile` forwards to `~/.bashrc` (the common case) | `~/.bashrc` sourced **twice** | sourced once |
| `~/.bash_profile` exists, does not forward | `~/.bashrc` sourced once | not sourced — matches a real login shell |
| No profile file at all, only `~/.bashrc` | sourced once | sourced once (fallback kept) |

## Why

tty7 spawns bash non-login so that `--rcfile` is honored at all, and the rcfile replays what a login shell would have sourced. But it ran the profile chain *and then* sourced `~/.bashrc` unconditionally on top. A login shell never sources `~/.bashrc` itself — it only ever arrives because the profile that won the chain forwarded to it, and a `~/.bash_profile` whose entire contents are that forward is close to universal.

So for most users every pane ran their whole `~/.bashrc` twice. Visible symptoms: `fastfetch`-style banners printing twice, completion scripts sourced twice, anything appending to `PROMPT_COMMAND` stacking up, and any `eval` with side effects running again.

Reproduced on macOS with a stock forwarding `~/.bash_profile`; the new test pins it at exactly one sourcing.

Note this is why zsh is unaffected: the zsh path uses `ZDOTDIR` redirectors, so zsh itself decides which startup files apply. bash has no `ZDOTDIR`, so the chain has to be replayed by hand — and the hand-written version had this extra line.

## Testing

- New `a_forwarding_bash_profile_does_not_pull_bashrc_in_twice`: writes a throwaway `$HOME` with a forwarding `.bash_profile` and a counting `.bashrc`, runs real `bash --rcfile … -i`, asserts exactly one sourcing. Unix-only, skips cleanly if bash can't start — same shape as the existing `parse_check` tests.
- Verified the new test actually guards the fix: it reports `got 2` against the previous rcfile.
- `cargo test -p tty7-core --lib shell_integration` — 39 passed.
- `cargo fmt --check` clean.

The pre-existing `bash_rcfile_sources_user_config_then_appends_integration` only asserted the rcfile *mentions* `~/.bashrc`, which the buggy version satisfied too; it is kept as-is and the behavioral test carries the guard.

### DIFF
--- a/crates/tty7-core/src/daemon/shell_integration.rs
+++ b/crates/tty7-core/src/daemon/shell_integration.rs
@@ -745,6 +745,28 @@ fn bash_rcfile() -> String {
 # Replays what a real *login* shell would have sourced, in the same order —
 # necessary because tty7 spawns bash non-login (see shell_integration.rs) so
 # that --rcfile is honored at all; bash silently ignores it for login shells.
+#
+# ~/.bashrc needs both halves of a rule that cannot be written as a static
+# order, because the two shapes in the wild want opposite things:
+#
+#   * a ~/.bash_profile that forwards to ~/.bashrc (the common case, and what
+#     every "put this in your .bash_profile" guide produces). Sourcing ~/.bashrc
+#     again afterwards runs the user's whole file twice: banners print twice,
+#     completions get sourced twice, anything appending to PROMPT_COMMAND stacks.
+#   * a ~/.bash_profile that does *not* forward — written by someone whose
+#     terminal spawns non-login shells, so ~/.bashrc always arrived on its own
+#     and the profile only ever had to hold login-time settings. Dropping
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Updated to preserve both cases: profiles that forward to `~/.bashrc` no longer source it twice, while non-forwarding profiles still receive it once. Added coverage for both `.`/`source` forwarding and non-forwarding profiles.

PR #289, opened 2026-08-01
fix(terminal): render ANSI text decorations

l0ng-ai:
## Summary

- render ANSI SGR 9 strikeout, including decorated blank cells
- draw SGR 4:2 double underline as two distinct strokes
- preserve SGR 4:4 dotted and SGR 4:5 dashed underline patterns
- align custom underline geometry to device pixels and merge contiguous cell runs
- preserve decorations on native box-drawing and powerline glyphs

Follow-up to #288.

## Testing

- `cargo fmt --check`
- `cargo test -p tty7 terminal::element::tests --bin tty7-app` (42 passed)
- `cargo test -p tty7 --bin tty7-app` (892 passed, 3 ignored)


l0ng-ai:
Updated the renderer: dotted/dashed cadence is now measured in logical pixels (while stroke thickness remains device-pixel snapped), ordinary underlines now cover blank cells consistently, and special underlines no longer emit redundant shaped blanks for native box-drawing cells.

PR #292, opened 2026-08-01
fix(windows): prevent daemon from inheriting policy that blocks scoop junctions

ArnoChenFx:
## Background

When tty7 is launched through certain Windows shell brokers, the GUI process can
inherit an enforced `ProcessRedirectionTrustPolicy`.

That policy is inherited by the persistent daemon and the shells running inside
ConPTY. As a result, PowerShell cannot traverse user-created junctions such as
Scoop's `current` links, causing shims including `oh-my-posh` and `fzf` to fail
with errors such as:

- `Shim: Could not determine if target is a GUI app`
- `Shim: Could not create process`

Windows Terminal is unaffected because its process tree does not inherit the
same enforced policy.

## Root Cause

`ProcessRedirectionTrustPolicy` cannot be relaxed after it has been enabled.
Starting the daemon with the regular process creation path therefore propagates
the policy into every terminal session created by that daemon.

The fix must prevent the daemon from inheriting the policy in the first place.

### Windows daemon startup

- Detect whether the current process has the enforced Redirection Trust bit.
- When enforced, create the daemon with `STARTUPINFOEXW` and
  `PROC_THREAD_ATTRIBUTE_PARENT_PROCESS`.
- Use the interactive desktop shell as the daemon's logical parent.
- Preserve the existing process creation path when the policy is not enforced.
- Keep the Windows-specific implementation isolated in
  `daemon/spawn/windows.rs`.

This gives the daemon the normal desktop process policy while preserving the
current environment and command-line arguments.

### Regression coverage

Add a Windows regression test that:

1. Enables Redirection Trust inside a disposable test process.
2. Confirms that an ordinary child reports the inherited policy as `ENFORCED`.
3. Creates a real directory junction.
4. Starts another child through the clean logical-parent path.
5. Confirms that the child reports `CLEAN_OK` and can traverse the junction.

The test checks the inherited policy directly instead of assuming every CI
runner must reject its own junctions. This keeps it deterministic on elevated
GitHub-hosted Windows runners.

### ConPTY exit ordering

The Windows process-exit monitor can observe a short-lived shell exiting before
the ConPTY reader has delivered its final output. This allowed `Exited` to reach
clients before the last output frame.

The monitor now gives the ConPTY reader a bounded drain window before publishing
the fallback exit event. This preserves final-output ordering while retaining a
timeout fallback for pipes that never reach EOF.

### Integration-test cleanup

The attach/replay integration test now closes its streaming session explicitly
and removes the pane through a separate `PaneClient` connection. This keeps the
test focused on replay behavior and avoids a Windows shutdown race during test
cleanup.

### The daemon has no standard handles on this path

Naming a logical parent makes handle inheritance follow *that* process, so
tty7 cannot hand the child a `NUL` handle the way the ordinary
`Stdio::null()` path does. The daemon therefore starts with NULL standard
handles — and `run_daemon` wrote two startup notes with `eprintln!`, which
panics when the write fails. The daemon died before serving anything, on
exactly the machines this PR targets.

Those notes, and the `TTY7_TRACE` line in the pane reader, now go through a
write that tolerates a missing stderr.

### Falling back instead of refusing to start

`GetShellWindow` returns NULL whenever there is no interactive Explorer to
borrow — it is restarting, the shell was replaced, the session has no
desktop — and `OpenProcess` can be refused outright. The clean-parent path
now logs a warning and degrades to the ordinary spawn instead of failing
the launch: losing it only costs junction traversal inside the shells.

### The exit fallback is real again

The drain window described above was not actually in the code; the monitor
released the pseudoconsole and returned. The reader is still the preferred
reporter, since it publishes `Exited` only after forwarding everything up
to EOF — but EOF is not guaranteed. A grandchild holding the ConPTY output
pipe open keeps `ClosePseudoConsole` from ever completing, and the pane
would then read as alive forever. A bounded window now sits behind it;
`DeathReporter` is idempotent, so whichever side gets there first wins.

### Test coverage

- The Redirection Trust regression could pass vacuously: libtest exits 0
  when `--exact` matches nothing, so a stale `TEST_NAME` would turn the
  whole test green without running it. It now asserts the inner run happened.
- `push_quoted_arg` is covered — trailing backslashes, embedded quotes, and
  the empty argument, whose doubling rules silently merge arguments when wrong.
- Two Windows tests for the drain window: one where EOF never arrives, one
  where the reader reports first and the monitor neither waits out its
  window nor publishes a second `Exited`.

Note the clean-parent spawn changes the daemon's token: it now derives from
Explorer, so an elevated tty7 starts a medium-integrity daemon. That is the
intent of borrowing the desktop policy, recorded here because it is a
security-boundary change rather than a mitigation-only one.


crates/tty7-core/src/daemon/pane.rs
```diff
@@ -614,6 +624,20 @@ impl DeathReporter {
...
 }
 
+/// Waits briefly for the ConPTY reader to publish the ordered exit event.
+///
+/// The process monitor is only a fallback for Windows pipes that never reach
+/// EOF. If the reader has already claimed the reporter, it is either emitting
+/// the final Output frames or resolving the exit code, so no second report is
+/// needed. Otherwise the deadline preserves the monitor's liveness guarantee.
+#[cfg(windows)]
+fn wait_for_windows_reader_report(death: &DeathReporter) {
```
l0ng-ai: The grace period does not establish the claimed ordering guarantee. If the ConPTY reader is delayed for more than 500ms (or is blocked in `reader.read` while output is still pending), the monitor calls `death.report()` and sends `Exited`. The reader can subsequently acquire `state` and still call `fan_out_output`—there is no `alive`/`reported` guard before forwarding output—so clients can observe `Exited` followed by final `Output` and may discard it. Please synchronize the fallback with actual reader drain/EOF (or otherwise serialize/handle late output), rather than allowing a fixed timeout to claim the reporter.

ArnoChenFx: Fixed as suggested.

PR #293, opened 2026-08-01
fix(ui): soften overlay scrollbars

l0ng-ai:
## Summary

- make tty7's overlay scrollbars use the scrolling-only presentation instead of inheriting the platform's always-visible mode
- reduce the resting and hover thumb contrast so scrollbars sit more quietly over sidebar and panel content

## Why

On macOS, `preferredScrollerStyle` can switch away from overlay mode depending on the system input-device configuration. tty7 mapped that state to `ScrollbarShow::Always`, and gpui-component renders that mode with an 8 px active thumb inside a 16 px hit area. The result was a wide, permanently visible gray bar in the tab sidebar.

These scrollbars are intentionally overlaid on content, so using the scrolling-only presentation keeps them discoverable while scrolling without leaving heavy chrome over the list at rest.

## Verification

- `cargo fmt --check`
- `cargo check -p tty7`
- manually verified the sidebar in a dev build with an isolated config directory

### DIFF
--- a/src/ui/scrollbar.rs
+++ b/src/ui/scrollbar.rs
@@ -19,6 +19,11 @@ pub(crate) fn with_vertical_scrollbar(
                 .left_0()
                 .right_0()
                 .bottom_0()
+                // No `scrollbar_show` override: it falls back to
+                // `cx.theme().scrollbar_show`, which `apply_theme` derives from
+                // `should_auto_hide_scrollbars()` — the OS "show scroll bars"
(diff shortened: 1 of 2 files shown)


l0ng-ai:
Updated to keep the requested softer thumb colors while removing the per-widget `ScrollbarShow::Scrolling` override. These scrollbars now respect the theme setting derived from the OS “show scroll bars” accessibility preference.

PR #298, opened 2026-08-01
fix(core): make Codex hook commands work under PowerShell on Windows

### LINKED ISSUE #297: [Windows] Codex hooks installed by tty7 fail when Codex executes them through PowerShell
# [Windows] Codex hooks installed by tty7 fail when Codex executes them through PowerShell

## Environment

- tty7 26.8.1, Windows 11
- Codex CLI 0.146.0 (npm), sessions launched inside tty7 PowerShell panes

## Symptom

Installing the Codex agent integration writes `~/.codex/hooks.json` with commands like:

```json
"\"C:\\Program Files\\tty7\\tty7-app.exe\" agent-hook codex session-start"
```

On Windows, Codex executes hook commands through the session's environment shell. When the pane shell is PowerShell, the command is run as:

```
pwsh -NoProfile -Command "C:\Program Files\tty7\tty7-app.exe" agent-hook codex session-start
```

PowerShell's `-Command` parsing drops the quotes around the path, so PowerShell tries to run `C:\Program` and exits with code 1. Codex reports `hook: SessionStart Failed` / `hook: UserPromptSubmit Failed`, and tty7 never receives the `session-start` / `prompt-submit` / `stop` events.

## Reproduction

```
pwsh -NoProfile -Command "C:\Program Files\tty7\tty7-app.exe" agent-hook codex session-start
```

Expected: exit 0.
Actual: exit 1, `C:\Program: The term 'C:\Program' is not recognized as a name of a cmdlet...`

Verified working forms:

- `pwsh -NoProfile -Command "& 'C:\Program Files\tty7\tty7-app.exe' agent-hook codex session-start"` → exit 0 (PowerShell only)
- `C:\PROGRA~1\tty7\tty7-app.exe agent-hook codex session-start` → exit 0 under both `cmd.exe` and PowerShell (space-free 8.3 short path)

## Why a manual edit is not enough

`refresh_hooks_at_launch()` (release builds) treats any installed command that differs from `HookTarget::hook_command()` as `Outdated` and reinstalls the buggy template on the next tty7 launch, so users cannot permanently patch `hooks.json` by hand.

## Suggested fix (agreed: Option A — bare executable name via PATH)

In `crates/tty7-core/src/core/agent_hooks.rs`, for local Windows installs, generate the hook command with the bare executable name when it resolves from PATH to the same binary:

```
tty7-app.exe agent-hook codex session-start
```

tty7 already adds its install directory to PATH (`install_cli_on_path`), so the name resolves on a default install. The bare name has no spaces or shell metacharacters, works under `cmd.exe`, PowerShell, and bash, and has no 8.3 short-name dependency. Fall back to the quoted full path when the name does not resolve (non-PATH installs and remote targets remain a follow-up). See PR #298.

Notes for maintainers:

- A PowerShell-only form (`& 'C:\Program Files\tty7\tty7-app.exe' ...`) fixes PowerShell panes but breaks `cmd.exe` panes, so it is not a drop-in template change.
- Remote Windows targets need server-side short-path resolution; the PR covers local installs only.
- Any template change invalidates Codex's persisted hook trust hash (`trusted_hash` in `config.toml`), so users will need to re-trust/reinstall hooks after upgrading — worth a note in the install UX or changelog.

shihuaidexianyu:
## Summary

Fixes #297: on Windows, the Codex hooks generated by tty7 fail when Codex executes them through a PowerShell session.

Codex runs hook commands through the session's environment shell. On a PowerShell pane that means:

```
pwsh -NoProfile -Command "C:\Program Files\tty7\tty7-app.exe" agent-hook codex session-start
```

PowerShell's `-Command` parsing drops the quotes around the path, so PowerShell tries to execute `C:\Program` and exits with code 1. Codex reports `hook: SessionStart Failed` and tty7 never receives the `session-start` / `prompt-submit` / `stop` events.

## Change (Option A, per review discussion)

For local Windows installs, when the executable resolves by its bare file name from PATH — tty7 adds its install directory to PATH via `install_cli_on_path` — generate the hook command without any path quoting:

```
tty7-app.exe agent-hook codex session-start
```

The bare name has no spaces and no shell metacharacters, so it is executed correctly by `cmd.exe`, PowerShell, and bash alike, with no dependency on 8.3 short names. The PATH lookup verifies that the first resolvable match is the same binary (`self.exe`); otherwise the command falls back to the existing quoted full path (unchanged behavior for non-PATH installs and remote targets).

## Verification

- The bare-name command `tty7-app.exe agent-hook codex session-start` exits 0 under both `cmd /C` and `pwsh -NoProfile -Command` (verified with the real binary; the name resolves on a default install).
- The PATH resolution helper is verified: the real exe resolves to `tty7-app.exe`, and a file in a non-PATH directory correctly does not.
- The previous long-path template fails under PowerShell with `C:\Program: The term 'C:\Program' is not recognized...`.
- The previously proposed 8.3 short-path form was dropped per review: it depends on short-name generation and can still contain shell metacharacters (verified: `dev&ops(50%)$` → `DEV&OP~1`).
- `cargo fmt --check` passes; the PATH-resolution logic was verified in a standalone Windows build. Full `cargo check` could not run locally because the machine lacks NASM/CMake (`aws-lc-sys`), so CI is the authoritative build check.

## Notes

- PATH changes only affect newly started processes; users should restart Codex sessions after (re)installing hooks.
- Remote Windows targets need their own PATH handling (left as future work in #297).
- Any template change invalidates Codex's persisted `trusted_hash` for the installed hooks, so users will need to re-trust the hooks after upgrading (worth a note in the install UX or changelog).

### DIFF
--- a/crates/tty7-core/Cargo.toml
+++ b/crates/tty7-core/Cargo.toml
@@ -141,7 +141,6 @@ getrandom = "0.3"
 # tree on hangup (ConPTY's `kill` only reaches the shell itself).
 windows-sys = { version = "0.59", features = [
     "Win32_Foundation",
-    "Win32_Storage_FileSystem",
     "Win32_System_Console",
     "Win32_System_Diagnostics_ToolHelp",
     "Win32_System_Threading",
--- a/crates/tty7-core/Cargo.toml
+++ b/crates/tty7-core/Cargo.toml
@@ -141,6 +141,7 @@ getrandom = "0.3"
 # tree on hangup (ConPTY's `kill` only reaches the shell itself).
 windows-sys = { version = "0.59", features = [
     "Win32_Foundation",
+    "Win32_Storage_FileSystem",
     "Win32_System_Console",
     "Win32_System_Diagnostics_ToolHelp",
     "Win32_System_Threading",
--- a/crates/tty7-core/src/core/agent_hooks.rs
+++ b/crates/tty7-core/src/core/agent_hooks.rs
@@ -1206,9 +1206,12 @@ mod tests {
         let local = local_host();
         let here = HookTarget::local(&*local).expect("home resolves in tests");
         let exe = std::env::current_exe().unwrap();
+        let command_exe = here
+            .hook_command_exe()
+            .unwrap_or_else(|| format!("\"{}\"", exe.display()));
         assert_eq!(
             here.hook_command(HookAgent::Claude, "stop"),
-            format!("\"{}\" agent-hook claude stop", exe.display())
+            format!("{command_exe} agent-hook claude stop")
         );
     }
 
@@ -1250,6 +1253,9 @@ mod tests {
         let exe = exe_json.trim_matches('"').to_string();
         let host = local_host();
         let target = HookTarget::local(&*host).expect("home resolves in tests");
+        let hook_exe_raw = target.hook_command_exe().unwrap_or_else(|| exe_raw.clone());
+        let hook_exe_json = serde_json::to_string(&hook_exe_raw).unwrap();
+        let hook_exe = hook_exe_json.trim_matches('"');
 
         let copilot = copilot_hooks_json(&target).expect("copilot content builds");
         let parsed: serde_json::Value = serde_json::from_str(&copilot).expect("valid JSON");
@@ -1267,11 +1273,11 @@ mod tests {
                 "copilot {event} carries the emitter"
             );
         }
-        assert!(copilot.contains(&exe));
+        assert!(copilot.contains(hook_exe));
 
         let opencode = opencode_plugin_js(&target).expect("opencode content builds");
         assert!(opencode.contains("agent-hook opencode"));
-        assert!(opencode.contains(&exe));
+        assert!(opencode.contains(hook_exe));
         assert!(opencode.contains(r#"process.env["TTY7"]"#));
 
         let pi = pi_extension_ts(&target).expect("pi content builds");
@@ -1300,7 +1306,7 @@ mod tests {
                 "grok {event} matcher"
             );
         }
-        assert!(grok.contains(&exe));
+        assert!(grok.contains(hook_exe));
     }
 
     #[test]
--- a/crates/tty7-core/src/core/agent_hooks.rs
+++ b/crates/tty7-core/src/core/agent_hooks.rs
@@ -354,16 +354,18 @@ impl<'a> HookTarget<'a> {
     /// path containing spaces, so the usual `"C:\Program Files\..."` form is
     /// parsed as `C:\Program` and fails — and even a quoted path without
     /// spaces is a syntax error in PowerShell (invoking a quoted path requires
-    /// the `&` call operator). Resolving the space-free 8.3 short path lets us
-    /// emit the path bare, which both `cmd.exe` and PowerShell execute.
+    /// the `&` call operator). When the executable resolves by its bare file
+    /// name from PATH, we can emit the name without any quoting, which every
+    /// shell (`cmd.exe`, PowerShell, bash) executes correctly.
     ///
-    /// Returns `None` when no space-free short path is available; callers then
-    /// fall back to the quoted long path.
+    /// Returns `None` when the executable is not resolvable by name from PATH,
+    /// or when the first PATH match is a different binary; callers then fall
+    /// back to the quoted full path.
     fn hook_command_exe(&self) -> Option<String> {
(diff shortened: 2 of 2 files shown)


l0ng-ai:
I found two issues with this approach:

1. If 8.3 short names are disabled or unavailable, `GetShortPathNameW` returns no value and the code falls back to the existing quoted long-path command. That is the same form this PR identifies as failing under PowerShell, so Codex hooks remain broken for those installations.
2. The bare-path check only rejects spaces. Windows paths may legally contain PowerShell/cmd metacharacters such as `&`, `$`, `;`, `%`, or `(`, and an 8.3 path can retain them. A path such as `C:\Users\dev&ops\...\tty7-app.exe` would be split or otherwise misparsed by the shell instead of invoking tty7.

shihuaidexianyu:
Thanks for the review — both points are valid, and I've verified them:

- With 8.3 short names disabled, the fallback is the exact quoted long-path form that fails under PowerShell, so those installs stay broken.
- 8.3 short names can retain shell metacharacters. I confirmed on Windows that `dev&ops(50%)$` shortens to `DEV&OP~1` — `&`, `$`, `%` and `(` all survive, so a bare path is still unsafe.

That made me reconsider the approach entirely. Since no single static command string can be parsed correctly by both `cmd.exe` and PowerShell for arbitrary legal paths, I'd like to discuss two directions:

**Option A: bare executable name via PATH**

tty7 already adds its install directory to PATH (`install_cli_on_path: true`), and on a default Windows install `where tty7-app.exe` resolves to the real binary. If hook install/refresh verifies that the PATH-resolved executable is the same binary the hook system uses (`self.exe`), the hook command can simply be:

```
tty7-app.exe agent-hook codex session-start
```

No quotes, no spaces, no metacharacters — works under `cmd.exe`, PowerShell and bash alike, and has no 8.3 dependency at all.

Caveats:
- PATH updates only affect newly started processes, so users would need to restart Codex sessions after (re)installing hooks.
- Remote/SSH machines need their own PATH handling; that could be a follow-up.

**Option B: generate the command per shell**

tty7 knows which shell a pane uses, so the installer could emit `& '<path>'` for PowerShell panes and `"<path>"` for cmd panes (and POSIX-style quoting elsewhere). This is the most general fix for arbitrary paths. The trade-off is that `hooks.json` is per-machine while panes can use different shells, so we'd need a policy — e.g., regenerate when an agent starts in a different shell, or key off the machine's default shell. It's a larger change, but it fully removes the quoting problem.

My preference would be Option A (with the existing quoted form as a documented fallback when PATH resolution fails), but I'm happy to go with whichever you prefer — happy to rewrite the PR along either line.

l0ng-ai:
Thanks — I think Option A is the right direction for local Windows hooks.

shihuaidexianyu:
Rewritten per Option A: for local Windows installs, the hook command now uses the bare executable name when it resolves from PATH to the same binary (\	ty7-app.exe agent-hook codex <event>\), with the quoted full path as the fallback. The 8.3 short-path code and the extra \windows-sys\ feature are removed. Verified locally: the name resolves on a default install, and the bare-name command exits 0 under both \cmd /C\ and \pwsh -NoProfile -Command\.

shihuaidexianyu:
CI follow-up:

The Windows job was failing in two tests because the assertions still assumed that every local hook command contains the quoted full `current_exe()` path. With the new Windows behavior, Cargo's test executable directory is on `PATH`, so the same binary resolves by its bare name and the generated command is intentionally `tty7_core-….exe agent-hook …` instead.

Commit `1382c71` updates the two affected tests:

- `the_hook_command_names_the_binary_on_that_machine` now expects the PATH-resolvable bare name when available, while retaining the quoted-full-path fallback.
- `owned_file_contents_carry_marker_and_exe` uses that same expected hook executable for Copilot/OpenCode/Grok. Pi still checks the full executable path because its bridge invokes `spawnSync` directly rather than using `hook_command`.

`cargo fmt --check` and `git diff --check` pass locally. I also attempted `cargo test -p tty7-core --lib core::agent_hooks::tests`, but the local build stopped in `aws-lc-sys` before the Rust tests ran because NASM is not installed; the newly triggered Windows CI run will be the authoritative test result.

shihuaidexianyu:
Update: the rerun is now fully green. The Windows `Build` and `Test` steps both passed, confirming the two adjusted assertions match the intended PATH-resolvable executable behavior: https://github.com/l0ng-ai/tty7/actions/runs/30711867905

Issue #302, opened 2026-08-02
i18n: multilingual UI/CLI output (en/zh-Hans)


shihuaidexianyu:
Type: Idea

### What's going on?

tty7 currently hardcodes English strings across the GUI and CLI. The project already ships `README.zh-CN.md` and `docs/features.zh-CN.md`, but the product UI and command-line output are English-only. Adding multilingual support would make the tool usable by non-English contributors and users, starting with Simplified Chinese alongside English.

### Proposed behavior

- Add a locale model: follow the system locale by default, with an explicit override such as a `language` field in `config.json` or a Settings control.
- Extract user-facing strings from `src/ui/*`, `src/terminal/*`, and `crates/tty7-cli/*` into a translation layer. Prefer the existing `rust-i18n` dependency already in the tree, or an equally lightweight local lookup that keeps the binary small.
- Keep machine-facing output stable: `--json` field names, protocol strings, and log messages stay unchanged.
- Untranslated keys fall back to English; CLI help, hints, and errors get the same treatment as GUI strings.
- Docs and skills can be localized in a follow-up; the existing zh-CN doc pattern maps naturally to the same effort.

### In scope

- GUI: menus, settings panels, tabs/sidebar labels, palettes, dialogs, toasts, and other user-visible text.
- CLI: help, usage errors, hints, and text tables. JSON output remains locale-neutral.
- Terminal UI overlays generated by `src/terminal/*`.

### Out of scope

- Wire protocol payloads and JSON field names.
- Internal log files.
- Keyboard/IME behavior; only labels and messages are in scope.

### Acceptance criteria

- A user can select a locale from Settings or set it in `config.json`; new UI text honors the selection without a restart, or the restart requirement is documented.
- English remains the baseline, and Simplified Chinese is shipped for the high-traffic screens.
- A maintainable check ensures newly added user-facing strings are registered in the translation layer instead of silently staying English-only.
- Existing zh-CN docs stay consistent with the UI terminology.

### Environment

- Platform: all (macOS / Windows / Linux)
- Version: current `main`

Recommended v1 contract: extract string keys plus en/zh-Hans, then broaden locales later rather than building plugin-loading translation infrastructure up front.

shihuaidexianyu:
If this is a step the project plans to take, I can open a PR to implement it. The suggested v1 scope is string extraction plus en/zh-Hans, then adding more locales later.

l0ng-ai:
[...]

PR #303, opened 2026-08-02
feat(ui): GUI localization for en and zh-Hans

shihuaidexianyu:
Refs #302.

This adds GUI localization for English and Simplified Chinese while keeping CLI output stable in English.

## Scope

- Add a key-based GUI translation layer in `src/ui/i18n.rs`.
- Add an explicit `gui_language` setting with `en` and `zh-CN` as the only supported values.
- Default the GUI to English and fall back to English for unsupported values; there is no automatic system-locale inference.
- Refresh open windows, menus, settings controls, and cached UI state when the language changes.
- Localize the current GUI surfaces, including the Windows Explorer context-menu settings merged from `main`.
- Leave CLI and protocol output in English for scripts and agent integrations.

## Merge compatibility fix

The latest `main` added Windows Explorer settings with raw string search entries after this branch had converted the search index to `L10nKey`. The Explorer title and keywords now use localized keys, and the complete Explorer settings UI and operation feedback are translated. Operation results are stored as semantic state so changing languages cannot leave stale translated feedback behind.

## Verification

- `cargo fmt --check`
- `cargo build --locked`
- `cargo test -p tty7 --bin tty7-app --locked -- --format=terse` — 801 passed, 3 ignored
- Explorer-focused tests — 6 passed


shihuaidexianyu:
Hold on, I still have more commits to push. Please don’t merge yet.

shihuaidexianyu:
All done!

sailtonight (commented):
​

l0ng-ai (commented):
- **Medium — `src/ui/i18n.rs:973`:** Auto locale detection reads only POSIX environment variables, so native Windows/macOS GUI launches can miss the system locale. It also classifies `zh_TW`, `zh_HK`, and `zh_Hant` as Simplified Chinese. Please use cross-platform locale APIs and restrict Simplified Chinese classification to `zh-Hans`/equivalent locales.
- **Medium — `src/ui/app.rs:3720`, `src/ui/app.rs:3732`:** Changing the process-global locale refreshes only the active window, leaving other windows and locale-dependent cached controls stale. Please notify all registered windows and rebuild locale-dependent cached state.

shihuaidexianyu:
I reviewed the follow-up commit `65ee002e`. It addresses both review points: it uses a cross-platform locale API and avoids mapping `zh-TW`, `zh-HK`, `zh-MO`, and `zh-Hant` to Simplified Chinese, and it refreshes locale-dependent state across all registered windows.

That said, I think we can simplify this further by removing automatic language inference altogether. The explicit language setting already gives users control, while auto-detection adds a platform dependency and requires us to maintain increasingly subtle locale-classification rules (for example, a bare `zh` identifier is still ambiguous).

To reduce maintenance complexity and make startup behavior deterministic, I propose that we default `gui_language` to English, keep explicit `en` and `zh-CN` options in settings/configuration, and remove the `auto` option, `sys-locale` dependency, and system-locale detection code.

l0ng-ai:
That makes sense to me. Removing automatic inference avoids the ambiguous bare zh case and keeps the behavior deterministic without adding a platform dependency. I'll follow up by defaulting gui_language to en, keeping only the explicit en and zh-CN choices, and removing the uto option, sys-locale, and the detection/classification code. The multi-window and cached-control refresh fix will remain.

shihuaidexianyu:
> That makes sense to me. Removing automatic inference avoids the ambiguous bare zh case and keeps the behavior deterministic without adding a platform dependency. I'll follow up by defaulting gui_language to en, keeping only the explicit en and zh-CN choices, and removing the �uto option, sys-locale, and the detection/classification code. The multi-window and cached-control refresh fix will remain.我觉得这样很合理。取消自动推断功能可以避免“zh”这种模棱两可的情况，同时确保行为的确定性，而无需依赖特定的平台设置。我会将 gui_language 的默认值设置为“en”，只保留“en”和“zh-CN”这两种选项，同时删除“Auto”选项、sys-locale 相关设置以及用于检测/分类的代码。多窗口功能和缓存控制的刷新功能则保持不变。

I am working on it, please wait for a moment.

shihuaidexianyu:
Fixed the post-merge CI failure in `6358ddb`.

The Windows Explorer search entry added on `main` still used raw strings, while this branch now requires `L10nKey`, which caused the `expected L10nKey, found &str` errors on Linux, macOS, and Windows.

The fix:

- moves the Explorer search title and keywords to localized keys;
- localizes the full Explorer settings section, statuses, buttons, and action feedback;
- stores action feedback as semantic state so it is retranslated after a language change;
- adds regression tests for the localized search entry, Chinese coverage, and error placeholder substitution.

Local verification passes: `cargo build --locked`, all 804 GUI tests (801 passed, 3 ignored), and all 6 Explorer-focused tests.

PR #304, opened 2026-08-02
fix(input): forward macOS editing shortcuts to foreground TUIs

### LINKED ISSUE #301: macOS: Cmd+Backspace is swallowed in foreground TUI apps
### Type

### What's going on?

On macOS, Cmd+Delete (the keyboard's Delete key, reported by GPUI as `platform+backspace`) is swallowed when a foreground TUI application is running.

Steps to reproduce:

1. Open tty7 on macOS.
2. Run Pi, Claude Code, or OpenCode.
3. Type text into the application's prompt.
4. Press Cmd+Delete / Cmd+Backspace.

Expected behavior:

The input before the cursor is deleted. With the cursor at the end, this clears the current input, matching the behavior in other macOS terminal emulators. Sending Ctrl+U manually produces the expected result in these TUIs.

Actual behavior:

Nothing happens. Plain Backspace and Ctrl+U continue to work.

The same behavior reproduces in Pi, Claude Code, and OpenCode, which suggests that the shortcut is being lost in tty7 rather than in one specific TUI.

### tty7 version (bugs)

26.8.1. The relevant code is also unchanged on current `main` (`0a3accd`).

### Platform (bugs)

macOS 26.1, Apple Silicon (`arm64`).

### Anything else?

`TerminalView::on_key_down` routes platform-only chords through `handle_cmd_shortcut`. Its `"backspace"` branch only edits tty7's inline command editor when `input_active()` is true, but returns `CmdKey::Consumed` unconditionally. While a foreground TUI owns the prompt, `input_active()` is false, so the shortcut is consumed without writing anything to the PTY.

The inline tty7 editor behavior should remain unchanged. For the foreground-TUI case, translating Cmd+Backspace to Ctrl+U (`0x15`) would match the observed behavior of other macOS terminals and work across the affected TUIs. A regression test could simulate `platform+backspace` while the inline editor is inactive and assert that the daemon receives `0x15`.

Related context: #184 enabled Kitty keyboard negotiation, and #222 added readline parity for tty7's own inline editor, but neither handles this foreground-TUI Command shortcut.

loscoy:
## Summary

- preserve tty7's existing inline-editor behavior when it owns the shell prompt
- forward macOS foreground-TUI editing chords as readline controls: Cmd+Left -> Ctrl+A, Cmd+Right -> Ctrl+E, Cmd+Backspace -> Ctrl+U, and Cmd+Forward Delete -> Ctrl+K
- send LF for the default Shift+Enter binding when a foreground TUI has not enabled the Kitty keyboard protocol, while preserving Kitty's `CSI 13;2u` encoding when negotiated
- release held input before shortcut bytes and taint typeahead where needed so input ordering and cleared text stay correct
- keep Alt+Enter legacy encoding, custom InsertNewline rebindings, and multi-key Shift+Enter prefixes intact

## Testing

- `cargo +1.97.1 fmt --check`
- `git diff --check`
- `cargo +1.97.1 test --locked` (all test binaries passed; main suite: 919 passed, 3 ignored)
- Rust LSP diagnostics on the changed files (no errors)
- three rounds of `codex review --uncommitted`; the final review found no functional issues
- manual validation on macOS 26.1 / Apple Silicon using an ad-hoc arm64 DMG, covering foreground TUI mappings and tty7 inline-editor regressions

Fixes #301

### DIFF
--- a/src/terminal/view.rs
+++ b/src/terminal/view.rs
@@ -45,7 +45,8 @@ actions!(
         FindNext,
         FindPrevious,
         ClearScrollback,
-        InsertNewline
+        InsertNewline,
+        InsertNewlineFallback
     ]
 );
 
@@ -1184,6 +1185,17 @@ impl TerminalView {
         }
     }
 
+    fn send_shortcut_bytes(&mut self, bytes: &[u8], key: &str, cx: &mut Context<Self>) {
+        let shell_owns_prompt = self.shell_owns_prompt();
+        self.release_hold();
+        self.send_to_pty(bytes, cx);
+        if !shell_owns_prompt {
+            let alt = self.on_alt_screen();
+            self.typeahead
+                .observe(RawInput::Key { key, plain: false }, alt);
+        }
+    }
+
     fn handle_cmd_shortcut(
         &mut self,
         ks: &gpui::Keystroke,
@@ -1222,6 +1234,8 @@ impl TerminalView {
                 if self.input_active() {
                     self.editor_move_edge(false, m.shift);
                     cx.notify();
+                } else if cfg!(target_os = "macos") && self.accepts_input(cx) {
+                    self.send_shortcut_bytes(&[0x01], "a", cx);
                 }
                 CmdKey::Consumed
             }
@@ -1229,6 +1243,8 @@ impl TerminalView {
                 if self.input_active() {
                     self.editor_move_edge(true, m.shift);
                     cx.notify();
+                } else if cfg!(target_os = "macos") && self.accepts_input(cx) {
+                    self.send_shortcut_bytes(&[0x05], "e", cx);
                 }
                 CmdKey::Consumed
             }
@@ -1241,19 +1257,7 @@ impl TerminalView {
                     self.cursor_visible = true;
                     cx.notify();
                 } else if cfg!(target_os = "macos") && self.accepts_input(cx) {
-                    let shell_owns_prompt = self.shell_owns_prompt();
-                    self.release_hold();
-                    self.send_to_pty(&[0x15], cx);
-                    if !shell_owns_prompt {
-                        let alt = self.on_alt_screen();
-                        self.typeahead.observe(
-                            RawInput::Key {
-                                key: "u",
-                                plain: false,
-                            },
-                            alt,
-                        );
-                    }
+                    self.send_shortcut_bytes(&[0x15], "u", cx);
                 }
                 CmdKey::Consumed
             }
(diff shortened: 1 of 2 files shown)


tungloong:
Thanks for this — the Cmd+Backspace → Ctrl+U path matches what I hit in Claude Code / Codex / cursor-agent.

The same swallow also happens for **Cmd+Left** and **Cmd+Right** in `handle_cmd_shortcut`: when `input_active()` is false they still return `CmdKey::Consumed` and never write to the PTY. On a normal shell prompt the inline editor handles them; inside a foreground TUI they do nothing.

Would it make sense to extend this PR (or a follow-up) with the same Ghostty/iTerm-style legacy mapping for the TUI case?

| Chord | When `!input_active()` |
|---|---|
| Cmd+Left | Ctrl+A (`0x01`) |
| Cmd+Right | Ctrl+E (`0x05`) |
| Cmd+Backspace | Ctrl+U (`0x15`) — already in this PR |
| Cmd+Delete (optional) | Ctrl+K (`0x0b`) |

Inline-editor behavior when `input_active()` should stay as-is. Happy to open a separate issue if you'd rather keep #301 scoped to Backspace only.

loscoy:
Confirmed: `left`, `right`, and the physical forward-`delete` branch have the same consume-without-forwarding shape when tty7’s inline editor is inactive.

On balance, I’ll extend #304 to cover the same macOS foreground-TUI compatibility path for Cmd+Left → Ctrl+A, Cmd+Right → Ctrl+E, and physical forward Delete → Ctrl+K, while preserving the existing inline-editor behavior.

I also found that Shift+Enter does not insert a newline in a foreground TUI. It is currently registered as tty7’s `InsertNewline` action for the inline editor, so I’ll include its foreground-TUI propagation/encoding path and regression coverage in this PR as well.

loscoy:
@tungloong The extended mappings are now pushed in `9aaf3b1`:

- Cmd+Left -> Ctrl+A (`0x01`)
- Cmd+Right -> Ctrl+E (`0x05`)
- Cmd+Backspace -> Ctrl+U (`0x15`)
- Cmd+Forward Delete -> Ctrl+K (`0x0b`)

The existing inline-editor behavior remains local. The PR now also handles the default Shift+Enter binding in foreground TUIs: legacy mode sends LF (`0x0a`), while negotiated Kitty mode keeps `CSI 13;2u`. Custom newline rebindings, multi-key Shift+Enter prefixes, and Alt+Enter legacy encoding have regression coverage.

The full Rust test suite passed, and the resulting arm64 DMG was manually validated on macOS 26.1.

sailtonight (commented):
​

PR #308, opened 2026-08-02
feat: add CLI support for opening directories in new tabs

ArnoChenFx:
- support `tty7 [PATH]` to open a directory in a new tab
- activate the most recent local tty7 window through the daemon
- launch `tty7-app` with the requested directory when the GUI is not running

### DIFF
--- a/crates/tty7-cli/src/cli.rs
+++ b/crates/tty7-cli/src/cli.rs
@@ -38,7 +38,7 @@ pub struct Cli {
 
     #[arg(
         value_name = "PATH",
-        help = "Launch or activate the GUI, at PATH if given (not wired up yet)"
+        help = "Launch or activate the GUI, opening a new tab at PATH if given"
     )]
     pub path: Option<String>,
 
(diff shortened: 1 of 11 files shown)


sailtonight (commented):
​

l0ng-ai (commented):
Medium — `src/ui/windows.rs:208`: when the GUI process is alive but has no open windows, a pathless `tty7` request is reported as delivered yet `open_from_cli` returns without creating or activating a window. Please open a default or restored window in this case and add coverage.`n`nLow — `crates/tty7-cli/src/commands.rs:156` and `src/main.rs:156`: `to_string_lossy` changes valid non-UTF-8 Unix paths before they cross the protocol. Please preserve native path bytes in the protocol, or explicitly reject unsupported paths instead of silently changing them.

ArnoChenFx:
> Medium — `src/ui/windows.rs:208`: when the GUI process is alive but has no open windows, a pathless `tty7` request is reported as delivered yet `open_from_cli` returns without creating or activating a window. Please open a default or restored window in this case and add coverage.`n`nLow — `crates/tty7-cli/src/commands.rs:156` and `src/main.rs:156`: `to_string_lossy` changes valid non-UTF-8 Unix paths before they cross the protocol. Please preserve native path bytes in the protocol, or explicitly reject unsupported paths instead of silently changing them.

Fixed

Issue #306, opened 2026-08-02
Add verified in-app updates for packaged installs


ayamir:
## Type

Idea — suggest an improvement or new capability

## What's going on?

tty7 can detect a newer GitHub release, but the current action sends the user to the Releases page to download and install it manually. Packaged desktop installs should be able to download, verify, install, and restart from inside the app.

### Proposed behavior

- Keep the existing launch check and **Settings → About → Check Now** entry point.
- When a newer stable release exists, select the exact asset for the running installation and offer **Update and Restart**.
- Show explicit states for checking, downloading/verifying, installing, up to date, and failure.
- If the installation shape is not safe to replace, keep the current Releases-page fallback rather than guessing.

Possible first-party package mapping:

- macOS app bundle → a signed/notarized `tty7-<version>-macos-<arch>.zip` containing `tty7.app` (the DMG remains for first-time installs)
- Windows Inno install → `tty7-<version>-windows-x86_64-setup.exe`
- Linux AppImage → the matching `.AppImage`
- source builds, package-manager installs, portable Windows ZIPs, read-only locations, and unsupported architectures → do not self-update

### Integrity and replacement requirements

The updater should treat release metadata and downloaded bytes as untrusted:

1. Download the release's existing `checksums.txt` and require an exact filename entry.
2. Verify SHA-256 before any replacement.
3. Verify that the package reports the expected version.
4. On macOS, also verify the bundle signature and require the downloaded app to have the same designated code-signing requirement as the running app.
5. Replace atomically where possible, retain the previous app/binary until the new one launches, and roll back on failure.
6. Never silently fall back to installing an asset for another architecture or installation type.

### Daemon/session lifecycle is the main product decision

The app and local server ship from the same build. Replacing/restarting only the GUI can leave an old server running, and replacing the daemon endpoint incorrectly can strand live sessions. This already caused the failure documented in #42.

The update UX therefore needs an explicit policy. The simplest safe policy is:

- warn that updating restarts tty7's local server;
- save tabs/layout;
- stop the old local server before replacing the app, which ends running local shells;
- start the new app/server and reopen saved tabs with fresh shells;
- reconnect remote workspaces to their existing remote servers.

If ending local shells is not acceptable for an updater, the alternative needs a real cross-version daemon/session migration or a deferred-install-on-quit design. The updater should not silently strand or replace sessions.

### Release/bootstrap note

[...]

Issue #42, opened 2026-07-10
daemon upgrade/replace silently strands old daemon's sessions with live processes


l0ng-ai:
## What happened

After the app auto-updated to v0.8.0, a new daemon was spawned and took over `~/.config/tty7/daemon.sock`. The old daemon (started 6 days earlier) kept running with **14 pane shells still alive — 11 of them hosting live `claude` sessions and one an active `ssh` connection — but with zero client connections and no way for the GUI to ever reach them again**, since it only held an fd to the replaced socket inode.

Nothing surfaced this to the user: no warning, no orphan indicator. The stranded panes sat invisible for days (0% CPU, but holding memory, ptys, and user state).

Observed state:

- old daemon `tty7 --daemon` (pid 4786, started Jul 4) — listener fd pointed at a socket inode no longer on the filesystem; `lsof` showed no peers
- new daemon (pid 66107, started Jul 10 09:36) owned `daemon.sock`; all GUI socket peers matched it
- `update.json` mtime 10:31 and the 10 restored panes under the new daemon all started 10:31:41, so the swap lines up with the auto-update

## Expected

One (or more) of:

1. On startup, the new daemon (or GUI) detects a previous daemon still alive and **adopts / migrates its sessions**, or at least re-exposes them as attachable windows.
2. If migration isn't possible across versions, the old daemon should be told to **gracefully shut down its sessions** (HUP the panes) instead of lingering unreachable.
3. At minimum, **surface it**: warn that N sessions from a previous daemon are stranded, with a way to list/kill them.

## Cleanup note

`kill <old-daemon-pid>` does cascade correctly (ptys close → shells get SIGHUP → children exit), so a built-in "reap stranded daemon" action should be cheap to add.

PR #309, opened 2026-08-02
feat(update): install verified macOS releases in app

ayamir:
## Summary

- install verified stable releases from **Settings → About** without opening a browser
- add a focused, feature-gated `tty7-updater` helper for checksum/signature validation, atomic replacement, relaunch, and rollback
- publish a signed/notarized macOS `.app.zip` alongside the existing DMG

## Scope: macOS only

This first implementation intentionally supports **packaged macOS app bundles only**.

- macOS `.app`: offers **Update and Relaunch** when the exact architecture-specific ZIP and `checksums.txt` are present
- Windows and Linux: keep the existing Releases-page fallback
- source builds, non-writable bundles, unsupported architectures, and packages without the helper: keep the Releases-page fallback
- Nightly packaging remains unchanged; the updater consumes stable releases only

The narrower scope follows the direction discussed in #306 and keeps the first end-to-end release path small enough to validate before adding other platform adapters.

## Daemon and shell behavior

The updater replaces and relaunches only the GUI bundle. It does **not** stop the local server.

On relaunch, tty7's existing protocol handshake decides what happens:

- matching `PROTOCOL_VERSION`: reuse the running server and preserve shells
- incompatible/legacy server: keep it running and show the existing **Keep Shells / Restart** prompt

This avoids silently stranding or terminating sessions, including the failure mode described in #42.

## Integrity and rollback

`tty7-updater`:

1. verifies the downloaded asset against the release's exact `checksums.txt` entry
2. extracts the ZIP and verifies the expected bundle version
3. verifies the bundle code signature and requires the same designated signing requirement as the current app
4. re-verifies after the GUI exits
5. replaces the bundle with same-filesystem renames
6. restores and relaunches the previous app if the new app exits during the launch grace period

## Verification

- `cargo test -p tty7 --bin tty7-app` — 914 passed, 3 ignored
- `cargo test -p tty7 --features updater --bin tty7-updater` — 4 passed
- `cargo test -p tty7-core daemon::install::checksums::tests` — 9 passed
- workflow YAML parsing, `bash -n .github/scripts/bundle-macos.sh`, and `git diff --check`
- local stable bundle smoke test: ZIP contains `tty7-app`, CLI, and `tty7-updater`; version and deep code signature verify
- local Nightly-path smoke test: no helper and no updater ZIP are produced

The local machine still hits its existing `hdiutil: create failed - device not configured` issue when creating the DMG; the app bundle and updater ZIP are produced and verified before that step.

Related to #306.

### DIFF
--- a/.github/scripts/bundle-macos.sh
+++ b/.github/scripts/bundle-macos.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 # Usage: bundle-macos.sh <target-triple> <arch-label>
-# Package the release binary into dist/tty7.app and wrap it in a
-# drag-to-Applications DMG: dist/tty7-<version>-macos-<arch>.dmg.
+# Package the release binary into dist/tty7.app, then publish both:
+#   dist/tty7-<version>-macos-<arch>.zip  (in-app updater)
+#   dist/tty7-<version>-macos-<arch>.dmg  (drag-to-Applications install)
 #
 # Signing posture is chosen from the environment:
 #   * Developer ID secrets present (APPLE_SIGNING_IDENTITY + APPLE_CERTIFICATE)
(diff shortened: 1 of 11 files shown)


sailtonight (commented):
​

l0ng-ai (approved):
Reviewed the updater flow, release packaging, rollback safety, and the follow-up fix. Required CI is green.

PR #323, opened 2026-08-03
fix(bundle): declare macOS TCC privacy keys for child processes

Gabyran:
## Summary

Declare the macOS TCC (privacy) usage strings in the macOS bundle so that programs run inside tty7 panes can request protected resources with a normal one-time prompt instead of being **denied outright with no prompt**.

## Problem

tty7's panes are forked from the bundled executable, so macOS attributes a child process's protected-resource request to `tty7.app` as the responsible process. `tty7.app` currently declares **no usage strings at all**, which means a program you run in a pane that asks for the camera, microphone, contacts, calendars, photos, location, reminders, or Apple Events is **denied outright** — no prompt ever appears, and the user cannot even grant it from System Settings. This is the same failure as anthropics/claude-code#82543, where the reporter notes the identical command works from a terminal that declares the key. Both kitty (`net.kovidgoyal.kitty`) and Kaku (`fun.tw93.kaku`) ship this key set for exactly this reason.

## Fix

- Add the full usage-string set to the generated `Info.plist`, kitty-style: "A program running inside tty7 would like to access …" — Camera, Microphone, Contacts, Calendars, Reminders, PhotoLibrary, Location, Motion, LocalNetwork, Bluetooth, SpeechRecognition, SystemAdministration, and AppleEvents.
- Use the current macOS spellings: `NSCalendarsFullAccessUsageDescription`, `NSRemindersFullAccessUsageDescription`, and `NSLocationUsageDescription` (not the iOS `NSLocationWhenInUseUsageDescription`).
- Add **no** entitlement to match. The entitlements set stays at the three gpui's Metal path needs. A hardened-runtime entitlement is checked against the process actually sending the request — the pane's child, carrying its own signature — and entitlements are per-executable and never inherited, so camera / microphone / location / apple-events on `tty7.app` would do nothing for a pane while widening what injected code could reach under an identity that already holds `disable-library-validation`. The declaration side is the half that gates the child.
- Document in `docs/features.md` (and the Chinese translation) that Full Disk Access — `~/Library/Mail`, `~/Library/Messages`, `~/Library/Safari`, `~/Library/Containers` — still needs a manual System Settings grant, since Apple defines no `NS*UsageDescription` key for that class.

## Verification

- Both generated plists lint clean under `plutil` on macOS with `${VERSION}` substituted; `bash -n` clean on the script; trailing newline at EOF intact.
- Key set matches kitty's `macos_info_plist()` one for one, minus the dict-shaped iOS-only `NSLocationTemporaryUsageDescriptionDictionary`.
- Attribution confirmed against the real topology: `daemon::spawn::spawn_detached` launches `current_exe --daemon` and the only bundled caller is the GUI, so the process that forks every pane is `tty7-app` itself — same signature, same bundle, same `Info.plist`.

`NSMotionUsageDescription` is kept for kitty parity but is inert: Core Motion has no macOS implementation. It is deliberately left out of the user-facing docs list.


l0ng-ai:
Thanks for digging into this — there **is** a real gap here, but I don't think this diff closes it, and I'd like the problem statement corrected before we land anything.

### The symptom in the description isn't TCC-usage-string territory

The repeated *"tty7 would like to access data from other apps"* prompts for `~/Library/Containers`, `~/Library/Mail`, `~/Library/Messages` and `~/Library/Safari` are the **app-data / Full Disk Access** class of TCC. There is no `NS*UsageDescription` key for those services — Apple doesn't define one. The only remedy is the user granting tty7 Full Disk Access in System Settings, and nothing in this diff affects it. Reports of that prompt recurring generally trace back to per-process attribution churn, not to a missing plist key.

### The real gap this *should* be fixing

Panes are forked from the bundled executable, so TCC attributes a child's request to `tty7.app` as the responsible process. `tty7.app` currently declares no usage strings at all, which means a program you run in a pane that asks for Camera / Microphone / Contacts / Calendars / Photos / Reminders / Location / Apple Events is **denied outright, with no prompt** — the user can't even grant it from System Settings. That's the same failure as anthropics/claude-code#82543, where the reporter notes the identical command works from a terminal that does declare the key. Both kitty and Kaku ship this key set for exactly this reason.

### Why the current diff doesn't close it

- It adds the `com.apple.security.device.camera`, `device.audio-input` and `personal-information.location` **entitlements** but none of `NSCameraUsageDescription`, `NSMicrophoneUsageDescription`, `NSLocation*UsageDescription`. Entitlements are per-executable and are not inherited by children; the usage string on the responsible bundle is what TCC actually checks. So those three resources stay exactly as broken as they are today, while the entitlement half — the half that doesn't help here — is granted.
- The five folder/volume keys (`NSDesktopFolderUsageDescription`, `NSRemovableVolumesUsageDescription`, `NSNetworkVolumesUsageDescription`, …) are the *optional* class: Apple documents these as changing the wording of a prompt the user gets either way. They're cosmetic, and neither kitty nor Kaku declares them. Meanwhile every key those two do ship — Camera, Microphone, Reminders, Location, PhotoLibrary, SpeechRecognition, LocalNetwork, Bluetooth, SystemAdministration — is missing here.
- `NSAppleEventsUsageDescription` paired with `com.apple.security.automation.apple-events` is the one part that's exactly right. Keep that.

### Security posture

We already sign with `disable-library-validation` and `allow-unsigned-executable-memory`. Handing that same binary camera, microphone, address-book and location entitlements widens what injected code can reach under tty7's identity, and Apple's guidance is to use only the entitlements that are strictly necessary. `bundle-macos.sh` already reasons this way in its own comments when it declines to give the CLI the GUI's entitlements — I'd like this change to engage with that rather than sidestep it.

### What I'd like to see

1. Corrected problem statement — drop the Full Disk Access framing, describe the child-process-denied-without-prompt failure.
2. The full usage-string set, at minimum `NSCameraUsageDescription` / `NSMicrophoneUsageDescription` / `NSLocation*UsageDescription` so the entitlements already in this diff aren't dead weight.
3. Either drop the entitlements that have no matching usage string, or justify each one.
4. Consider phrasing the strings kitty-style — "A program running inside tty7 would like to access …" — since the requester really is the child process, not tty7 itself.
5. Restore the trailing newline at EOF in `bundle-macos.sh`.
6. A short docs/README note that `~/Library/Mail|Messages|Safari|Containers` requires the user to grant Full Disk Access manually, since that's the question this PR started from.

For transparency on how this was checked: the bundle config, entitlements and signing setup were read on `main`, the plist and entitlements heredocs were extracted and parsed, and the key requirements were taken from Apple's current documentation (`required` vs `optional, but highly recommended` wording) plus the kitty and Kaku bundles as prior art. No macOS machine was available, so the TCC behaviour itself was not observed live — if you can reproduce a child-process denial before and after, that would settle it.

Holding off on merging until the above is sorted. Happy to keep iterating.

l0ng-ai:
Two corrections to my own review above, from a second pass over Apple's current key list — both affect what the final key set should be, so worth having before you rework this:

1. **`NSCalendarsUsageDescription` is deprecated.** Apple's docs: *"has been deprecated. If your app needs read and write access to a person's calendar data, use `NSCalendarsFullAccessUsageDescription` instead."* Same story for `NSRemindersUsageDescription` → `NSRemindersFullAccessUsageDescription`. Our runners are `macos-14`/`macos-15-intel` and we set no `LSMinimumSystemVersion`, so the full-access keys are the right target for the shipping audience; keep the old spellings only if you want a pre-14 fallback.

2. **On macOS the location key is `NSLocationUsageDescription`** — I wrote `NSLocation*UsageDescription` above, which is ambiguous. It is specifically *not* `NSLocationWhenInUseUsageDescription`, which is the iOS spelling and would be silently ignored in a Mac bundle. Worth pinning down since a typo'd key fails exactly the same way as a missing one: no prompt, no System Settings entry.

Also, so it doesn't mislead anyone reading the checks: **the green CI on this PR means nothing for this change.** `ci.yml` never invokes `bundle-macos.sh` — it only runs on release/nightly. I did validate both heredocs out of band (with `${VERSION}` substituted, the plist and entitlements XML parse cleanly via `plistlib`, no illegal `--` inside the XML comments, no shell-significant characters in the unquoted `<<PLIST` heredoc), so the syntax is fine; it's the semantics above that need work.

One thing I can confirm in your favour, since it's the premise the whole PR rests on: the child-process attribution story is sound, and `bundle-macos.sh` is the only Info.plist/entitlements generator in the repo, so there's no second place this needs applying. TCC attributes a pane child's request to the responsible process, reads the usage string from *its* bundle, and records the grant against its bundle ID — and since entitlements are per-executable and not inherited, the declaration side really is the part that matters. No extra hardened-runtime entitlement is needed for the child case.

Still holding for the rework in my previous comment, with the two key-name corrections folded in.

Gabyran:
Thanks for the detailed review — reworked the PR to address all six points plus the two key-name corrections.

**1. Problem statement corrected** — dropped the Full Disk Access framing. The PR body now describes the real failure: panes are forked from the bundled executable, so TCC attributes a child's request to `tty7.app`, and without any usage strings a program run in a pane that asks for camera/mic/contacts/etc. is **denied outright with no prompt** (matching anthropics/claude-code#82543).

**2. Full usage-string set added** (kitty/Kaku set + the required extras), all kitty-style: "A program running inside tty7 would like to access …" — Camera, Microphone, Contacts, Calendars, Reminders, PhotoLibrary, Location, Motion, LocalNetwork, Bluetooth, SpeechRecognition, SystemAdministration, AppleEvents.

**3. Key-name corrections folded in** — `NSCalendarsFullAccessUsageDescription`, `NSRemindersFullAccessUsageDescription`, and `NSLocationUsageDescription` (macOS spelling, not the iOS `NSLocationWhenInUseUsageDescription`).

**4. Entitlements trimmed to minimal** — dropped the camera / audio-input / location / addressbook / calendars entitlements that had no matching usage string. Kept only `com.apple.security.automation.apple-events`, which pairs with `NSAppleEventsUsageDescription`. The comment in the entitlements heredoc explains the least-privilege rationale (entitlements are per-executable and never inherited by children; the child's access is gated by the responsible bundle's usage string, and granting the GUI bundle camera/mic/location would only widen what injected code can reach under tty7's identity).

**5. Trailing newline restored** at EOF in `bundle-macos.sh`.

**6. Docs note added** — `docs/features.md` now has a "macOS privacy" section explaining both things usage strings *don't* cover: Full Disk Access (Mail/Messages/Safari/Containers requires a manual System Settings grant, no plist key exists) and the entitlement-gated classes that are deliberately not granted to the bundle.

Both heredocs re-validated: the plist and entitlements parse cleanly via `plistlib` with `${VERSION}` substituted, and the trailing newline is confirmed present.

l0ng-ai:
The reworked version looks right — all six points plus the two key-name corrections are in. This time I had a Mac in front of me, so the checks below are observed rather than inferred.

### Verified

- Both heredocs extracted and linted with the real `plutil` (not just `plistlib`), `${VERSION}` substituted: `Info.plist: OK`, `ent.plist: OK`. 13 `NS*UsageDescription` keys land in the plist, the XML comments inside `<dict>` parse fine, and nothing in the added text is shell-significant inside the unquoted `<<PLIST` heredoc.
- Key names are the current macOS spellings — `NSCalendarsFullAccessUsageDescription`, `NSRemindersFullAccessUsageDescription`, `NSLocationUsageDescription`. No iOS `WhenInUse` variant crept in.
- kitty parity checked against `macos_info_plist()` in its `setup.py`: your set matches its keys one for one. The only kitty key you don't ship is `NSLocationTemporaryUsageDescriptionDictionary`, which is a dict-shaped iOS thing — fine to omit.
- Trailing newline at EOF restored (the file ends `…$DMG"\n`).
- The premise holds in our actual topology, which is the part I couldn't check last time. `daemon::spawn::spawn_detached` launches `current_exe --daemon`, and on macOS the only caller in the bundle is the GUI (`tty7-server` calls it too but isn't bundled). So the process that forks every pane *is* `tty7-app` — same code signature, same bundle, same `Info.plist`. The usage strings reach the right process. `Contents/MacOS/tty7` resolves its main bundle to `tty7.app` as well, so even a CLI-rooted process would read the same strings.

### One thing I'd still change

The entitlements comment says Apple Events is "the one resource where the entitlement is genuinely exercised". I can't find anything that exercises it: no `NSAppleScript`, no `osascript`, no Apple-event send anywhere in `src/`, `crates/`, or gpui's `platform/mac`.

And for the case this PR is about it wouldn't help even if we did. The hardened-runtime automation entitlement is checked on the **sending** process, which here is the child — `/usr/bin/osascript`, Apple-signed, its own identity. What TCC reads off `tty7.app` is `NSAppleEventsUsageDescription`, and that is already in `Info.plist`. Entitlements aren't inherited, which is the same argument the diff uses to drop camera / mic / location; it applies here unchanged.

Worth noting alongside it: the entitlements heredoc only runs inside the `if [[ -n "$SIGN_ID" && -n "${APPLE_CERTIFICATE:-}" ]]` branch, so adhoc and local builds never get it at all, while `Info.plist` applies to every build. That asymmetry is another sign the entitlement isn't the load-bearing half.

So either drop `com.apple.security.automation.apple-events` and keep the usage string, or keep it and reword the comment to say plainly that it's speculative cover in case tty7 itself ever sends an Apple event. I'd take the first — it's what the rest of the diff argues for, and it keeps the entitlement set at exactly what gpui's Metal path needs.

### Nits

1. `~/Library/Mail/Messages/Safari/Containers` reads as a single nested path. Make it `~/Library/{Mail,Messages,Safari,Containers}` in both docs.
2. Section placement diverges between the two files: the English doc puts **macOS privacy** before **Localization**, the Chinese one appends **macOS 隐私** after **本地化**. They're parallel translations — keep the order the same.
3. `NSMotionUsageDescription` is Core Motion, which doesn't exist on macOS, so the key is inert. kitty ships it too, so I'm not asking you to remove it — just flagging that it's parity decoration rather than a gap being closed, in case anyone reads the docs list as thirteen things that now work.

Sort out the entitlement point (either way) and the two docs nits and I'll merge.

l0ng-ai:
Pushed the three items myself rather than bounce them back — they were small and this had waited long enough. `cf788de` on your branch, plus a merge of current `main` so the strict branch check is satisfied.

**Entitlement** — dropped `com.apple.security.automation.apple-events`; the entitlements set is back to exactly the three gpui's Metal path needs. `NSAppleEventsUsageDescription` stays in `Info.plist`, which is the half that actually gates the child. Rewrote the comment to explain why there is deliberately *no* entitlement matching the usage strings, instead of justifying one.

**Docs** — the four Full Disk Access paths are spelled out separately now (`~/Library/Mail`, `~/Library/Messages`, `~/Library/Safari`, `~/Library/Containers`) rather than run together as one nested path; the English and Chinese sections sit in the same place (after Performance notes, before Localization); and I reworked the lead paragraph to state the forked-pane attribution up front, since that is the reason any of this is needed. Dropped motion from the user-facing list — the key stays for kitty parity, but Core Motion has no macOS implementation and listing it reads as a capability that now works. Kept a closing line making clear that declaring a usage string is not the same as holding the permission.

Re-verified after the edits: `plutil -lint` clean on both generated plists, `bash -n` clean on the script, trailing newline intact, 13 usage strings still present.

Thanks for sticking with this through two rounds of review — the analysis in your rework was right, and the diff is in good shape. Merging once CI is green.

Issue #331, opened 2026-08-03
OSC 11 background color query gets no response in daemon-backed panes


yhzhu99:
### Type

Bug — terminal background-color query is not answered

### What's going on?

On Windows, tty7 does not return a reply to the standard OSC 11 background-color query (`ESC ] 11 ; ? BEL`). Applications that select a light/dark theme from the terminal background therefore time out and fall back to a dark theme, even when tty7 is configured with a light theme.

This is visible with Pi `v0.83.0`: setting `"theme": "light/dark"` under tty7 with `catppuccin_latte` still selects Pi's dark theme, making yellow warnings render as `RGB(255,255,0)` on a light background. Forcing Pi to `"theme": "light"` works around it.

### Minimal reproduction

Save as `osc11-probe.js` and run `node osc11-probe.js` inside a tty7 pane:

```js
if (!process.stdin.isTTY || !process.stdout.isTTY) process.exit(2);
process.stdin.setRawMode(true);
process.stdin.resume();

let bytes = Buffer.alloc(0);
process.stdin.on("data", chunk => {
  bytes = Buffer.concat([bytes, chunk]);
});

process.stdout.write("\x1b]11;?\x07");
setTimeout(() => {
  process.stdin.setRawMode(false);
  process.stdin.pause();
  console.log(`OSC11_HEX=${bytes.toString("hex")}`);
  process.exit(0);
}, 1500);
```

Actual output:

```text
OSC11_HEX=
```

Expected: an OSC 11 response containing tty7's active terminal background, for example `ESC ] 11 ; rgb:eeee/ffff/ffff BEL` (exact RGB depends on the active preset).

### Notes

The source already has an apparent response path in `src/terminal/view.rs`:

```rust
AlacEvent::ColorRequest(idx, fmt) => {
    // idx 257 -> theme.background
    self.terminal.write(fmt(rgb).into_bytes());
}
```

So this may be a regression in the daemon-backed output/event path rather than missing OSC parsing. The query reaches the pane but no bytes return to the child PTY.

Pi first probes color-scheme DSR and then falls back to OSC 11, so implementing either protocol would help; OSC 11 is the broadly compatible path already represented in tty7's code.

### Environment

- Windows 11 Enterprise `10.0.26200`
- Reproduced indirectly on official `26.8.2-nightly.20260802` through Pi's auto-theme detection
- Reproduced directly with the raw probe on source build `07bb3a771450dd29ce57494e0ec0efca7c1320f0`
[...]

PR #346, opened 2026-08-04
fix(windows): sync ConPTY colors with active theme

### LINKED ISSUE #345: Windows: ConPTY color table never matches the active theme — OSC 10/11 queries are swallowed, palette-aware apps render wrong light/dark UI
### User-visible bug

With a **light** theme (e.g. `catppuccin_latte`), the terminal background is light, but TUI applications that detect the terminal background render a **dark** UI. Concrete example: OpenAI codex CLI renders a near-black input box with dark text on a light background — low contrast, effectively unusable. Switching the terminal theme to dark "fixes" it only because the app happens to be dark.

### Root cause (verified empirically on Windows 11, tty7 26.8.x)

1. **tty7 uses ConPTY on Windows** (`conhost.exe` children of `tty7-app.exe`).
2. **The ConPTY color table is never initialized from tty7's theme.** It stays at the Windows console defaults (`ColorTable[0]=0x0C0C0C` dark background, `ColorTable[7]=0xCCCCCC`) regardless of the active theme. Verified from inside a real tty7 pane with `GetConsoleScreenBufferInfoEx`:
   ```
   wAttributes=0x0007  CT[0]=0x000C0C0C  CT[7]=0x00CCCCCC
   ```
   (theme was light, `catppuccin_latte`)
3. **OSC 10/11 queries are never answered inside ConPTY.** A child process sending `ESC]10;?ESC\ ESC]11;?ESC\` receives no reply (3s+ wait) and the query is not forwarded to the terminal either — conhost consumes it. This matches the existing comment in `crates/tty7-core/src/daemon/pane.rs:336` ("ConPTY consumes OSC 11 queries before tty7's emulator can answer them").
4. Consequently, palette-aware apps have only one source of truth on Windows: the console color table read via `GetConsoleScreenBufferInfoEx` (that is exactly codex's fallback path — it tries OSC 10/11 first, then falls back to the console API). The table reports dark, the app renders dark. The same code works correctly under Windows Terminal because WT keeps its ConPTY color table in sync with its theme.

### Why the current `COLORFGBG` workaround doesn't help

`pane.rs` sets `COLORFGBG=0;15` / `15;0` as a "conventional fallback hint" for TUI apps. This is a legacy vim/emacs convention — modern palette-aware TUIs (codex, Neovim's bg detection, etc.) do not read it, and even the xterm/colorterm docs don't treat it as authoritative. The color table is the interface these apps actually use.

### Fix suggestion

Mirror what Windows Terminal does: when a Windows pane is spawned, initialize the pseudoconsole's color table and default attributes from the active theme palette (the daemon already knows the theme via the cached `Appearance` in `crates/tty7-core/src/core/machine.rs` (`note_appearance`/`appearance()`), written by `src/ui/theme.rs` `apply_theme`).

Notes for implementation:
- The **terminal-side** ConPTY pipe handles reject console APIs (`Get/SetConsoleScreenBufferInfoEx` → `ERROR_INVALID_HANDLE`, verified), and `portable-pty 0.8.1` does not expose a usable handle for this. From **inside** the PTY, `SetConsoleScreenBufferInfoEx` on the child's screen-buffer handle works (verified — the table updates and reads back correctly).
- So the fix likely needs one of:
  - extending `portable-pty` to expose the ConPTY output handle / HPCON so the daemon can call `SetConsoleScreenBufferInfoEx` (WT-compatible approach), or
  - a verified mechanism to push colors from the terminal side (e.g. OSC 4/10/11 SET through the PTY input pipe — needs verification that conhost processes terminal-side color SETs).
- `COLORFGBG` can stay as a legacy hint, but it should not be the primary mechanism.

### Evidence recap

- Probe run inside a real tty7 pane (`tty7 run`): OSC 10/11 → no reply; `GetConsoleScreenBufferInfoEx` → dark table while theme is light; `SetConsoleScreenBufferInfoEx` from the child → succeeds, table updates.
- codex 0.146.0 Windows probe source (`codex-rs/tui/src/terminal_probe.rs`): OSC 10/11 first, `GetConsoleScreenBufferInfoEx` fallback; palette choice in `codex-rs/tui/src/style.rs` driven by `default_bg()`.

### Environment

- tty7 26.8.x (Windows 11 26200, x64)
- Repro app: `codex` CLI 0.146.0 (any palette-aware TUI works, e.g. Neovim `:set background` detection)

yhzhu99:
## Summary
- persist the active theme's ANSI table and exact default foreground/background alongside the existing appearance hint
- initialize the Win32 console table from inside ConPTY before launching the configured pane command
- preserve the real shell PID across the bootstrap so titles, process snapshots, teardown, and exit monitoring keep their existing semantics
- keep old dark-only appearance caches readable and retain COLORFGBG as a legacy hint

## Why this approach
ConPTY's terminal-side pipe handles reject console screen-buffer APIs, while a process inside the pseudoconsole can update the table successfully. The bootstrap is a private mode of the existing tty7 executable, so it adds no shipped helper binary or portable-pty fork and preserves the original argv, cwd, environment, and stdio.

## Verification
- cargo test -p tty7-core daemon::conpty_bootstrap::tests -- --nocapture (5 passed)
- cargo test -p tty7-core core::machine::tests::the_appearance_hint_round_trips_and_defaults_to_light (passed)
- cargo check -p tty7 --bin tty7-app (passed)
- cargo check -p tty7-server (passed)
- git diff --cached --check (passed before commit)

The ConPTY regression test opens a real pseudoconsole, writes a synthetic palette and default attributes, launches the real child process, and verifies that child reads the exact values back through GetConsoleScreenBufferInfoEx.

A full cargo test -p tty7-core run was also attempted: 738 passed, 5 failed, and 1 was ignored. The failures are environment-dependent existing tests (three remote-link helper lookup cases, one router helper lookup case, and the local WSL prompt-cycle expectation); none exercises the new bootstrap path.

Closes #345

### DIFF
--- a/crates/tty7-core/src/core/machine.rs
+++ b/crates/tty7-core/src/core/machine.rs
@@ -1239,7 +1239,7 @@ pub fn appearance_path() -> io::Result<PathBuf> {
     Ok(data_dir()?.join(APPEARANCE_FILE))
 }
 
-/// The light/dark mode the GUI last applied, cached beside the machine tree.
+/// The appearance the GUI last applied, cached beside the machine tree.
 ///
 /// The daemon needs it when it spawns a pane on Windows, where ConPTY drops an
 /// OSC 11 background query before tty7's emulator can answer it (see
@@ -1256,6 +1256,17 @@ pub fn appearance_path() -> io::Result<PathBuf> {
 pub struct Appearance {
     #[serde(default)]
     pub dark: bool,
+    /// The active theme's ANSI table, encoded as `0xRRGGBB`.
+    ///
+    /// Older appearance files contain only `dark`; keeping this optional lets
+    /// them continue to parse while a freshly-painted GUI supplies the table
+    /// ConPTY needs for Win32 console color queries.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub ansi16: Option<[u32; 16]>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub foreground: Option<u32>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub background: Option<u32>,
 }
 
 /// Read the cached appearance, or the default when there is none to read.
@@ -1275,13 +1286,12 @@ pub fn appearance() -> Appearance {
 
 /// Record the appearance the GUI just applied. A no-op when it has not changed,
 /// so repainting the same theme does not touch the disk.
-pub fn note_appearance(dark: bool) {
+pub fn note_appearance(appearance: Appearance) {
     let Ok(path) = appearance_path() else { return };
-    let next = Appearance { dark };
-    if read_appearance(&path) == next && path.exists() {
+    if read_appearance(&path) == appearance && path.exists() {
         return;
     }
-    if let Err(e) = write_appearance(&path, next) {
+    if let Err(e) = write_appearance(&path, appearance) {
         log::warn!("could not write {}: {e}", path.display());
     }
 }
@@ -1354,20 +1364,38 @@ mod tests {
 
         assert_eq!(
             read_appearance(&path),
-            Appearance { dark: false },
+            Appearance::default(),
             "a hint nobody has written yet reads as the default preset"
         );
 
-        write_appearance(&path, Appearance { dark: true }).unwrap();
-        assert_eq!(read_appearance(&path), Appearance { dark: true });
+        let dark = Appearance {
+            dark: true,
+            ansi16: Some(std::array::from_fn(|i| i as u32 * 0x010101)),
+            foreground: Some(0xfafafa),
+            background: Some(0x101010),
+        };
+        write_appearance(&path, dark).unwrap();
+        assert_eq!(read_appearance(&path), dark);
 
-        write_appearance(&path, Appearance { dark: false }).unwrap();
-        assert_eq!(read_appearance(&path), Appearance { dark: false });
+        write_appearance(&path, Appearance::default()).unwrap();
+        assert_eq!(read_appearance(&path), Appearance::default());
+
+        std::fs::write(&path, r#"{"dark":true}"#).unwrap();
+        assert_eq!(
+            read_appearance(&path),
+            Appearance {
+                dark: true,
+                ansi16: None,
+                foreground: None,
+                background: None,
+            },
+            "the pre-palette cache format remains readable"
+        );
 
         std::fs::write(&path, "not json").unwrap();
         assert_eq!(
             read_appearance(&path),
-            Appearance { dark: false },
+            Appearance::default(),
             "a corrupt hint must not decide the background either"
         );
     }
--- a/crates/tty7-core/src/daemon/conpty_bootstrap.rs
+++ b/crates/tty7-core/src/daemon/conpty_bootstrap.rs
@@ -1,0 +1,341 @@
+//! Initializes Win32 console colors before the real pane command starts.
+//!
+//! ConPTY consumes OSC default-color queries and its host-side pipe handles
+//! are not console screen-buffer handles. The first process *inside* the
+//! pseudoconsole can use the console API, though, so Windows pane commands are
+//! prefixed with this executable's private bootstrap mode. It updates the
+//! palette and then runs the original argv with inherited stdio.
+
+#![cfg(windows)]
+
+use std::ffi::{OsStr, OsString};
+
+use portable_pty::CommandBuilder;
+
+use crate::core::machine::Appearance;
+
+const BOOTSTRAP_ARG: &str = "--tty7-conpty-bootstrap";
+const PALETTE_ENV: &str = "TTY7_CONPTY_PALETTE";
+const PID_CHANNEL_ENV: &str = "TTY7_CONPTY_PID_CHANNEL";
+
+pub(crate) struct BootstrapReceiver(std::net::TcpListener);
+
+impl BootstrapReceiver {
+    pub(crate) fn receive(self, bootstrap_pid: u32) -> Option<u32> {
+        use std::io::Read as _;
+
+        self.0.set_nonblocking(true).ok()?;
+        let deadline = std::time::Instant::now() + std::time::Duration::from_secs(1);
+        let mut stream = loop {
+            match self.0.accept() {
+                Ok((stream, _)) => break stream,
+                Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => {
+                    if std::time::Instant::now() >= deadline {
+                        return None;
+                    }
+                    std::thread::sleep(std::time::Duration::from_millis(5));
+                }
+                Err(_) => return None,
+            }
+        };
+        stream
+            .set_read_timeout(Some(std::time::Duration::from_secs(1)))
+            .ok()?;
+        let mut bytes = [0; 4];
+        stream.read_exact(&mut bytes).ok()?;
+        let shell_pid = u32::from_le_bytes(bytes);
+        crate::daemon::winproc::snapshot()
+            .iter()
+            .any(|p| p.pid == shell_pid && p.parent == bootstrap_pid)
+            .then_some(shell_pid)
+    }
+}
+
+/// Prefix a pane command with the current tty7 executable's private bootstrap.
+pub(crate) fn wrap_command(
+    cmd: &mut CommandBuilder,
+    appearance: Appearance,
+) -> anyhow::Result<Option<BootstrapReceiver>> {
+    let Some(mut colors) = appearance.ansi16 else {
+        return Ok(None);
+    };
+    let (foreground_index, background_index) = if appearance.dark { (15, 0) } else { (0, 15) };
+    if let Some(foreground) = appearance.foreground {
+        colors[foreground_index] = foreground;
+    }
+    if let Some(background) = appearance.background {
+        colors[background_index] = background;
+    }
+    let listener = std::net::TcpListener::bind((std::net::Ipv4Addr::LOCALHOST, 0))?;
+    let address = listener.local_addr()?;
+    let exe = std::env::current_exe()?;
+    cmd.get_argv_mut()
+        .splice(0..0, [exe.into_os_string(), OsString::from(BOOTSTRAP_ARG)]);
+    cmd.env(PALETTE_ENV, encode_palette(appearance.dark, colors));
+    cmd.env(PID_CHANNEL_ENV, address.to_string());
+    Ok(Some(BootstrapReceiver(listener)))
+}
+
+/// Run the private bootstrap mode, returning `None` for every normal launch.
+pub fn run_if_requested() -> Option<i32> {
+    let mut args = std::env::args_os().skip(1);
+    if args.next().as_deref() != Some(OsStr::new(BOOTSTRAP_ARG)) {
+        return None;
+    }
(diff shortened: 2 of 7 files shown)


l0ng-ai:
Thanks for this — the core insight is right, and I want to say that up front because the conclusion below is "hold off". Running the palette call *from inside* the pseudoconsole is the one place where `SetConsoleScreenBufferInfoEx` actually applies, the PID handoff so `shell_pid` semantics survive is thoughtful, and the code is clean and well-commented. The problem is the choice of wrapper binary, and it's the kind of thing that looks fine in every test we have.

### Blocker: in the shipped build the bootstrap is a GUI-subsystem image, which the loader does not attach to a console

`conpty_bootstrap.rs:71` re-execs `std::env::current_exe()`. On Windows the daemon is itself `current_exe() --daemon` (`daemon/spawn.rs`), and the shipped Windows payload is only `tty7-app.exe` + `tty7.exe` (`.github/scripts/bundle-windows.ps1`). `src/main.rs` compiles `tty7-app.exe` as:

```rust
#![cfg_attr(
    all(target_os = "windows", not(debug_assertions)),
    windows_subsystem = "windows"
)]
```

So in release the bootstrap is a GUI-subsystem process — and a GUI-subsystem image is not connected to the console by the loader. Measured on a Windows 11 box: a GUI-subsystem child of a ConPTY-hosted parent gets `ERROR_INVALID_HANDLE` from `CreateFileW("CONOUT$")` and a failing `GetConsoleCP()`, while the **byte-identical console-subsystem build** of the same code succeeds and reads back exactly the defaults from #345 (`ColorTable[0] = 0x0c0c0c`, `ColorTable[7] = 0xcccccc`).

That makes this worse than a no-op. `portable-pty` sets the child's stdio to `INVALID_HANDLE_VALUE` and relies solely on `PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE` for the handoff (`portable-pty-0.8.1/src/win/psuedocon.rs`), so a bootstrap that the loader skips has no console *and* no valid stdio. Best case `set_console_palette` fails and #345 is unfixed in the build the reporter runs; worst case the shell it spawns with `Stdio::inherit()` inherits no ConPTY, allocates a fresh console of its own, and the pane comes up blank.

The reason nothing catches it: `not(debug_assertions)` means `cargo run`, `cargo test` and every CI job build console-subsystem binaries, including the new `console_palette_is_inherited_by_the_real_pane_process` test. The green checks on this PR say nothing about the release configuration. It also explains an asymmetry worth noting — `tty7-server.exe` *is* console-subsystem, so the CLI path would work; the GUI path is the broken one.

Fixing this needs either a console-subsystem helper that's actually staged into the bundle (a new shipped binary, plus installer/zip changes) or a different mechanism — which is why I'm not patching it on top.

### Blocker: no `SetConsoleCtrlHandler`, so the first Ctrl+C breaks pane teardown

The bootstrap is attached to the pane's console, so `CTRL_C_EVENT` is delivered to it, and with no handler installed the default action **terminates** it — while the shell, which handles Ctrl+C itself, keeps running. From that point the pane's `child` handle refers to a dead process: the exit status the daemon eventually reports becomes `0xC000013A` instead of the shell's, and `child.kill()` in `DaemonPane::kill()` no longer reaches anything. There's no `SetConsoleCtrlHandler` anywhere in the diff. It needs a handler returning `TRUE` for `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` — specifically *not* `SetConsoleCtrlHandler(NULL, TRUE)`, whose ignore-flag children inherit.

### Should-fix: nothing terminates the shell on hangup any more

`winproc::descendants` excludes the root, so `kill_descendants(shell_pid)` kills the shell's children but not the shell, and `child.kill()` now only reaches the bootstrap. Teardown ends up depending on ConPTY closing. `crates/tty7-core/Cargo.toml` documents the invariant this breaks ("ConPTY's `kill` only reaches the shell itself"). A `winproc::terminate(pid)` after the descendants loop covers it, and is harmless on the non-bootstrap path.

### Should-fix: the palette is pushed only at spawn

Switching theme, or the OS flipping light/dark, leaves every already-open pane on the old table — so the reported symptom comes back for existing panes. The PR title says "sync ConPTY colors with active theme"; as written it's "seed the palette at pane creation". Worth either handling or stating explicitly.

### Should-fix, smaller

- `conpty_bootstrap.rs:95` and `:120`: `eprintln!` writes into the pane's own ConPTY, so a failure puts an untranslated English line at the top of the user's shell. This repo routes user-visible text through `src/ui/i18n.rs`; a daemon-side log is probably the right answer here instead.
- `conpty_bootstrap.rs:24-51`, called synchronously from pane creation: `receive()` can block the interactive new-pane path for up to ~1s of accept polling plus a 1s read timeout whenever the handoff doesn't happen. Since `receive` already takes a `winproc::snapshot()` to validate parentage, "find the bootstrap's single child in the snapshot" would drop the loopback channel entirely.
- `conpty_bootstrap.rs:178`: overriding `wAttributes` changes what conhost emits for default and erase attributes, while tty7's own renderer still paints SGR 30/107 from the theme's real `ansi16[0]`/`ansi16[15]` — which the console table has just overwritten with fg/bg. Worth one visual check that `cls` and a full-screen TUI in a light theme still repaint with the theme background.

### Nits

- Repo idiom for a Windows-only module is `#[cfg(windows)] pub(crate) mod x;` in `daemon/mod.rs` (see `winproc`), not an unconditional `pub mod` with an inner `#![cfg(windows)]`. Non-Windows is otherwise gated correctly — no cross-platform regression.
- `rgb_to_colorref` sits after the `#[cfg(test)] mod tests` block; it belongs with the other helpers.
- `!= TRUE` on BOOL returns; the rest of the tree compares `!= 0` (see `winproc.rs`).
- `machine.rs`: the new `foreground`/`background` fields have no doc comments while `ansi16` does, in an otherwise densely documented file. `pane.rs` also reads the appearance file twice per spawn now.
- `console_palette_is_inherited_by_the_real_pane_process` opens a real pseudoconsole and recursively re-runs the test binary. Nested ConPTY isn't available everywhere — in a constrained sandbox every pseudoconsole client including `cmd.exe` died with `STATUS_DLL_INIT_FAILED (0xC0000142)` — so consider skipping when `openpty`/spawn fails, or it'll flake for people running the suite in containers/VMs.

### The alternative I'd rather we evaluate first

`PSEUDOCONSOLE_PASSTHROUGH_MODE` is defined but unused in `portable-pty` (we currently pass only `RESIZE_QUIRK | WIN32_INPUT_MODE`). In passthrough mode the OSC 10/11 queries reach tty7's own emulator — the component that actually knows the theme — instead of conhost's palette. That fixes the problem at the layer that owns the answer, gets live theme changes for free rather than as follow-up work, and needs no wrapper process, no PID handoff and no console-subsystem helper in the bundle. It's also a plausible explanation for why these apps behave under Windows Terminal. Before we commit to the re-exec design I'd like to know whether passthrough covers #345 — if it does, most of this PR's scaffolding goes away.

### How this was checked

The diff was read in full against `main`, and the subsystem/bundle/spawn claims were verified in the tree (`src/main.rs`, `bundle-windows.ps1`, `daemon/spawn.rs`) and in vendored `portable-pty` 0.8.1. The GUI-vs-console-subsystem console attachment was measured directly with a pair of otherwise identical probe binaries under a real pseudoconsole, which is where the `CONOUT$`/`GetConsoleCP` and `ColorTable` numbers above come from. The Ctrl+C and teardown consequences are read off the code paths, not observed live — if you can show a release build where a pane survives Ctrl+C with the correct exit status, that would change the picture, and it's the evidence I'd want in the PR either way.

Holding off on merging for now. The passthrough question is the one I'd tackle next, and if you'd rather hand this off than rework it, say so and I'll pick it up.

l0ng-ai:
Following up on my review with measurements, because they change the design rather than the details — and the diagnosis in #345 is what led here, so thank you for it.

I went looking at how other terminals answer this on Windows. Warp has no `SetConsoleScreenBufferInfoEx` anywhere in its tree; it loads its **own** `conpty.dll` (`app/src/terminal/local_tty/windows/conpty_api.rs`, `CreatePseudoConsole` flags `0`), vendors the pair under `app/assets/windows/{x64,arm64}/`, and answers `OSC 10/11` from its emulator. wezterm bundles the same pair. That pointed at the possibility that the in-box `conhost` is the only thing swallowing the query — and that `portable-pty 0.8.1` already prefers a sideloaded `conpty.dll` over `kernel32`'s (`src/win/psuedocon.rs:44-59`).

Measured on Windows 11 26200, one probe binary run twice, the only difference being whether Microsoft's redistributable pair sits beside it:

```
in-box conhost:  terminal side sees <ESC>[?9001h<ESC>[?1004h        -> client: no reply, 2.5s timeout
bundled ConPTY:  terminal side sees ... <ESC>]11;?<BEL>             -> client: the reply, byte-for-byte
```

And in a real tty7 pane under `catppuccin_latte`, same config, same binary:

```
without the pair:  client: NO REPLY within 2.5s
with the pair:     client: BACKGROUND = #eff1f5     <- the preset's exact background
```

So the query is answerable end to end without touching the console color table at all: `terminal::view` already replies to `OSC 10/11/12` from the live theme (the #331 fix). That also gets the thing this PR could not — live theme changes and already-open panes — for free, because every query is answered from the theme at the moment it is asked.

Given that, the bootstrap design is solving a problem that disappears one layer down, and the two blockers from my review (the GUI-subsystem re-exec, the missing `SetConsoleCtrlHandler`) do not need solving. I've opened #360, which is packaging only: vendor the pair, stage it in the bundle and the installer, add it to the updater's managed roots, fail the release build on a mismatched or missing half.

I'll leave this open until #360 lands so you can push back on any of it — if the measurements don't hold up on your machine I'd rather know before it merges. Your `set_console_palette` is the part I'd have kept if the table route had been necessary, and the empirical work in #345 — establishing that the terminal-side handles reject the console APIs and that only a process *inside* the pseudoconsole can set the table — is what made it obvious that the fix had to move below that layer. Sorry for the round trip, and thank you for both.

l0ng-ai:
#360 landed (3ae3404), and #345 closed with it, so this can close too. Recording why here rather than just hitting the button, since the work in this PR is what got us there.

What shipped is the packaging route: Microsoft's redistributable `conpty.dll` + `OpenConsole.exe` staged beside `tty7-app.exe`, which `portable-pty 0.8.1` already prefers over `kernel32`'s. No console color table, no wrapper process, no PID handoff. The measurements from my last comment held up through review — in a real pane under `catppuccin_latte` the `OSC 11` reply comes back `#eff1f5`, the preset's exact background, answered by `terminal::view` from the live theme. Which means live theme switches and already-open panes work too, the one thing the seed-at-spawn approach couldn't reach.

One gap the table route would have covered and this one doesn't: an app that reads `GetConsoleScreenBufferInfoEx` and never sends an OSC probe still sees the Windows console defaults. codex tries OSC first and falls back, so it's fine; I don't know of a real app that's console-API-only. If one turns up, your `set_console_palette` is where I'd start — it's the part of this diff I'd have kept, and the two blockers I raised (the GUI-subsystem re-exec, the missing `SetConsoleCtrlHandler`) are both solvable, just not worth a bootstrap process on spec.

The empirical work in #345 is what made the fix findable: establishing that the terminal-side handles reject the console APIs, and that only a process *inside* the pseudoconsole can set the table, is what pointed at the layer below. Thank you for both, and sorry for the round trip.

Reopen if you think the console-API gap is worth covering now.

Issue #363, opened 2026-08-06
Remote server downloads ignore Windows system proxy


shihuaidexianyu:
On Windows, when a tool like Clash enables **System Proxy**, it writes the proxy address into the registry at:

```
HKCU\Software\Microsoft\Windows\CurrentVersion\Internet Settings
  - ProxyEnable
  - ProxyServer
  - ProxyOverride
```

tty7 currently behaves inconsistently with this setting:

- **Update check** (`src/core/update.rs`) uses `reqwest_client`, which reads the Windows system proxy from the registry by default. It works through Clash.
- **Remote server installer / bundled-server fallback** (`crates/tty7-core/src/daemon/install/download.rs`) uses `ureq` with only the `rustls` feature enabled. `ureq` reads `HTTP_PROXY`/`HTTPS_PROXY` environment variables, but **does not read the Windows registry** unless the `win-system-proxy` feature is turned on.

This leads to a confusing situation: the app can detect a new release through the proxy, but the actual download times out trying to reach GitHub directly.

## Proposed fix

Enable `ureq`'s `win-system-proxy` feature in `crates/tty7-core/Cargo.toml`:

```toml
ureq = { version = "3", default-features = false, features = [
    "rustls",
    "win-system-proxy",
], optional = true }
```

This makes `ureq` fall back to the registry-based system proxy on Windows when no environment variable is set. It is a no-op on macOS and Linux.

## Scope

This affects release asset downloads performed by the daemon:

- Remote `tty7-server` install / upgrade
- Bundled-server fallback when the local copy is missing
- Update package download

SSH connections (russh) are TCP-based and are not affected by HTTP system proxies, which is the expected behavior.

## Alternatives considered

- Rely on environment variables: insufficient for GUI-launched apps on Windows, because `tty7-app.exe` is typically started from the Start Menu / Explorer and does not inherit shell environment variables. The daemon inherits its environment from the GUI, so it would also miss the proxy.
- Add a manual proxy setting in Settings: more flexible for PAC/corporate environments, but overkill for the common Clash/v2rayN System Proxy case.

PR #367, opened 2026-08-06
feat(proxy): macOS system proxy, Windows SOCKS parsing, manual override

### LINKED ISSUE #365: Improve proxy support: macOS system proxy, Windows SOCKS, and manual override
Following up on #363 / #364, the one-line `win-system-proxy` fix covers the common Windows + Clash HTTP system-proxy case. There are still three related gaps that affect real-world usage:

## 1. macOS system proxy is ignored by release downloads

The GUI update check uses reqwest, which reads macOS system proxy settings via `SCDynamicStore`. However, the remote server installer / bundled-server fallback uses `ureq`, which has no macOS system-proxy feature. On macOS, release downloads therefore bypass the system proxy if no environment variable is set.

## 2. Windows SOCKS-only system proxies are mis-parsed

When ureq's `win-system-proxy` feature reads the registry, it always constructs the proxy URL as `http://{ProxyServer}`. If the registry value is SOCKS-only, e.g. `socks=127.0.0.1:1080`, ureq treats it as an HTTP proxy and fails to connect.

Clash's default mixed port works because it writes an HTTP proxy like `127.0.0.1:7890`, but other tools or user configurations may use `socks=...`.

## 3. No per-app manual proxy setting

Users cannot override the system proxy just for tty7. This is useful when:
- They don't want to enable system-wide System Proxy.
- The system proxy is wrong (e.g. corporate PAC that tty7 cannot evaluate).
- They need tty7 to use a different proxy than the rest of the OS.

## Proposed scope

- Add a `http_proxy: Option<String>` config field (manual override).
- Add a Settings row where users can enter `http://127.0.0.1:7890`, `socks5://127.0.0.1:1080`, etc.
- Build a small cross-platform helper in `crates/tty7-core` that resolves the effective proxy in order:
  1. Manual config `http_proxy`
  2. macOS: `SCDynamicStore` system proxy
  3. Windows: registry `ProxyEnable`/`ProxyServer` (correctly parse `http=...;https=...;socks=...`)
  4. Environment variables (`HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY`)
- Enable ureq's `socks-proxy` feature so SOCKS URLs work.
- Make the GUI update check (`src/core/update.rs`) respect the manual config too.

SSH connections remain unaffected (expected — they are TCP, not HTTP).

shihuaidexianyu:
Fixes #365.

This closes the three remaining proxy gaps from #363 / #364:

1. **macOS system proxy for release downloads** — the GUI update check already reads `SCDynamicStore` via reqwest, but the daemon's `ureq` downloader did not. The new resolver reads `HTTPEnable`/`HTTPProxy`/`HTTPPort`, HTTPS, SOCKS and `ExceptionsList` from `SCDynamicStoreCopyProxies`.

2. **Windows SOCKS-only system proxies are no longer mis-parsed** — ureq's built-in `win-system-proxy` feature always builds `http://{ProxyServer}`, so a registry value like `socks=127.0.0.1:1080` was treated as HTTP. The new registry parser handles `http=...;https=...;socks=...` and maps `socks=...` to `socks5://...`.

3. **A per-app manual proxy setting** — a new `http_proxy` config field overrides the system proxy just for tty7. A row in **Settings → About → HTTP/SOCKS proxy** accepts values like `http://127.0.0.1:7890` or `socks5://127.0.0.1:1080`.

## Resolution order

`crates/tty7-core/src/daemon/install/proxy.rs::resolve()` picks the first available of:

1. `Config::http_proxy` (manual override)
2. Windows: `HKCU\...\Internet Settings` (`ProxyEnable`, `ProxyServer`, `ProxyOverride`)
3. macOS: `SCDynamicStoreCopyProxies`
4. `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` environment variables

## What changed

- `crates/tty7-core/Cargo.toml`
  - Enabled `ureq`'s `socks-proxy` feature.
  - Added Windows `winreg` and macOS `system-configuration` + `core-foundation` dependencies.
- `crates/tty7-core/src/core/config.rs`
  - Added `http_proxy: Option<String>`.
- `crates/tty7-core/src/daemon/install/proxy.rs` (new)
  - Cross-platform resolver + Windows registry parser + macOS SCDynamicStore reader.
- `crates/tty7-core/src/daemon/install/download.rs`
  - `HttpsFetcher` now configures its `ureq::Agent` from the resolver.
- `crates/tty7-core/src/daemon/install/mod.rs`
  - `default_fetcher()` passes the configured `http_proxy` into `HttpsFetcher`.
- `src/core/update.rs`
  - `fetch_latest_release()` uses `ReqwestClient::proxy_and_user_agent()` when a manual proxy is set.
  - `prepare_update()` reuses the same `HttpsFetcher` path, so the manual/system proxy also covers update downloads.
- `src/ui/app.rs`, `src/ui/settings.rs`, `src/ui/i18n.rs`
  - Added the **HTTP/SOCKS proxy** input row in Settings → About, with English and Chinese labels, description, and search keywords.

## Verified

- `cargo check -p tty7-core --features remote-install` passes.
- `cargo check -p tty7` passes.
- `cargo test -p tty7-core --features remote-install proxy` passes (Windows `ProxyServer` parsing tests).
- `cargo test -p tty7 settings::tests` passes (settings search index includes the new row).

## Notes

- SSH connections remain TCP-based and are intentionally unaffected.
- PAC / WPAD is still not supported; this covers explicit HTTP/SOCKS proxies only.


l0ng-ai:
Pushed a follow-up commit to this branch (c0cc194) — hope you don't mind me editing directly, happy to move it to a separate PR if you'd rather review it that way.

The one that blocks: the macOS reader mapped an `HTTPS proxy` entry to a `https://` URL, and ureq reads that as `ProxyProtocol::Https`, i.e. a CONNECT proxy reached **over TLS** (see the doc comment on the enum). macOS' HTTPS proxy is a plain HTTP proxy that happens to be used for https traffic, so the common Clash/Surge setup — 7890 in both the HTTP and HTTPS fields — would TLS-handshake against a cleartext port and fail every download. It now emits `http://` for both entries, which is what the Windows path was already doing via `normalize_proxy_addr`.

Second one: `parse_proxy_server` used `?` on each `;`-separated part, so a trailing `;` or an `ftp=` entry made the whole function return `None` and silently disabled the proxy. It now skips parts it can't parse.

The rest is cleanup:

- Reuse ureq's own `ProxyProtocol: TryFrom<&str>` and let `ProxyBuilder` fill in default ports — `proxy_protocol` and `default_port` were byte-for-byte copies of ureq's tables.
- Keep `tty7-core`'s core-foundation at 0.10 for `daemon::pane` and reach the 0.9 types through `system_configuration`'s `pub extern crate core_foundation` re-export, so the crate-wide downgrade isn't needed.
- Gate `winreg` / `system-configuration` behind `remote-install`, since that's the only thing using them.
- Move the `ProxyServer` parsing out of the `windows` module so its tests compile and run on every host — as written they only existed on the Windows runner, so `cargo test … proxy` on a Mac was running zero tests. Added coverage for the manual-proxy path too (16 tests total).
- Normalise `http_proxy` once in `Config::sanitize`, and route the update check through the same normaliser as the downloader so a bare `127.0.0.1:7890` proxies both (previously it only reached ureq; `Url::parse` in `fetch_latest_release` saw the raw value).

One product note, which is why the settings copy changed: in a *terminal*, a row labelled "HTTP/SOCKS proxy" reads as "this proxies what I run in my panes", which it doesn't. It's now "Proxy for updates" and the description says explicitly that programs in a pane are unaffected. An unparsable value is also rejected with an inline hint instead of being stored silently.

Local: `cargo test` green for both crates, `cargo fmt --check` clean, and the no-`remote-install` / `tty7-server` builds still check. I could not build the `x86_64-pc-windows-msvc` target locally (no MSVC toolchain for `ring`/`aws-lc-sys`), so the Windows path is on CI.

PR #373, opened 2026-08-06
feat(windows,macos): clickable toasts, richer context, and i18n

### LINKED ISSUE #371: Windows notifications: click-to-reveal, richer context, and i18n
## Problem

Windows desktop notifications in tty7 currently have three gaps:

1. **Not clickable.** `src/terminal/remote.rs:notify_desktop` drops the notification handle after showing it. Clicking the toast only foregrounds the app via the Start Menu shortcut; it cannot jump to the specific pane/workspace that produced the notification.
2. **Too little context.** The finish toasts in `src/terminal/view.rs` only say `"Command finished after {secs}s"` or `"Finished after {secs}s"`. They do not mention the machine, workspace, or shell that completed the work.
3. **No i18n.** Those same strings are hard-coded English `format!` calls. `src/ui/i18n.rs` already provides thread-safe `t()`/`t_fmt()` functions that do not require a `cx`, so localization is purely a plumbing issue.

## Goals

- Clicking a completion toast on Windows reveals the originating pane, tab, and window.
- Toast title/body include machine + workspace + shell + command/agent + duration.
- All user-facing toast strings go through `L10nKey` so they can be translated.

### 1. Clickable Windows toasts

- Extend `notify_desktop` to accept an optional `leaf_id: Option<u64>`.
- On Windows, when `leaf_id` is present, use the `windows` crate (already a dependency for AUMID) to show a `ToastNotification` with an `Activated` handler.
- The handler sends `TrayAction::RevealPane { leaf_id }` through the existing tray dispatch channel (`src/ui/tray/mod.rs`). This reuses the existing routing that finds the owning window, activates the tab, and focuses the pane.
- Non-targeted toasts (OSC 9/777 passthrough, errors) keep using `notify-rust`.

### 2. Richer payload

Change the notification helpers in `src/terminal/view.rs` to build title/body from data already in scope:

- `machine` / `workspace`: from `self.host_id` + `owner_workspace()` + `machine_mirror::display_name_for`.
- `shell` / `command`: from `running_title` / `title`.
- `agent`: from `agent.display_name()`.
- `duration`: already computed.

Example Windows toast:

```
Title: xiphos · web-project
Body:  npm build — finished after 12s
```

### 3. i18n keys

Add `L10nKey` variants in `src/ui/i18n.rs`:

- `NotifyCommandFinished`
- `NotifyCommandFinishedWithCommand`
- `NotifyAgentFinished`
- `NotifyAgentWaiting`
- `NotifyTurnFinished`

Replace the hard-coded `format!` strings in `src/terminal/view.rs:296-311` and `2682-2702` with `t_fmt(...)` calls, providing English and Chinese (Simplified) translations.

## Files expected to change

- `Cargo.toml` — add `UI_Notifications` / `Data_Xml_Dom` features to the existing `windows` dependency.
- `src/terminal/remote.rs` — `notify_desktop` plus a Windows-specific toast module.
- `src/terminal/view.rs` — pass context into notification helpers and build richer strings.
- `src/ui/tray/mod.rs` — expose a way for the toast activation handler to send `TrayAction::RevealPane`.
- `src/ui/i18n.rs` — new notification keys and translations.

## Notes / open questions

- macOS/Linux: keep using `notify-rust`; click-to-reveal can be added later with platform-specific APIs.
- Should the toast also include action buttons such as "Dismiss" or "Copy output"?
- Should we keep `notify-rust` on Windows for the no-AUMID fallback path?

shihuaidexianyu:
Closes #371

Depends on #367 (this branch currently includes those changes because the i18n infrastructure and AUMID setup live there).

## What changed

- **Clickable desktop notifications (Windows + macOS).** `notify_desktop` now accepts an optional `leaf_id`.
  - On Windows we show a WinRT toast with an `Activated` handler.
  - On macOS we use `mac-notification-sys`'s `wait_for_click` API and handle `NotificationResponse::Click`.
  - Both handlers send `TrayAction::RevealPane { leaf_id }` through the existing tray dispatch channel, so clicking the toast reveals the originating pane, tab, and window.
- **Richer notification content.** Titles now show `{machine} · {workspace}` and bodies include the shell/command or agent name plus duration, instead of only `"finished after Xs"`.
- **i18n.** Added `L10nKey` variants for command/agent/turn finish and agent-waiting messages, with English and Chinese (Simplified) translations.

## Files changed

- `Cargo.toml` / `Cargo.lock` — added `mac-notification-sys` on macOS; added `Foundation`, `UI_Notifications`, and `Data_Xml_Dom` features to the existing `windows` dependency.
- `src/terminal/remote.rs` — `notify_desktop_at` + platform-specific clickable toast paths.
- `src/terminal/view.rs` — build richer titles/bodies and route them through the new API.
- `src/ui/tray/mod.rs` — expose a global sender so toast activation callbacks can enqueue `TrayAction::RevealPane`.
- `src/ui/i18n.rs` — new notification keys and translations.

## Notes

- Non-targeted toasts (OSC 9/777 passthrough, errors) still use `notify-rust`.
- Linux keeps using `notify-rust`; click-to-reveal can be added later with a D-bus action listener.
- `cargo check -p tty7 --target x86_64-pc-windows-msvc` passes (Windows + macOS code is `cfg`-gated, so the macOS path will be exercised by CI on macOS).

## TODO before undrafting

- [ ] Rebase onto `main` after #367 lands.
- [ ] Manual test on Windows: click a long-running command toast and verify it reveals the correct pane.
- [ ] Manual test on macOS: click a long-running command toast and verify it reveals the correct pane.
- [ ] Verify Chinese translations look natural in the toast UI.


l0ng-ai:
Pushed `923bed3` onto this branch (fast-forward, nothing of yours rewritten). Happy to drop or rework any of it — here is what it changes and why.

**The one that breaks the feature**

`notify_desktop_at` was handed `self.pane_id`, but `TrayAction::RevealPane` matches leaves on their gpui entity id (`leaf.entity_id().as_u64()`, see `ui/app.rs` `agent_rows` / `owns_leaf`). Those are different id spaces, so `tab_ix` never matched and clicking a toast only raised the window — the reveal never happened. Now it passes `cx.entity_id()`, and the parameter takes an `EntityId` instead of a bare `u64` so the two numbers can't be swapped at a call site again.

**Delivery paths**

- *Windows AUMID.* `toast_app_id().unwrap_or("tty7")` named an id the shell has never indexed, and that is exactly the case the comment above the notify-rust path warns about: `Show` reports success and the toast is dropped. That is worse than the status quo (a PowerShell-branded toast). Now an unregistered AUMID falls back to the notify-rust path.
- *Windows toast lifetime.* One dedicated thread now builds, shows and holds the 32 most recent toasts, instead of a thread per toast sleeping 60s. A toast stays clickable for as long as Action Center keeps it, threads don't accumulate, and the WinRT objects never cross a thread boundary.
- *macOS app identity.* `Notification::send` pins the delivering application under a `Once` on first use, so skipping `ensure_notification_app()` let the `com.apple.Finder` default win — permanently, for every later notification including the notify-rust ones. It now runs before the first send.
- *macOS thread lifetime.* `wait_for_click` blocks with no timeout (`rust_wait_for_notification` waits on a condvar that only a click or dismissal releases), so a banner nobody touches parks its thread for the rest of the session. Capped at 8 outstanding; past that we deliver fire-and-forget.
- *Text sanitizing.* Notification text comes off the terminal, and a stray `ESC` in a window title makes the toast XML fail to parse — losing the whole notification with only a `warn!`. Control characters are stripped and the text clamped (96 / 512 chars) on every path.

**Smaller things**

- Titles stay at two segments (agent name or machine, then workspace). Three segments overflow a Windows toast title, which is a single line. A local pane with an unnamed workspace now reads `tty7` rather than `this computer · Untitled`.
- The tray sender is replaceable instead of a `OnceLock`: `init` runs again whenever the window count returns from zero, and pinning the first channel kept its dead dispatch loop alive while sending later clicks nowhere.
- Dropped the ureq `win-system-proxy` line, which belongs to #367. Its two commits are still in this branch's history — a rebase after #367 lands will clear them.
- Tests for the sanitizer, the title composition and `xml_escape`; `xml_escape` is no longer `cfg(not(test))`, so Windows CI actually covers it.

**Verification, honestly**

- `cargo test -p tty7 --bin tty7-app`: 976 passed. macOS `cargo check` clean.
- The Windows path can't be cross-compiled here (aws-lc-sys wants MSVC), so I copied it verbatim into a throwaway crate depending only on `windows 0.58` and checked it for `x86_64-pc-windows-msvc` — clean. That verifies API usage and types, not runtime behaviour.
- **Neither platform has been clicked on real hardware.** Your two manual-test TODOs still stand; the Windows one especially, since the toast thread is new code.

PR #385, opened 2026-08-07
fix(terminal): add 0.5px overlap to eliminate 1px gaps in powerline rendering

cwatanab:
### Summary
Added a 0.5px overlap for cell bounds and ensured proper background quad rendering to resolve 1px vertical gap lines between Powerline segments (e.g., half-circle caps and triangle arrows).

### Visual Improvements
- **Before**: Noticeable 1px vertical gap lines appeared at cell boundaries (e.g., between the left purple cap and pink segment, and between triangle arrows).
- **After**: All vertical gap lines are completely eliminated, rendering smooth and seamless continuous Powerline segments.

### DIFF
--- a/src/terminal/element.rs
+++ b/src/terminal/element.rs
@@ -672,6 +672,30 @@ fn powerline_path(bounds: Bounds<Pixels>, shape: PowerlineShape) -> gpui::Path<P
         p.line_to(c);
         p
     };
+    let half_circle = |anchor_x: Pixels, dir: f32| {
+        const SEGS: usize = 6;
+
+        let (rx, ry) = (bounds.size.width.as_f32(), bounds.size.height.as_f32() / 2.);
+
+        let at = |scale: f32, theta: f32| {
+            let x = anchor_x.as_f32() + dir * rx * scale * theta.cos();
+            let y = ymid.as_f32() + ry * scale * theta.sin();
+            point(
+                px(x.clamp(x0.as_f32(), x1.as_f32())),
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Thanks for the report and the proposed fix! This seam was already addressed on `main` by #336 (merged 2026-08-04): when a separator's solid closing edge falls between device pixels, the anti-aliased path edge lets the background leak through as a 1px vertical line. #336 fixes it by painting an opaque cover quad over exactly that fractional device-pixel column (`powerline_solid_edge` in `src/terminal/element.rs`), which keeps the cell geometry unchanged and preserves anti-aliasing on the slopes and curves — so the 0.5px bounds overlap shouldn't be needed on top of it.

The fix just missed the v26.8.1 release (2026-08-01), which is why stable still shows the gaps. I reproduced the exact artifact you describe by disabling #336's cover quad on current `main`, and with it enabled the seams are gone at every separator boundary (verified with a pixel-level scan at 150% display scale on Windows).

Could you try a nightly build? If the gaps are gone there, this can be closed as fixed by #336. If you still see seams on nightly, please share your OS, display scale factor, and font size — that would mean there's a case the closing-edge cover doesn't handle, and we'd definitely want to dig into it.

cwatanab:
I confirmed that the issue is resolved in nightly `26.8.2-nightly.202608071638`.
Tested on Windows 11 (25H2), and the seams at all separator boundaries are completely gone.

Thanks for the fix! This can be closed as resolved by #336.

PR #407, opened 2026-08-08
fix(windows): never join the pane reader on the UI thread

l0ng-ai:
## Problem

On Windows, connecting (reconnecting) to a remote workspace could freeze the whole window — Windows marks it *Not Responding*. Force-killing and relaunching "fixes" it until the next time, because the trigger is stale state: old reader threads parked on a link whose SSH leg has gone zombie (sleep/resume, network blip — exactly the situations that cause a reconnect).

## Root cause

`detach_link`, `adopt_relink`, and `Drop for RemoteTerminal` all tear a pane link down as **shutdown the writer, then `join()` the reader thread** — and all three run on the UI thread (`relink_panes` right after a reconnect succeeds, `on_release_in` when a pane closes, preemption release).

That design counts on `shutdown()` waking the reader out of its blocking `read()`. It does on unix; **on Windows it does not** — Winsock `shutdown(SD_BOTH)` leaves a parked `recv` parked (verified with a minimal probe: recv still blocked 5s after shutdown). So the join only returns when the peer closes the connection. A routed pane's peer is the local daemon proxying an SSH leg; when that leg is a zombie, no FIN ever comes, and the join hangs the UI thread indefinitely.

The reader teardown had zero Windows coverage: the whole reader test module is `cfg(all(test, unix))`.

## Fix

Never join. Teardown now:
- sets a per-reader `quit` flag, shuts the writer down, and **abandons** the thread;
- the reader's `read` always times out within 500 ms, so even a reader parked on a silent link notices `quit` promptly and exits on its own;
- the reader re-checks `quit` **under the term lock before every grid mutation**, and `adopt_relink` sets the flag *before* taking the lock to reset the grid — so a retired reader can never touch the freshly adopted link's grid, and it no longer marks the pane as exited.

## Tests

New `cfg(all(test, windows))` module:
- `detach_link_returns_promptly_when_the_peer_stays_silent` — fails on the old code by blocking >3 s (observed; it actually blocks forever);
- `adopt_relink_swaps_links_promptly_when_the_old_peer_stays_silent` — the reconnect path: swap must not block, output on the adopted link must reach the grid, and the pane must not be marked exited.

Full suite: 906 passed / 0 failed on Windows. Clippy clean for the touched file.

### DIFF
--- a/src/terminal/remote.rs
+++ b/src/terminal/remote.rs
@@ -185,6 +185,14 @@ pub struct RemoteTerminal {
     route: PaneRoute,
     proxy: EventProxy,
     reader_thread: Option<JoinHandle<()>>,
+    /// Tells the *current* reader thread to bow out. Teardown must never wait
+    /// for the reader: on Windows, `shutdown()` does not wake a thread parked
+    /// in a blocking `read()` on the same socket, so joining it from the UI
+    /// thread hangs the whole window whenever the peer has gone silent (the
+    /// exact state a zombie SSH leg leaves behind). The reader re-checks this
+    /// flag under the term lock before every grid mutation, so once it is set
+    /// the abandoned thread can only exit, never write.
+    reader_quit: Arc<AtomicBool>,
 }
 
 /// The workspace id a spawn carries, so the pane's shell gets `$TTY7_WS` and a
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Addressed the review finding in 14acb9a: the quit flag is now checked once per frame at the top of the inner take_frame loop, so a retired reader stops processing its buffered frames entirely — none of the Cwd/Prompt/Agent/Exited arms can touch the shared state that survives a relink.

On the child_exited scenario specifically: the adopt test now also fires a late `Exited` frame down the abandoned link after the swap, waits out a full quit-poll tick, and asserts the pane is neither child-exited nor torn down. (The exact interleaving the review describes — frames already sitting in `pending` at the instant quit is set — isn't reachable deterministically from outside the thread, but with the per-frame gate the property holds by construction: every frame dispatch is preceded by a quit check, and quit is set before the swap.)

Full suite still green: 906 passed / 0 failed.

PR #410, opened 2026-08-08
fix(workspace): stop switching workspaces from destroying live sessions

l0ng-ai:
Switching workspaces could throw away every running session in the workspace being left, and in the worst case delete that workspace outright — from the machine tree and from `views.json` both.

Three independent guesses were each authorizing an irreversible act. All three are replaced by an answer from whoever actually holds the truth.

| | Before | After |
|---|---|---|
| Restoring a pane the tree marks `live: false` | Pane id erased, fresh shell spawned over the running one | Id goes down regardless; attaching decides, and spawns fresh only when the pane really is gone |
| A second server starting against one config dir | Sockets split between two processes, neither aware | Second server finds the seat taken and exits |
| Deciding another server is dead | One failed `connect()`, then unlink its socket | Advisory lock — held means alive, released by the kernel however the holder dies |
| `MachineGet` when the control server owns no panes | Every pane reported dead, no error anywhere | Cannot arise: one server owns both endpoints |
| Window that rebuilt zero tabs | Marked informed, then allowed to delete the workspace | Informed only if the rebuild produced something |
| Deleting an "empty" workspace | Window's own emptiness was enough | Window **and** the machine's mirror must agree it is empty |

## Why

The three bugs share one shape: **an irreversible destruction authorized by an inference that can be wrong.** The costs are not symmetric — guessing "dead" when a pane is alive loses a session for good, while guessing "alive" when it is dead costs one failed attach.

How they combined, as observed:

1. Two servers started 8 seconds apart. The first bound `control.sock`; the second failed to reach it once, unlinked `daemon.sock` and bound its own. Both stayed up — the first serving the machine tree with an empty pane registry, the second holding every pane.
2. `MachineGet` therefore reported all 34 panes dead. `tty7 status` said `panes 0` while 19 agent sessions were running.
3. Switching a workspace rebuilt all of them, replacing live shells with fresh ones.
4. After a restart left the pane socket unreachable, every tab failed to rebuild. The window was marked informed anyway, and the next switch deleted three workspaces — 20 tabs — from the tree and the store.

```mermaid
flowchart TD
    A[server starts] --> B{lock free?}
    B -- no --> C[exit: another server serves this dir]
    B -- yes --> D[bind endpoints, serve]
    D --> E[client opens a workspace]
    E --> F{pane attaches?}
    F -- yes --> G[reuse the running shell]
    F -- no --> H[spawn fresh, resume the agent]
    E --> I{rebuild produced tabs?}
    I -- no --> J[stay uninformed: emptiness is not a fact]
    I -- yes --> K[informed]
    K --> L{delete an empty workspace?}
    L --> M{mirror agrees it is empty?}
    M -- no --> N[keep it]
    M -- yes --> O[delete]
```

## Testing

- `a_pane_the_tree_calls_dead_still_goes_down_by_its_id` — the id survives `live: false`, so the attach gets to settle it
- `a_lowered_leaf_carries_its_pane_id_and_its_agent_whatever_live_says` — updated; it previously asserted the behavior that caused the loss
- `only_a_mirror_that_agrees_lets_an_empty_window_delete_its_workspace` — four cases, including the regression: informed window, mirror still holding tabs, deletion refused
- `the_second_claim_is_refused_while_the_first_is_held`, `a_lock_left_behind_by_a_dead_holder_is_claimable` — the lock excludes a live holder and is not fooled by a leftover file
- Full suites green: 855 (`tty7-core`), 1052 (`tty7`)
- Manual, on the machine where this happened: built and installed, then confirmed one daemon process, `daemon.lock` present, `tty7 status` reporting real pane counts (was `0`), and `tty7 pane ls` showing `LIVE yes` (was `no` for every pane). Restarting only the GUI reattached the running panes instead of rebuilding them.
- Follow-up commit: a corrupt `views.json` is now quarantined (`views.json.corrupt`) before being ignored, the way `machine.json` already was — previously the next save silently overwrote it. The quarantine helpers moved to `config` so both loaders share them, and the no-lock-primitive platform fallback in `singleton` now reports `Unavailable` instead of `Taken`. Guard: `a_corrupt_views_file_is_kept_aside_before_being_ignored`.

### DIFF
--- a/crates/tty7-core/src/core/config.rs
+++ b/crates/tty7-core/src/core/config.rs
@@ -622,6 +622,39 @@ pub fn strip_bom(text: &str) -> &str {
     text.strip_prefix('\u{FEFF}').unwrap_or(text)
 }
 
+/// Sets a corrupt state file aside (copied, the original left in place) so the
+/// caller can fall back to defaults without silently destroying what was there.
+pub(crate) fn quarantine(path: &std::path::Path) {
+    let aside = quarantine_path(path);
+    match std::fs::copy(path, &aside) {
+        Ok(_) => log::warn!("the previous contents were kept at {}", aside.display()),
+        Err(e) => log::warn!("could not keep a copy at {}: {e}", aside.display()),
+    }
+}
+
+/// Like [`quarantine`], but moves the file out of the way — for files that
+/// cannot even be read, where copying would fail too.
+pub(crate) fn quarantine_by_rename(path: &std::path::Path) {
+    let aside = quarantine_path(path);
+    match std::fs::rename(path, &aside) {
+        Ok(()) => log::warn!("the previous contents were moved to {}", aside.display()),
+        Err(e) => log::warn!("could not move the file to {}: {e}", aside.display()),
+    }
+}
+
+fn quarantine_path(path: &std::path::Path) -> PathBuf {
+    const MAX_QUARANTINED: u32 = 8;
+
+    let base = path.with_extension("json.corrupt");
+    if !base.exists() {
+        return base;
+    }
+    (1..MAX_QUARANTINED)
+        .map(|n| path.with_extension(format!("json.corrupt.{n}")))
+        .find(|candidate| !candidate.exists())
+        .unwrap_or(base)
+}
+
 pub fn write_atomic(path: &std::path::Path, bytes: &[u8]) -> std::io::Result<()> {
     write_atomic_mode(path, bytes, false)
 }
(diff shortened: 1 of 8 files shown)


l0ng-ai:
All three review points addressed in 5bc6c68:

- **`alive_panes_on` fail-open**: a failed `List` RPC now surfaces as `None` instead of an empty alive-set, and `pane_attachable(None, ..)` answers "attachable" — the attach itself settles it, the same way `session_from_tree` now leaves it to. If the daemon really is unreachable, the attach fails and the pane falls to the fresh-spawn path as before; a transient error no longer respawns a whole batch of live sessions. Guarded by a new case in `restore_only_attaches_panes_the_workspace_owns_or_nobody_claims`.
- **`pub fn run()` bypassing the lock**: removed — it had zero callers, and any future one would have silently reintroduced the split-brain.
- **`Singleton::path()`**: removed along with the field behind it.

Suites green: 856 (`tty7-core`), 1052 (`tty7`).

PR #415, opened 2026-08-08
fix(terminal): resize ConPTY panes with conhost's semantics, in stream order

l0ng-ai:
Maximizing a pane after (or during) a burst of output garbled it: old output interleaved with later prompts, cursor stranded mid-screen. Two distinct resize disagreements stacked up.

## 1. Semantics: conhost re-anchors, the grid disagreed (the visible bug)

ConPTY emits **no repaint** after `ResizePseudoConsole` - measured against a live pane, a quiescent resize produces zero bytes. Conhost just re-anchors its own layout and keeps painting with absolute cursor addresses computed against it:

- **Grow**: rows and cursor stay pinned; blank rows open below. Nothing returns from scrollback. (Measured: prompt at row 29 of 30; after growing to 50 rows, PSReadLine still paints at `\e[29;36H`.)
- **Shrink**: the last *written* row scrolls to the new bottom - PSReadLine parks a continuation hint below the prompt, and conhost counts it. (Measured: cursor 37 of 50 -> row 19 of 20, one row above the alacritty-style answer.)

The vendored grid resized the alacritty way: growing pulled scrollback into view and pushed the cursor down by that many rows. After a maximize every absolute-CUP paint - PSReadLine redraws the prompt per keystroke - landed that many rows off, mid-screen inside the old output.

Fix: `alacritty_terminal` fork gains a `conpty_resize` mode (l0ng-ai/alacritty@1276f128, on the `tty7` branch, with grid tests) mirroring conhost's model for the primary screen; the alt screen keeps stock behavior since its application repaints itself. Every Windows pane opts in - shells, `wsl.exe`, `ssh.exe` are all presented through conhost alike. Cost: maximizing no longer reveals extra scrollback below the fold, same as Windows Terminal.

## 2. Stream order: reflow raced the output backlog

The client reflowed its grid the instant the window changed, then kept parsing whatever the daemon still had queued - up to the gate's 16 MiB of old-width bytes - into the new-width grid. The daemon now echoes a `Size` frame to the controlling subscriber *from inside the pane state lock*, i.e. at the exact stream position where the PTY geometry changes (after every old-width byte, before the post-resize output), and the client defers its reflow to that marker - the same geometry-tagging the reattach replay has always used. Observers apply live `Size` frames at the right moment now too instead of holding them for a snapshot that never comes.

Gated on a new `resize-echo` daemon feature: against an older daemon (the normal state right after an in-place update) the client keeps the old reflow-at-request-time path.

## Verification

- Fork: 141 `alacritty_terminal` tests pass incl. 3 new ConPTY grid tests.
- Daemon: unit test asserts `Output -> Size -> Output` stream order; live probe against an isolated daemon confirms exactly one echo with correct geometry mid-flood.
- Client: new cfg(unix) test asserts pre-echo bytes parse at the old width and post-echo bytes at the new width.
- Manual acceptance in an isolated dev instance: maximize after/during heavy output stays clean; post-maximize commands paint at the prompt; shrink and drag-resize track correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,22 +130,42 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Maximizing a pane mid-burst no longer garbles it** — the client reflowed
-  its grid the instant the window changed, then went on parsing whatever the
-  daemon still had queued for it, and that backlog (up to the 16 MiB the
-  output gate allows) was produced at the old width. Old-width bytes laid
-  out on a new-width grid, topped by ConPTY's own post-resize repaint, left
-  command output interleaved with prompts and the cursor stranded
-  mid-screen. Reattach never had this bug: its replay tags every ring
-  segment with the geometry it was recorded at. A live resize now works the
-  same way — the daemon echoes a `Size` frame to the controller at the
-  exact stream position where the PTY changed geometry, and the client
-  defers its reflow to that marker, so every queued byte still parses into
-  the grid it was rendered for. Observers, which already got live `Size`
-  frames but only ever applied them after a replay snapshot, now follow the
-  controller's resizes at the right moment too. A client facing an older
-  daemon that never echoes keeps the reflow-at-request-time behavior (new
-  `resize-echo` feature probe).
+- **Maximizing a Windows pane no longer shreds it** — two separate resize
+  disagreements stacked up here, and the visible one was the second.
+
+  ConPTY emits no repaint after a resize; conhost silently re-anchors its
+  own layout and keeps painting with absolute cursor addresses computed
+  against it. Measured against a live ConPTY: growing the window keeps
+  rows and cursor pinned and opens blank rows at the bottom — nothing is
+  restored from scrollback — and shrinking scrolls the last *written* row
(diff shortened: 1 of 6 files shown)


l0ng-ai:
Review findings addressed in 2995bb0:

**1. Remote routes** — correct, and it's deliberate: only the local daemon's feature set is probed, so the client can't *rely* on an echo from a remote daemon; deferral must not hang on a maybe. The inaccurate doc comment ("never send the echo") is fixed — a current remote daemon does echo, the router forwards it, and it lands as a harmless same-size no-op after the eager reflow. Per-connection feature probing is the follow-up that extends deferral to remote panes; the scope limit is now stated in the `resize_echoed()` docs and the CHANGELOG.

**2. Reader mid-chunk race** — real, considered, and accepted: even with perfect daemon-side ordering, conhost keeps producing old-width bytes into the kernel pipe until it processes the resize, and neither residue (one 64 KiB read + the pipe) can be attributed to a geometry — the bytes carry no tags and ConPTY emits no sync marker. This is the same ambiguity every ConPTY terminal accepts on a mid-output resize; the echo exists to close the 16 MiB gate window, not this one. Now documented on `resize_state()`.

Also restored the full two-mechanism CHANGELOG entry (conhost resize semantics + stream-order echo) — the committed version had lost the semantics half.

PR #421, opened 2026-08-09
feat(terminal): include fish_history in command history search

dongyuwei:
Ctrl+R command history currently only reads .zsh_history / .bash_history; fish users' fish_history (YAML - cmd: / when: records) is never parsed or read, locally or on remote panes.
- parse_fish_history: record-by-record tolerant YAML parsing — a single malformed record (common in files accumulated across fish 3.x/4.x writers) no longer discards the whole history
- load_shell_history: includes $XDG_DATA_HOME/fish/fish_history, falling back to ~/.local/share/fish/fish_history
- shell_history_names: remote panes' far-end history sources now include fish
- 6 new tests; full suite: 1073 passed

## Manual verification

1. **Seed a unique command into fish's own history** — in a plain terminal
   (Terminal.app / iTerm, *not* tty7), run it in an **interactive** fish shell
   (note: `fish -c '…'` is non-interactive and does not write history):

   ```sh
   echo tty7_fish_probe_$(date +%s)
 ```

 2. Confirm it reached the history file (fish 4 writes it out promptly):
    ```sh
tail -3 ~/.local/share/fish/fish_history
# → a `- cmd: echo tty7_fish_probe_…` record appears
    ```
 3. Build and run tty7 from this branch (the dev alias uses a throwaway
    config dir so nothing in your real ~/.config/tty7 is touched):
    ```sh
cargo dev
    ```
 4. Search in a fish pane: open a pane running fish, press Ctrl+R at the
    prompt and type tty7_fish_probe:
   - Before this change: the marker never appears — tty7 only read
     .zsh_history / .bash_history, never fish_history.
   - After: the marker shows up, alongside your regular fish history.
 5. Regression: Ctrl+R still lists zsh/bash history as before, and commands
    typed inside tty7 panes are still recorded as usual.

### DIFF
--- a/src/terminal/history.rs
+++ b/src/terminal/history.rs
@@ -106,8 +106,12 @@ pub fn load_with_shell_files(scope: &Scope, shell_files: Vec<Vec<u8>>) -> Histor
     normalize(raw)
 }
 
-pub fn shell_history_names() -> [&'static str; 2] {
-    [".zsh_history", ".bash_history"]
+pub fn shell_history_names() -> [&'static str; 3] {
+    [
+        ".zsh_history",
+        ".bash_history",
+        ".local/share/fish/fish_history",
+    ]
 }
 
 fn looks_absolute(p: &str) -> bool {
@@ -298,6 +302,13 @@ fn load_shell_history() -> Vec<Raw> {
         add(home.join(".zsh_history"));
         add(home.join(".bash_history"));
     }
+    // XDG_DATA_HOME is independent of HOME (and the only reliable signal on
+    // Windows runners, where HOME may be unset); fall back to ~/.local/share.
+    if let Some(xdg) = std::env::var_os("XDG_DATA_HOME").filter(|d| !d.is_empty()) {
+        add(PathBuf::from(xdg).join("fish/fish_history"));
+    } else if let Some(home) = std::env::var_os("HOME") {
+        add(PathBuf::from(home).join(".local/share/fish/fish_history"));
+    }
     files.sort_by_key(|p| {
         std::fs::metadata(p)
             .and_then(|m| m.modified())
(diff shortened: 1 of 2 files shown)


l0ng-ai:
Thanks for this — fish history in the Ctrl+R menu is worth having. I pushed a follow-up commit instead of sending it back, because the format turned out to be a trap.

**`fish_history` looks like YAML and isn't.** fish escapes exactly two characters when it writes a command (`\` → `\\`, newline → `\n`) and quotes nothing, so a real history file is full of records libyaml rejects. Checked against the pinned serde_yaml 0.9:

```
- cmd: git commit -m "fix: crash on start"   -> ScannerError   (dropped)
- cmd: echo foo: bar                         -> ScannerError   (dropped)
- cmd: [ -f x ]; and echo y                  -> ParserError    (dropped)
- cmd: rg --files # every file               -> "rg --files"   (truncated!)
```

A record that fails to parse is a record that vanishes from history search with no diagnostic — so every conventional-commit message, every `echo a: b`, and every `[ ... ]` test was missing, and anything with a ` #` came back silently shortened and would have run a different command on recall. `\n` and `\\` were never reversed either, so multiline and backslash commands came back wrong.

That also means `a_bad_fish_record_does_not_sink_the_rest_of_the_file` was enshrining `- cmd: this: is: broken` as malformed when it's just what fish writes for `echo this: is: broken`, and `fish_history_multiline_cmd_is_joined_with_newlines` passed only because it fed a YAML double-quoted scalar that fish never emits.

It's now a line scanner shaped like fish's own reader: a record starts at `- cmd:` in column 0, keys are indented under it, the rest of the line is the command. That also retires the per-record libyaml parser, which ran on the UI thread at pane construction over an uncapped file — fish keeps up to 256k entries.

**The remote path was reading fish history as bash lines.** `shell_history_names()` feeds the remote fetcher (`view.rs:2788`), but `load_with_shell_files` got bare `Vec<Vec<u8>>` and always called `parse_shell_history`, so a fish user's *remote* pane filled its history menu with literal `- cmd: git status` and `when: 1700000000` rows. Each fetched file now carries the name it came from and one `parse_history_file` picks the reader for both paths.

Two smaller things:

- Multiline commands are skipped rather than half-recalled. Everything downstream is single-line — `append` refuses embedded newlines, `handoff_line_to_shell` bails on them, and the reverse-search menu draws one fixed-height row per entry — so an entry that can't be run is worse than one that isn't offered. Same place bash/zsh multiline entries already land.
- Dropped the test that drove `load_shell_history` by mutating `XDG_DATA_HOME`. `set_var` races every other thread in a parallel test binary (that's why it's `unsafe` in edition 2024), and it cleared the caller's real value on the way out. The path is covered by constants plus a test pinning the XDG-relative and HOME-relative spellings to each other.

Full suite green (929 tests).

dongyuwei:
Thanks for review and follow-up.

PR #422, opened 2026-08-09
fix(wsl): fix WSL pane startup when fish is the default shell

ArnoChenFx:
Fix WSL panes failing immediately when the distribution uses fish or another non-POSIX shell as its default login shell.

tty7 previously launched its WSL bootstrap with:

```text
wsl.exe ... -- sh -c <bootstrap>
```

The `--` argument only ends WSL option parsing. WSL may still pass the remaining command through the distribution’s default shell. When fish is the default shell, fish parses the POSIX parameter expansions in the bootstrap before `sh` receives them and exits with an error such as:

```text
fish: ${ is not a valid variable in fish
```

The bootstrap now uses:

```text
wsl.exe ... --exec sh -c <bootstrap>
```

`--exec` bypasses the distribution’s default shell for the bootstrap command. The explicit `sh` process interprets the POSIX wrapper and then launches the user’s configured shell in login mode.

## Changes

- Replace the WSL bootstrap separator `--` with `--exec`.
- Preserve the selected WSL distribution and initial working directory.
- Preserve the existing `$SHELL`-based final login-shell selection.
- Preserve the Bash-specific rcfile integration path.
- Add a regression assertion requiring `--exec` in generated WSL arguments.
- Update the real PTY test to wait for `OSC 133 D` with the command exit status.
- Avoid treating fish’s initial status-less `OSC 133 D` prompt marker as command completion.

## Compatibility

The change only affects which interpreter parses tty7’s bootstrap script. It does not replace or bypass the user’s final login shell.

The resulting startup flow is:

```text
wsl.exe --exec sh -c <bootstrap>
  └─ user’s configured login shell
```

Common shells continue to use their existing final launch behavior:

- Bash: launched with tty7’s generated rcfile and interactive mode.
- fish: launched in login mode without parsing the POSIX bootstrap.
- zsh: launched in login mode.
- Nushell: launched in login mode.
- Other shells: launched through the existing generic login-shell fallback.
- Missing `$SHELL`: falls back to `/bin/sh -l`.

User-authored non-empty WSL arguments continue to disable automatic shell integration and are unaffected by this change.

### DIFF
--- a/crates/tty7-core/src/daemon/install/wsl.rs
+++ b/crates/tty7-core/src/daemon/install/wsl.rs
@@ -71,7 +71,13 @@ pub fn validate_distro(name: &str) -> Result<(), DistroNameError> {
 }
 
 pub fn wsl_args(distro: &str, argv: &[&str]) -> Vec<String> {
-    let mut args = vec!["-d".to_string(), distro.to_string(), "--".to_string()];
+    let mut args = vec!["-d".to_string(), distro.to_string()];
+    // `--` hands the rest of the line to the distro's default shell, which
+    // re-parses it before the program ever runs — fish chokes on the POSIX
+    // command lines we send, and every shell re-globs the arguments. `--exec`
+    // runs the argv itself. With no argv there is nothing to exec, and `--`
+    // keeps its other meaning there: start the default shell.
+    args.push(if argv.is_empty() { "--" } else { "--exec" }.to_string());
     args.extend(argv.iter().map(|a| (*a).to_string()));
     args
 }
@@ -672,7 +678,7 @@ mod tests {
     fn the_transport_command_line_is_the_designs() {
         assert_eq!(
             wsl_args("Ubuntu", &["tty7-server", "--stdio"]),
-            vec!["-d", "Ubuntu", "--", "tty7-server", "--stdio"]
+            vec!["-d", "Ubuntu", "--exec", "tty7-server", "--stdio"]
         );
         assert_eq!(
             wsl_args(
@@ -685,7 +691,7 @@ mod tests {
             vec![
                 "-d",
                 "Ubuntu-22.04",
-                "--",
+                "--exec",
                 "/home/me/.local/share/tty7/bin/tty7-server-26.7.5",
                 "--stdio",
             ]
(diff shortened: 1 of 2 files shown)


l0ng-ai:
Thanks — the diagnosis is right and I confirmed it locally: `wsl.exe --exec` keeps `$SHELL` and `--cd` intact (WSL's init still sets `SHELL` from `/etc/passwd`), so the bootstrap's `case "$SHELL"` still works after bypassing the default shell.

I pushed a follow-up commit rather than send it back, since two gaps only show up on the same machines this fixes:

**1. The same `--` is still on the transport path.** `install::wsl::wsl_args` — behind `RemoteLink::wsl`, `RemoteLink::wsl_shell` and the installer's `sh -s` / `tee` calls — kept emitting `--`. On a fish-default distro that left WSL *remote workspaces* failing to connect for exactly the reason local panes had just stopped failing. Every argv-carrying form now goes out under `--exec`; `--` keeps its other meaning ("just start the default shell") for the no-argv case.

**2. The fish panes this unblocks came up with no shell integration.** `WSL_EXEC_SCRIPT` only had a `*/bash)` arm, so fish fell to `exec "$SHELL" -l` — no OSC 133, no OSC 7, which means no prompt marks, no exit status, no cwd, and `tty7 wait` / busy-idle status dead in the pane. There's now a `*/fish)` arm carrying `FISH_INTEGRATION` inline, the way `remote::bootstrap_command` already does over SSH.

One subtlety worth recording: over SSH the far end's *own* shell parses the bootstrap, which is why `remote::fish_bootstrap` uses `fish_quote`. Here `sh` parses it, so the body has to be POSIX-quoted — `fish_quote` would double every backslash and hand fish `\\e]%s\\a`. A test pins that distinction, and another asks a real `sh -n` to parse the generated script on CI's Linux and macOS legs.

I verified the dispatch inside a real distro with a shim on `$SHELL`: valid POSIX sh, fish arm gets exactly `-C <2436-byte body> -l` with the body intact and no doubled backslashes, other shells fall through to `$SHELL -l`.

Also anchored the PTY tests' completion marker on its terminator. `contains("133;D;1")` is a prefix match on the exit status, so it also accepted `133;D;127` (bootstrap exec'd a shell with no `false` on PATH) and `133;D;130` (interrupted) — both tests would have gone green on a bootstrap that never ran the command they type.

zsh-default distros still land on the bare login-shell arm. That's unchanged by this branch rather than introduced by it, so I left it alone here.

PR #446, opened 2026-08-09
feat(sidebar): tail-first label elision and a hover card for hidden details

ArnoChenFx:
The sidebar squeezed every label with `truncate()`, which keeps the head and drops the tail — the exact part a path or branch is identified by. `window-transparency-backdrop` read `window-transp...` and deep paths lost their file name, so the limited width was spent on repeated prefixes while the distinguishing suffix vanished. This PR reworks row text around real glyph widths and adds a hover card that surfaces whatever a row had to hide.

## Changes

- **Measured elision** (`src/ui/tab_strip.rs`): `measure_text` shapes text with the window's `TextSystem` so truncation is exact for mixed CJK/Latin labels; `elide_path_keep_tail` drops whole path segments from the front (both `/` and `\` split segments) while keeping the root marker — drive letter, `~`, leading slash — and the leaf segment intact, falling back to character-level tail elision only when the last segment alone overflows; `elide_keep_edges` keeps both ends of a branch (`window-trans…backdrop`), cutting on separators.
- **Row layout** (`src/ui/tab_sidebar.rs`): the title elides tail-first against the row budget (sidebar width minus avatar, padding, badge); the branch is elided to exactly the width the diff counts leave — the counts never wrap or shrink; non-repo tabs now fill their second line with the compressed cwd carrying its root marker instead of leaving it blank.
- **Hover card**: a `Tooltip::element` shows every detail the row hid — the full path (including the `user@host:` prefix `short_title` strips), the full branch with diff counts, the full cwd, and the remote host the avatar only dots. It only appears when the row actually hid something, so an untruncated row stays quiet.
- `short_title` now also splits Windows backslash paths, so deep Windows directories get the same tail-first treatment.

## Testing

- 16 new unit tests pin the elision contract: root marker and leaf segment survival, branch head/tail survival, width-budget compliance, and Windows backslash paths (all fixtures use fictional paths).
- Manually verified at a 185px sidebar: deep paths render `…/tty7/src/ui` with the leaf intact, non-repo rows show the root-marked cwd, and the hover card appears only on rows that hid content.

image

PixPin_2026-08-09_21-21-53

### DIFF
--- a/src/ui/tab_sidebar.rs
+++ b/src/ui/tab_sidebar.rs
@@ -20,8 +20,8 @@ use crate::ui::i18n::{L10nKey, t, t_fmt};
 use crate::ui::reorder::{self, Reorder, Surface};
 use crate::ui::right_panel::RESIZE_HANDLE_WIDTH;
 use crate::ui::tab_strip::{
-    DragTab, REORDER_SLIDE_MS, abbreviate_home, elide_keep_edges, elide_path_keep_tail,
-    measure_text, strip_host_prefix,
+    DragTab, REORDER_SLIDE_MS, abbreviate_home, elide_keep_edges, elide_label,
+    elide_path_keep_tail, measure_text, strip_host_prefix,
 };
 
 const MIN_SIDEBAR_WIDTH: f32 = 180.;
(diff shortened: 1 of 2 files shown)


l0ng-ai:
Reviewed and pushed a follow-up commit to this branch (505670c) rather than sending it back — the diagnosis and the shape of the fix are right, measuring instead of counting characters is the correct call, and the doc comments carry their reasoning well. What follows is what the follow-up changes and why.

## Four gaps between what a row hides and what the card gives back

**A renamed tab could not be expanded.** The row now elides `tab.name` like anything else, but `sidebar_info` still carried the old `tab_title_tooltip` guard and returned early on any non-empty `tab.name`. That guard was correct when `tab_label` returned the custom name verbatim; now the row shortens it and nothing brings it back.

**An empty terminal title opened an empty card.** `leaf_title` returns `String::default()` when there is no terminal leaf. The row correctly falls back to `Shell 3`, but the card compared that placeholder against the empty string it came from, found them different, and rendered a blank first line on a row that had hidden nothing. The old code had an explicit `if raw.is_empty() { return None }` that did not carry over.

Both come from deriving the same strings in two places. The row now hands `sidebar_info` what it rendered next to what it rendered it from, and the card is decided by comparison alone — no second derivation to drift.

**Non-path titles lost their head.** `elide_path_keep_tail` was applied to every title, but a title is not always a path. Measured against the code as it stood:

```
"Backend server logs"    -> "… server logs"
"npm run dev -- --watch" -> "…v -- --watch"
```

For a command line or a name someone typed, the head is what names it — this reads as less than the CSS truncation it replaced. Added `elide_label`, which picks the rule: tail-first for a path, both edges otherwise.

**`elide_keep_edges` kept neither end on a separator-less token.** With no `-_/.` in the first 12 chars the head ran to the cap, `head…` alone overran the budget, and it fell through to a bare tail — at every budget from 40px to 120px:

```
"verylongbranchnamewithoutseps" -> "…eps" / "…utseps" / "…houtseps" / "…ithoutseps"
```

And in the other direction, `feature/some-really-long-thing` at 80px produced `feature/…` — a zero-length tail, the identifying part gone entirely. It now tries shorter heads before giving up and never trades the whole tail for a longer head. Same fixtures after:

```
"verylongbranchnamewithoutseps" @80  -> "ver…tseps"
"feature/some-really-long-thing" @80 -> "fea…thing"
"window-transparency-backdrop" @140  -> "window-…backdrop"   (unchanged)
```

## Smaller ones

- **Budget was ~8px too generous** — the list container's own `px_1` was not subtracted. Labels landing in that band overflowed into the `.truncate()` fallback, which is the exact head-kept/tail-dropped behavior this PR exists to remove. The row metrics are now named constants the layout reads from, instead of five bare literals restating it.
- **The active row renders at `FontWeight::MEDIUM` but was measured as regular** — wider in any proportional face, so the one row the user is looking at was the most likely to overflow.
- **Windows paths flipped separator when elided** — `C:\Users\dev\app` while it fits, `C:/…/app` once it does not. One tab spelling its location two ways depending on sidebar width. Paths are rejoined with the separator they arrived with, in `short_title` too.
- Diff counts measured the gap between them rather than a space standing in for it; the hover card wraps instead of truncating the one string it promised in full; the host is read off the same leaf as the title, so a split tab across two machines names the right one; `tab_label` is only built when there is a query to match it against.

## On the tests

They shape through gpui's `NoopTextSystem`, where every glyph is exactly one em — weight-agnostic and CJK-agnostic. That is good news for stability (identical on all three CI targets, no font flakiness), but it means `measure_text` inside a test *is* the character-counting model this PR replaces, so the headline claim — exact truncation for mixed CJK/Latin — rests on your manual verification, not on CI. Said so where the fixtures are built rather than leaving the comment implying the pixels are real.

Added coverage for the cases above (non-path label, separator-less token, tail-never-traded-away, separator preservation), and gave the tilde/absolute test the same "fixture must exceed the budget" assertion its drive-letter sibling already had. 1166 tests pass, `cargo build` and `cargo fmt --check` clean.

Thanks for this — the sidebar reads a lot better with it.

PR #464, opened 2026-08-10
fix(ui): dim panes by blending terminal colors toward the window background

ArnoChenFx:
Unfocused panes are faded with an element-opacity style (`opacity(0.55)` inactive, `opacity(0.45)` while dragged). Element opacity alpha-multiplies every quad and path **separately**, so stacked decorations dim unevenly: a powerline separator drawn over its segment quad keeps the segment's own dim visible through its `(1 - dim)` alpha and lands with a visible seam against the segment next to it — brighter than the segment on a dark preset (measured at 1.3–1.9× the segment's luminance), darker and more saturated on a light one.

image

## Change

The pane leaf now computes the per-frame `dim` (1.0 / [`INACTIVE_DIM`](src/ui/pane.rs) for an unfocused pane in a split / [`LIFTED_DIM`](src/ui/pane.rs) while dragged) and stores it on `TerminalView`; the terminal element applies it during paint by blending its colors toward the window background:

- **Under color** (`dim_under`): the active preset fill premultiplied by the window's own opacity — the same source `workspace_background` paints with. Solid fills match exactly; `Vertical`/`Horizontal` gradients are approximated by their midpoint stop; a bare theme falls back to the `theme().background` token. Looked up only on dimmed frames, and passed down pre-converted to `Rgba` — it is a frame constant, and every cell blends fg, bg and an optional underline color against it.
- **Per-cell blend** (`dim_cell`): fg, bg, and underline color blended in RGB space, alpha untouched. The blend happens on the *cell* rather than the palette because truecolor cells (direct `38;2;…`/`48;2;…` SGR, e.g. starship's prompt segments) carry their own `Spec` color that a palette-level dim would never see.
- **Paint colors** (`PaintColors::dimmed`): caret, selection/match backgrounds pre-blended; the raw `fg_rgb`/`bg_rgb` stay untouched so named-color cells are dimmed exactly once by the per-cell blend.
- **Bitmaps**: a translucent under-colored overlay quad dims the image as a single layer, matching the surrounding cells.
- **Hand-off**: the element resets `dim` to 1.0 at the end of every paint, so the pane leaf's per-frame write is structural — a render site that doesn't set it (a maximized pane renders without the chrome that normally drives it) paints at full brightness instead of inheriting a stale dim. This replaces the manual reset the maximized path used to carry.
- **Connecting screen / drag grip**: keep the plain opacity style; the grip is wrapped so it is not alpha-multiplied a second time.

Blending is linear in the composited result, so `blend(a) over blend(b)` equals `blend(a over b)` — the composite of stacked layers is exactly as dimmed as any single layer, with no seam.

## Trade-offs

- Gradient fills blend toward the midpoint stop, not the exact color behind each cell.
- On a translucent window, a dimmed pane's painted cells no longer let the desktop show through: they are painted opaque, pre-blended toward `fill × window_opacity`, which ignores the backdrop's own contribution. The old style left them at alpha `dim`.
- Only the terminal element's cells are dimmed; the search bar, completion menu and integration notice render outside the grid and stay at full brightness — the same "the fading is worn by what it holds" rule the drag grip already follows.


l0ng-ai (commented):
I ran this as an A/B: this branch and `main` (`0f5e637`) built as separate binaries, each in its own isolated instance, both showing an identical seeded scrollback — a starship-style powerline bar of truecolor segments joined by `` separators. Both panes unfocused, `dim_inactive_panes` on. Numbers below are sampled per-pixel from the screenshots, not eyeballed.

## The artifact is real, and the model behind the fix predicts it exactly

Separator-cell colour vs. the segment it should be continuous with, in the dimmed pane, **dark preset** (`under` = #000, dim = 0.55):

| junction | segments (lum) | `main` separator | this branch |
|---|---|---|---|
| pink → purple | 75 / 51 | (158,67,104) — lum **98** | (120,54,69) — lum 75, identical to the segment |
| purple → blue | 51 / 52 | (98,54,119) — lum **75** | (85,29,78) — lum 51 |
| blue → green | 52 / 57 | (46,89,104) — lum **78** | (29,56,90) — lum 52 |

On `main` the wedge sits at 1.3–1.9× the luminance of the segments on either side of it. `0.55·pink over 0.55·purple` = (158,67,104) predicts the measured pixel to the digit, so the diagnosis in the description holds. On this branch there is no separator colour at all: a run-length scan of the bar merges the triangle into the preceding segment.

Two things worth recording alongside that:

- **The light preset inverts the direction.** With `under` = #fff the same wedge comes out *darker and more saturated* than its segments — (210,118,156) against (235,169,183) — not brighter. It is also much less noticeable there, which is why this is easy to miss on the default theme.
- **Single-layer output is identical between the two builds** on both presets: every segment fill measured the same value before and after. Nothing that was already correct moved, which is the reassuring half of the result.

Approach looks right to me and I'd like to see it land. Four things first.

## 1. `blend_toward` reconverts `under` on every call

`Rgba::from(under)` runs once per colour per cell, and `under` is a frame constant. That is two wasted HSL↔RGB round trips per cell (fg + bg, more with an underline colour), on a path that reruns for every dimmed pane every frame. Taking `under` pre-converted as `Rgba` removes roughly a third of the added per-cell cost for a one-line change.

## 2. The `dim` hand-off is an unenforced invariant

`TerminalView::dim` is written by mutating the entity during the parent's `render`, and read later in the element's `paint`. It works, but nothing enforces that every render site sets it — the failure mode is a pane that stays dimmed with no compile-time or runtime signal. The maximized path in `app.rs` already had to be patched for exactly this, and that was with only two render sites in the tree. Either reset it to 1.0 once per frame at the root and let the pane override, or add a debug assertion that it was written this frame.

## 3. `dim_under` has no test

The arithmetic helpers are well covered, but the preset mapping itself — gradient midpoint, `opacity` premultiply, the no-preset fallback to `theme().background` — has no coverage, and it is the piece most likely to be silently broken by a future change to `presets::Fill` or `workspace_background`.

## 4. Three corrections for the description

- **"Truecolor cells … never pass through the palette, so a whole prompt stayed at full brightness while indexed content around it dimmed."** `Window::paint_quad` and `Window::paint_path` both multiply by `element_opacity()`, so under the old style truecolor backgrounds dimmed like everything else. Handling them in `dim_cell` is still correct and necessary — the stated symptom just isn't what the old code produced, and the claim will mislead whoever reads this via `git blame`.
- **"landed brighter"** is preset-dependent, per the light-preset measurement above.
- **Missing trade-off:** on a translucent window the dimmed pane no longer lets the desktop show through its painted cells. The old style left them at alpha `dim`; the new one paints opaque colours pre-blended toward `fill × window_opacity`, which ignores the backdrop's own contribution. I think that's a fair price for seam-free stacking, but it belongs in the description next to the gradient-midpoint approximation that is already called out.

One more behaviour change that isn't mentioned anywhere: the opacity style used to wrap the whole `TerminalView`, so the search bar, completion menu and integration notice faded with the pane. Now only the terminal element's cells are dimmed and those overlays stay at full brightness. That matches the "whatever fading the pane is under is worn by what it holds" rule the grip already follows, so I assume it's intended — worth a line either way.

ArnoChenFx:
> I ran this as an A/B: this branch and `main` (`0f5e637`) built as separate binaries, each in its own isolated instance, both showing an identical seeded scrollback — a starship-style powerline bar of truecolor segments joined by `` separators. Both panes unfocused, `dim_inactive_panes` on. Numbers below are sampled per-pixel from the screenshots, not eyeballed.
> 
> ## The artifact is real, and the model behind the fix predicts it exactly
> Separator-cell colour vs. the segment it should be continuous with, in the dimmed pane, **dark preset** (`under` = #000, dim = 0.55):
> 
> junction	segments (lum)	`main` separator	this branch
> pink → purple	75 / 51	(158,67,104) — lum **98**	(120,54,69) — lum 75, identical to the segment
> purple → blue	51 / 52	(98,54,119) — lum **75**	(85,29,78) — lum 51
> blue → green	52 / 57	(46,89,104) — lum **78**	(29,56,90) — lum 52
> On `main` the wedge sits at 1.3–1.9× the luminance of the segments on either side of it. `0.55·pink over 0.55·purple` = (158,67,104) predicts the measured pixel to the digit, so the diagnosis in the description holds. On this branch there is no separator colour at all: a run-length scan of the bar merges the triangle into the preceding segment.
> 
> Two things worth recording alongside that:
> 
> * **The light preset inverts the direction.** With `under` = #fff the same wedge comes out _darker and more saturated_ than its segments — (210,118,156) against (235,169,183) — not brighter. It is also much less noticeable there, which is why this is easy to miss on the default theme.
> * **Single-layer output is identical between the two builds** on both presets: every segment fill measured the same value before and after. Nothing that was already correct moved, which is the reassuring half of the result.
> 
> Approach looks right to me and I'd like to see it land. Four things first.
> 
> ## 1. `blend_toward` reconverts `under` on every call
> `Rgba::from(under)` runs once per colour per cell, and `under` is a frame constant. That is two wasted HSL↔RGB round trips per cell (fg + bg, more with an underline colour), on a path that reruns for every dimmed pane every frame. Taking `under` pre-converted as `Rgba` removes roughly a third of the added per-cell cost for a one-line change.
> 
> ## 2. The `dim` hand-off is an unenforced invariant
> `TerminalView::dim` is written by mutating the entity during the parent's `render`, and read later in the element's `paint`. It works, but nothing enforces that every render site sets it — the failure mode is a pane that stays dimmed with no compile-time or runtime signal. The maximized path in `app.rs` already had to be patched for exactly this, and that was with only two render sites in the tree. Either reset it to 1.0 once per frame at the root and let the pane override, or add a debug assertion that it was written this frame.
> 
> ## 3. `dim_under` has no test
> The arithmetic helpers are well covered, but the preset mapping itself — gradient midpoint, `opacity` premultiply, the no-preset fallback to `theme().background` — has no coverage, and it is the piece most likely to be silently broken by a future change to `presets::Fill` or `workspace_background`.
> 
> ## 4. Three corrections for the description
> * **"Truecolor cells … never pass through the palette, so a whole prompt stayed at full brightness while indexed content around it dimmed."** `Window::paint_quad` and `Window::paint_path` both multiply by `element_opacity()`, so under the old style truecolor backgrounds dimmed like everything else. Handling them in `dim_cell` is still correct and necessary — the stated symptom just isn't what the old code produced, and the claim will mislead whoever reads this via `git blame`.
> * **"landed brighter"** is preset-dependent, per the light-preset measurement above.
> * **Missing trade-off:** on a translucent window the dimmed pane no longer lets the desktop show through its painted cells. The old style left them at alpha `dim`; the new one paints opaque colours pre-blended toward `fill × window_opacity`, which ignores the backdrop's own contribution. I think that's a fair price for seam-free stacking, but it belongs in the description next to the gradient-midpoint approximation that is already called out.
> 
> One more behaviour change that isn't mentioned anywhere: the opacity style used to wrap the whole `TerminalView`, so the search bar, completion menu and integration notice faded with the pane. Now only the terminal element's cells are dimmed and those overlays stay at full brightness. That matches the "whatever fading the pane is under is worn by what it holds" rule the grip already follows, so I assume it's intended — worth a line either way.

Fixed

l0ng-ai (approved):
Verified `25e7dd9c` point by point — all four land, and two go further than asked:

1. **Pre-converted `under`** — the whole under pipeline is `Rgba` now, not just the `blend_toward` parameter: `premultiplied`/`fill_under`/`theme_under` never round-trip through HSL at all, and the tests tightened from `1e-3` to `1e-6` as a side effect. Better than what I asked for.
2. **Hand-off** — resetting to 1.0 at the end of every paint makes the invariant structural, and deleting the manual maximized-path reset in `app.rs` is the right proof that it works: the case that motivated the concern now falls out of the design instead of being patched around.
3. **`dim_under` coverage** — splitting the pure mapping out into `fill_under`/`theme_under` and testing those directly covers the midpoint, premultiply and fallback arms without needing a `cx`.
4. **Wording** — the code and test comments are accurate now (the truecolor claim as a counterfactual about palette-level dimming, "seam" with the palette-dependence noted), and the two trade-offs are recorded on `dim_under`.

I took the liberty of updating the PR description to match (point 4 applied to the body, plus the now-stale "maximized pane resets to 1.0" bullet replaced by the structural reset) — squash merge turns it into the commit message, so it needed the same corrections. CI is green on the head. Thanks for the careful follow-through — merging.

PR #481, opened 2026-08-11
feat(agent-hooks): report the opencode session id

cwatanab:
## Summary

The opencode plugin now captures `properties.sessionID` from events and forwards it to `agent-hook opencode <event>` on stdin as JSON. It also maps `session.status` (busy/idle) events.

## Motivation

Let tty7 recognize the opencode session running in a pane so it can resume it with `opencode --session <id>` after a restart.

## Changes

- `emit()` sends the session id on stdin as a JSON payload (Bun Shell `< ${Response}` redirect)
- `session-start` is emitted on the first event that names a session id (`announced` prevents re-announcing the same id)
- Added `session.status.busy → prompt-submit` / `session.status.idle → stop` so completion detection works even on opencode versions without `session.idle`
- Fallback capture from `input.sessionID` in `tool.execute.before`

## Verification

- `cargo test -p tty7-core` (961 passed, aside from a known flake)
- `cargo test -p tty7-server -p tty7-cli` all pass
- `cargo clippy -p tty7-core --tests` no new warnings
- Ran the generated plugin against a stub emitter; every event carried `{"session_id":"..."}` on stdin

### DIFF
--- a/crates/tty7-core/src/core/agent_hooks.rs
+++ b/crates/tty7-core/src/core/agent_hooks.rs
@@ -924,6 +924,12 @@ export const Tty7Presence = async ({{ $ }}) => {{
   const cmd = {prefix}
   let sessionId = ""
   let announced = ""
+  // A subagent runs in a child session on the *same* event stream, and its
+  // status events are indistinguishable from the pane's own — only
+  // `session.created`/`session.updated` name a parent. Remember the children,
+  // so a task tool cannot hand the pane the wrong session to resume, nor call
+  // the pane done when only the subagent is.
+  const children = new Set()
   const emit = (event) => {{
     // Every event carries the session id (`properties.sessionID`), so a pane
     // that restarts can resume the same session with `opencode --session`.
@@ -935,7 +941,7 @@ export const Tty7Presence = async ({{ $ }}) => {{
   // `session.created` event); the session-start report rides on the first
   // event that names one, so a restored pane can reattach to it.
   const capture = async (id) => {{
-    if (id) sessionId = id
+    if (id && !children.has(id)) sessionId = id
     if (sessionId && sessionId !== announced) {{
       announced = sessionId
       await emit("session-start")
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Reviewed — thanks, this is the right half of the feature to fill in. The Rust side already lifts `session_id` off the hook's stdin (`build_hook_sequence`) and `CLIAgent::OpenCode` already builds `opencode --session <id>`, so the plugin was the only thing not reporting it. I checked the two assumptions the generated JS rests on and both hold: Bun Shell accepts a `Response` as a `<` stdin redirect, and opencode's `session.status` carries `properties.sessionID` with `properties.status.type` of busy/idle.

One thing I pushed a fix for in cd3f8bd. A task-tool subagent runs in a **child session** whose events ride the same plugin stream and are structurally indistinguishable from the pane's own (upstream [opencode#30043](https://github.com/anomalyco/opencode/issues/30043)). So `capture(properties.sessionID)` latches whichever session spoke last. Driving the generated plugin against a stub `$` through "main session works → spawns a subagent → subagent finishes → main session finishes":

```
session-start  {"session_id":"ses_main"}
prompt-submit  {"session_id":"ses_main"}
session-start  {"session_id":"ses_child"}   <- the pane now reports the subagent
prompt-submit  {"session_id":"ses_child"}
stop           {"session_id":"ses_child"}   <- pane called done, main agent still working
stop           {"session_id":"ses_child"}
```

The pane would have stored `ses_child` as its session, so a restart resumes the subagent's conversation rather than the user's, and the tab flips to done mid-run.

The fix remembers child ids from `session.created`/`session.updated` (only those name a `parentID`) and lets their events pass without touching the pane's session. Same scenario after:

```
session-start    {"session_id":"ses_main"}
prompt-submit    {"session_id":"ses_main"}   x3 (the subagent's tool calls still count as work)
permission-request {"session_id":"ses_main"}
stop             {"session_id":"ses_main"}
session-end      {"session_id":"ses_main"}
```

Also passed `input.sessionID` into `capture()` from `permission.ask`, which the plugin API does hand over.

Note the load-time `session-start` is gone with this change — presence now starts at the first event that names a session. Process detection already tells tty7 the pane runs opencode, so nothing user-visible hangs on it.

PR #512, opened 2026-08-11
fix(files): replace what a drop lands on only once the copy is whole (#490)

### LINKED ISSUE #490: [文件拖放] 同批次拖入的同名文件互相静默覆盖；覆盖逻辑是"先删后拷"，拷贝失败则新旧俱失
**类型**：Bug　**平台**：全平台　**版本**：main @ 49bfe59（26.8.2），源码诊断

## 实际表现

把 `/from/a/readme.txt` 和 `/from/b/readme.txt` 一起拖进文件树的同一个目录：

1. 不弹任何冲突提示（两个源在目标目录都不存在，检测不到它们彼此同名）；
2. 两个文件依次写入同一路径，**后一个静默覆盖前一个**——用户以为拷了两个文件，实际只剩一个，且无任何提示。

更严重的连带问题：即便是正常的目标已存在冲突，用户确认"覆盖"后，实现是**先删除目标再拷贝**。一旦拷贝中途失败（断链、磁盘满），目标位置原有的文件已经没了，新文件也没拷成——覆盖操作把"保留旧文件"这个最低保障都丢了。

## 代码原理

- `src/ui/file_copy.rs:91-94`：冲突检测只检查"目标主机上是否已存在该路径"（`host.exists(&host.join(dir, &name))`），不检查**本批次拖放源之间的 basename 碰撞**；两个同名源都被放入 `planned`，目标路径相同。
- `file_copy.rs:101-114`：执行循环中，目标已存在时先 `host.remove(&dest, true)`（105-110 行），再 `copy_tree(...)`；copy 失败仅记录 `report.errors`，此时旧文件已被删除。

## 可行方案

1. 规划阶段把"批次内同名"也纳入冲突列表（`report.conflicts`），走既有的冲突确认流程，让用户选择跳过/重命名/覆盖。
2. 覆盖语义改为安全顺序：先拷到目标旁的临时名（如 `.tty7-partial-<name>`），成功后再替换目标；失败则清理临时文件、保留原文件。远程主机上 rename 替换即可，本地同理。
3. 测试：纯函数层面构造同名拖放批次，断言冲突被报告且不发生静默覆盖；模拟 copy 失败断言原文件保留。

如方向认可，愿意提交 PR。

biztex:
Closes #490.

Picked up from the tracking issue's priority list (#509): "唯一两条会毁数据 / 毁配置的" — this is the drop half.

## What was wrong

**Replacing destroyed first and copied second.** `copy_into_dir` did
`host.remove(&dest, true)` and only then walked the tree into place. A copy that
fails partway — a full disk, a control connection that drops mid-tree on a
remote workspace — left the destination holding neither the old thing nor a
whole new one. The failure notification then pointed at a hole where the file
used to be.

**Two dropped items of the same name landed on top of each other.** Conflict
detection only asked whether the *destination* already had the name, so
`~/a/notes.md` and `~/b/notes.md` dragged in together were both planned onto the
same path. The second overwrote the first, and the panel reported both as
copied — one file gone, nothing said.

## What this does

*Succeed, then replace.* The copy lands on `.tty7-partial-<name>` beside the
destination; what is there is moved to `.tty7-replaced-<name>` rather than
removed; a rename puts the new copy in place; the old one is removed only once
the new one is there. Any failure puts the old thing back, and in the one case
where even that fails it stays on disk under the name it was moved to instead of
not at all. The working names are probed rather than reused, so a leftover from a
copy that was killed outright — or a file of somebody's own by that name — is not
stepped on.

Only the path that has something to protect stages; a copy onto a free name is
unchanged, and the extra cost of a replacement is two `exists` probes and two
renames.

*One name is one file.* A source whose destination path another source in the
same drop already claimed is refused with a reason instead of being written over
it. The first claim stands.

## Notes on two judgement calls

- **Refused, not asked about.** #490 suggested routing batch-internal collisions
  into the existing conflict prompt. That prompt asks "Replace?", and replacing
  cannot produce both files — a "yes" would still silently lose one. So the
  duplicate is an error with a message rather than a question whose answers are
  both wrong. Happy to switch to a skip/rename/replace prompt if you would rather
  have one, but that is a bigger piece of UI than this bug needs.
- **Changelog, not a Fixed line.** `file_copy.rs` landed in #458, after v26.8.2,
  so the drop is still under `[Unreleased]`. Its entry now describes the
  behaviour it actually has rather than announcing a fix for a bug no release
  shipped.

## Tests

Three added to `file_copy`'s existing suite, two of which fail on the code they
replace:

- `two_dropped_items_of_the_same_name_do_not_land_on_top_of_each_other` — before:
  `copied == ["notes.md", "notes.md"]` with one file on disk.
- `a_replacement_that_fails_partway_leaves_what_was_there` — before: the
  destination's `keep.txt` is gone.
- `a_finished_replacement_leaves_no_working_files_behind` — guards the cleanup on
  the happy path.

New strings `FileDropNameTaken` and `FileDropNoWorkingName` are translated in
en/zh/ja, so `every_key_is_translated_in_every_locale` stays green.

**One thing I could not run:** I have no macOS/Windows machine, and my Linux box
is missing the system libraries gpui links (fontconfig, xkbcommon, wayland), so
`cargo test -p tty7` would not build here. I ran the two modules this touches —
both gpui-free — out of the tree in a scratch crate against the real
`LocalHost`: 19 tests, all green, including the i18n coverage tests. The logic is
`Host`-trait only, so the remote path is the same code, but it has not been
exercised against a live remote workspace. Worth a look on CI.

---

<sub>中文摘要：拖放覆盖原先是「先删后拷」，拷贝中途失败就新旧俱失；现在改为拷到
`.tty7-partial-<name>`、旧文件先改名到 `.tty7-replaced-<name>`、rename 就位后才删旧，
任何一步失败都会把旧文件放回去。同一批拖入的同名文件此前会静默互相覆盖且两个都报告为
「已复制」，现在第一个生效、其余带原因拒绝。新增 3 个测试，其中 2 个在旧代码上失败。</sub>


l0ng-ai (commented):
Thanks — this is the right shape, and better than the version I had written for the same issue on `fix/drop-forward-safety`, which I have since dropped in favour of this PR.

Two things it does that mine did not:

- **The destination is never the thing that is missing.** Mine staged the copy and then did `remove(dest)` → `rename`, leaving a window where the name did not exist at all. Moving the old entry aside and deleting it only after the swap lands is strictly better, and the fallback — if even the put-back fails, the old thing is still on disk under a findable name — is the right call.
- **Probing the working name instead of clearing it.** Mine did `if exists(temp) { remove(temp) }`, which would have deleted a file of the user's own that happened to be called `.tty7-partial-notes.md`. `free_name_beside` avoids that.

Agreed on both judgement calls, for what it is worth: an error rather than a "Replace?" question is right (the prompt's only two answers both lose a file), and only staging when there is something to protect keeps the common path free.

## One thing to fix before this goes in

`report.conflicts` is filled whether or not `overwrite` is set, and `file_tree.rs:1280` prompts on any non-empty `conflicts` — so the pass that runs *after* the user answers "Replace" reports the same conflict again and re-opens the dialog. Confirmed with a throwaway test on this branch:

```rust
let report = copy_into_dir(&*LocalHost::shared(), &[src], &dest_dir, true);
assert!(report.conflicts.is_empty());
// an overwrite pass still reported conflicts: ["note.txt"]
```

This predates the PR — it is on `main` today — but it matters here specifically, because the callback is an `else if`:

```rust
if !report.conflicts.is_empty() {
    app.file_tree_confirm_replace(...);
} else if let Some((name, e)) = report.errors.first() {
    ...
}
```

So during a replace, `report.errors` is never shown. That hides **this PR's own** `FileDropNameTaken` refusal whenever the same drop also lands on an existing name, and it hides a failed replacement — the case the staging work exists to make survivable. The user would see the Replace dialog a second time and no error at all.

One line, in the function you are already touching:

```rust
if !overwrite && host.exists(&dest) {
    report.conflicts.push(name.clone());
}
```

`conflicts` has exactly one consumer (`file_tree.rs:1281`), so nothing else depends on it being populated on the overwrite pass. Worth a test asserting that a replace-pass report carries no conflicts, next to the ones you added.

## Two smaller notes, neither blocking

- Between `rename(dest, aside)` and `rename(staged, dest)` on a **remote** host, each rename is a control round trip, so a link that drops in between leaves the destination absent and the old entry at `.tty7-replaced-<name>`. Your comment already owns this and it is the right trade. But the only trace is a `log::warn!` — consider putting the aside name into the error the panel reports, so the user is told where their file went rather than having to find it.
- `free_name_beside` probes then uses, so it is technically a TOCTOU. Fine for this context; not worth hardening.

Once the `conflicts` gate is in, this is good to merge and I will close mine out against it.

biztex:
Fixed in cbef3a7 — thanks, that one is nastier than it looks.

I reproduced it before touching anything, by splicing the new test onto the previous revision: `the_pass_that_carries_the_answer_does_not_ask_again` walks the real sequence — ask, answer, replace — and fails on the old code with `conflicts == ["note.txt"]` after the replace pass. So the dialog re-opens over a replacement that already happened, and because of the `else if` the same pass swallows every error it produced.

`if !overwrite && host.exists(&dest)` is in, in the one place that fills the field. Worth noting the field's own doc already promised exactly this — *"Non-empty only when the caller asked without `overwrite`"* — so this is the code catching up with its documented contract rather than a new rule. That does leave the early return above it (`if !overwrite && !report.conflicts.is_empty()`) redundant in its `!overwrite` half; I left it rather than widen the diff, but say the word and it becomes `if !report.conflicts.is_empty()`.

**The aside name in the error** — done, it was too good a point to leave. When the put-back also fails, what the panel reports now names where the old thing went:

> Could not copy note.txt: The copy could not be put in place; what was there is now named ".tty7-replaced-note.txt" in the same folder.

`free_name_beside` hands back the bare name alongside the path for it, because that sentence wants a name and not somebody else's whole remote path. New key `FileDropLeftAside`, translated in en/zh/ja; the `log::warn!` stays for the detail that does not belong on screen.

**The TOCTOU** — agreed, left alone.

No changelog change for the re-prompt: `file_copy.rs` has not been in a release, so the feature's `[Unreleased]` entry describing the behaviour it now actually has is the whole story.

Local runs are 20/20 across the two modules this touches — pulled out of the tree into a scratch crate against the real `LocalHost`, since gpui will not link on this box (no fontconfig/xkbcommon/wayland, no sudo). CI has the real build on all three platforms.

l0ng-ai (approved):
`cbef3a7` looks right, and reproducing it on the old revision first was the right instinct — `the_pass_that_carries_the_answer_does_not_ask_again` walks the actual two-call sequence, which is what makes it a test of the bug rather than of the fix. You are right that the field's doc already promised this; I missed that when I wrote the review, which makes it a better change than I asked for.

Two answers:

- **Leave the early return as `if !overwrite && !report.conflicts.is_empty()`.** It is redundant now, but it states the condition where it is read rather than relying on an invariant established twenty lines up, and the cost is a bool. If the guard above ever grows a third case, this one still says what it means.
- **The aside name in the error was worth doing** — and moving `remove(&staged)` above the put-back is the right order, not just a tidy-up: it frees the name before anything else can want it.

I updated the branch onto current `main` before merging, because your CI had run against a base ten commits old — `main` has taken the other eighteen issues from the #486–#504 batch since you opened this. All green on the fresh base.

On the local build: nothing here needed you to run gpui. The two modules you touched are pure and the CI matrix covers the rest, which is exactly how this change should have been verifiable — that it was is a property of where you put the logic.

Merging. Thanks — this landed better than the version I had written for the same issue, and I closed mine out against yours.

PR #532, opened 2026-08-12
Run the test suite under cargo-nextest for per-test timeouts and isolation

l0ng-ai:
Runs the test suite under `cargo-nextest`, so a hung test can be timed out and *named* instead of taking the job down silently, and fixes the 29 tests that only passed because they shared a process with a sibling that had done their setup.

**This is not a speedup.** It was opened as one; the measurement says otherwise, and the numbers below are what the branch actually does.

## What changes

| | Before | After |
|---|---|---|
| Test runner | `cargo test` — 16 binaries strictly sequential | `cargo nextest run --profile ci` — one pool across all binaries |
| A hung test | Step timeout kills the process tree at 20 min, log says nothing | The individual test is terminated at 3 min and named `TIMEOUT`; the rest of the suite finishes |
| A failing test | `fail-fast` stops the run | `fail-fast = false` — the complete list of what broke, in one run |
| Test isolation | One process, shared globals | One process per test |
| Tests runnable alone | 29 of them are not | All of them are |
| `tty7-cli::cli_e2e` | Inside the same `cargo test` | Its own `CLI end-to-end` step, still `cargo test` |
| Doctests | Ran (there are none) | Own Linux-only step, so a future doctest can't be silently retired |

Job names are untouched, so `rustfmt` and the three `build & test (<target>)` required checks still resolve.

## Measured cost

Warm-cache run [31551827957](https://github.com/l0ng-ai/tty7/actions/runs/31551827957) against warm-cache baseline [31499772887](https://github.com/l0ng-ai/tty7/actions/runs/31499772887):

| | Test execution before | after | Job total before | after |
|---|---|---|---|---|
| Windows | 57s | **48s** | 390s | **382s** |
| macOS | 35s | **52s** | 207s | **231s** |
| Linux | 29s | **33s** | 162s | **185s** |

nextest wins on Windows, where process startup was expensive, and loses on macOS and Linux, where spawning 2700 processes costs more than the sequential-binary startup it replaces. **Total CI wall clock is unchanged** — Windows is still the critical path at ~6m20s, and the two jobs that got slower are nowhere near it.

The original premise was wrong and is worth recording so nobody re-derives it: on Windows the per-binary times `cargo test` reported added up to 59s against 98s of wall clock, and that 39s gap looked like process-startup overhead. It is not — it is `cli_e2e`, which is `harness = false` and therefore prints no `test result:` line to be counted.

`CARGO_PROFILE_DEV_DEBUG=line-tables-only` is kept but did not pay off either: Windows test-target compile went 86s → 77s, within noise, and the Linux and macOS `Build` steps did not move at all.

## What this is actually worth

The suite hangs about one run in ten, on any branch. `cargo test` has no per-test timeout, so the job died at the step's 20-minute mark having reported nothing about which test was stuck — one such run held a runner slot for six hours. An earlier post-mortem step could not fix it: GitHub tears the step's process tree down before the next step runs. nextest terminates the individual test and names it.

And process isolation exposed 29 tests that were only passing by accident:

- **28** reached `Config::load`/`save` without pinning a scratch config dir, living off another test's `pin_test_config_dir()`. Each fails on `main` today under `cargo test -- --exact <name>`.
- **`a_seeded_detail_only_asks_for_the_files`** made a Host call on the UI thread. `host::guard_off_ui` is a `debug_assert` behind a process-wide `OnceLock<ThreadId>`, so it only ever judged whichever gpui test registered first, and passed vacuously for every later one. The seed load moves to a worker thread — where the real caller reaches it anyway.

## Serialization

Nine `static Mutex` test guards go quiet under process isolation. Seven guard process-global state that separate processes no longer share, so isolation is strictly stronger than the lock:

| Guard | Protects | Needs a group? |
|---|---|---|
| `lock_config_file`, `lock_session_file`, `pin_dir` | `set_config_dir`, over a scratch dir already keyed by pid | No |
| `responder_lock` | `AUTH_RESPONDER`, a `static OnceLock` | No |
| `claim_mailbox` | `AUTH_MAILBOX`, a `static Mutex<Vec<_>>` | No |
| `serial()` ×3 (scm::panel, scm::graph, file_tree) | Render-idle frame accounting | **Yes** — needs a quiet machine, not mutual exclusion |
| windows_env `SCRATCH` | One value name under `HKCU\Environment` | **Yes** — HKCU belongs to the user, not the process |

The mutexes all stay: they are what keeps `cargo test` correct, and one target still runs that way.

`tty7-cli::cli_e2e` is excluded from nextest and keeps its own step — it is `harness = false` by design, re-execing itself as the daemon under test and as a raw-mode TUI fixture, and nextest can only enumerate a libtest harness. That step must not name the package: `-p tty7-cli` resolves features for its graph alone while the rest of the job builds the default members together, and the two feature sets are two builds of `tty7-core`. Naming it rebuilt 91 crates — 1m03s on Linux, 2m12s on Windows — which the second commit fixes.

## Testing

- Full suite under nextest locally, three consecutive runs: `2732 tests run: 2732 passed, 5 skipped`, ~10.1s each, no flakes.
- CI green on all three targets, cold and warm.
- Same tests still pass under plain `cargo test` — the pins are additive, nothing left the default runner except cli_e2e.
- Verified the config-dir failures are pre-existing: `cargo test --bin tty7-app -- --exact terminal::remote::tests::write_sends_input_frames` fails on `main`.
- `nextest show-config test-groups` confirms the `render-idle` filter catches exactly the 20 tests that took the three `serial()` locks.
- The `-p` fix measured against a clean target directory: 91 crates compiled with it, zero without, same 13 tests either way.

**Not verified:** the `windows-registry` group matches nothing on macOS (those tests are `#[cfg(windows)]`) and a filter matching nothing is not an error, so the Windows job is its first real exercise.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -14,8 +14,9 @@
 # Excluding it here rather than teaching it the libtest protocol is the right
 # trade: converting it would mean giving up the shared daemon and the self-exec,
 # which is the whole design of the file. CI keeps running it, under plain
-# `cargo test`, in its own step — see the `CLI end-to-end` step in ci.yml. Keep
-# the two in sync: dropping the step would silently retire thirteen tests.
+# `cargo test`, in its own step — see the `CLI end-to-end` step in ci.yml, and
+# note the comment there about why that step must not name the package. Keep the
+# two in sync: dropping the step would silently retire thirteen tests.
 #
 # This lives on `default` rather than on `ci` so a local `cargo nextest run`
 # behaves the same way instead of erroring out.
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,104 @@
+# cargo-nextest's configuration. `.github/workflows/ci.yml` selects `--profile
+# ci`; a bare `cargo nextest run` locally uses `default` and inherits everything
+# not overridden below.
+
+[profile.default]
+# `tty7-cli`'s cli_e2e is `harness = false` (see crates/tty7-cli/Cargo.toml), and
+# deliberately so: the test binary *is* `fn main()`, and it re-execs itself as
+# the daemon under test (`TTY7_CLI_E2E_DAEMON=1`) and as a raw-mode TUI fixture
+# (`--tty7-e2e-paste-aware-fixture`), then runs its cases in order against the
+# one daemon it started. nextest can only drive a libtest harness — it enumerates
+# tests with `--list`, which that `main` ignores, so leaving it in fails the run
+# before a single test executes.
+#
+# Excluding it here rather than teaching it the libtest protocol is the right
+# trade: converting it would mean giving up the shared daemon and the self-exec,
+# which is the whole design of the file. CI keeps running it, under plain
+# `cargo test`, in its own step — see the `CLI end-to-end` step in ci.yml. Keep
+# the two in sync: dropping the step would silently retire thirteen tests.
+#
+# This lives on `default` rather than on `ci` so a local `cargo nextest run`
+# behaves the same way instead of erroring out.
+default-filter = "not binary_id(tty7-cli::cli_e2e)"
+
+[profile.ci]
+# One failure must not hide the rest. On CI the value of a run is the complete
+# list of what broke, and re-running to discover the second failure costs another
+# six minutes of runner time.
+fail-fast = false
+
+# This is as much the reason for using nextest as the speed is.
+#
+# The suite intermittently hangs — roughly one run in ten, on any branch, while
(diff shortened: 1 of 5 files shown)


l0ng-ai:
Closing: the speedup this was opened for does not exist. Measured warm-cache, nextest wins 9s on Windows and loses 17s on macOS and 4s on Linux, for no change in total CI wall clock. The original premise — a 39s process-startup gap on Windows — was a misreading: that gap is cli_e2e, which is `harness = false` and prints no `test result:` line to be counted.

PR #535, opened 2026-08-12
fix(wsl): stop handing bash an rcfile the distro cannot read

biztex:
> **Stacked on #533.** Its commit shows in this diff until that one merges; the fix here is the second commit, `a9e48c7`. Happy to rebase whenever #533 lands, or to fold this into it if you would rather have one.

Follow-up to the hole I named in #533 rather than fix there, since this one changes released behaviour.

## The bug

The WSL bootstrap writes the rcfile on the Windows side and names it to the distro over `/mnt`. That path is not a given — automount off, a root moved by `wsl.conf`, a distro with no drvfs at all — and the bash arm passed it straight to `--rcfile`:

```sh
*/bash) exec "$SHELL" --rcfile "$TTY7_RC" -i ;;
```

**bash ignores an unreadable `--rcfile` in silence.** And tty7 starts it non-login on purpose, so that `--rcfile` is honoured at all (the comment at `bash_rcfile()` says so), which means there is no `.bash_profile` pass to fall back on either. A pane on such a distro opened a shell that had read **no startup file of any kind** — not tty7's integration, which was the optional part, but the user's own `.bashrc`. Every alias, function and prompt they wrote, quietly absent, in a pane that otherwise looks fine.

## The fix

```sh
*/bash) [ -r "${TTY7_RC:-}" ] && exec "$SHELL" --rcfile "$TTY7_RC" -i; exec "$SHELL" -l ;;
```

Readable → unchanged. Not readable → a plain login shell, which reads everything the user wrote and gives up only the integration. Same shape as the zsh arm in #533, and the better of the two failures.

Demonstrated against a stub `bash` on `$SHELL`, with `TTY7_RC` naming a file that is not there:

```
before:  bash --rcfile /tmp/nope -i
after:   bash -l
```

## Tests

`cargo test -p tty7-core` — 1094 passing. The dispatch test (real `sh`, stub shells) grows a case asserting that an unreadable `TTY7_RC` execs `bash -l` and never mentions `--rcfile`; the arm's pinned text is updated alongside.

Unverified on a real distro with automount off — no Windows machine here. The behaviour is proven at the `sh` level, which is where the decision is made.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A WSL distro that cannot reach `/mnt` keeps its own `.bashrc`** — the
+  bootstrap handed bash a `--rcfile` living on the Windows side without
+  checking the distro could read it. bash ignores an unreadable `--rcfile`
+  without a word, and tty7 starts it non-login precisely so that `--rcfile` is
+  honoured at all — so a distro with automount off, or a root moved by
+  `wsl.conf`, opened a shell that had read no startup file the user wrote: not
+  tty7's, and not their `.bashrc` or `.bash_profile` either. The arm now checks
+  first and falls back to a plain login shell, which reads the same startup
+  files a pane without shell integration reads and only gives up the
+  integration.
+
 - **An idle window stops re-reading the repository on Linux** — with a
   repository open, tty7 ran `git status -uall` about two and a half times a
   second, forever, with nothing on screen changing and nobody touching the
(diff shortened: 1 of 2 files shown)


biztex:
The macOS leg went red on the first run here and is green now on an identical tree (force-pushed the same content to re-trigger, no edits). Leaving a note since the failed run stays in the check history.

The failure was `daemon::server::tests::conn::a_closed_writer_channel_wakes_the_controller_parked_in_read` — nothing this PR reaches; the diff is a string in the WSL bootstrap plus two assertions. #533 carries the same shell-integration code and passed macOS on its first run, and I checked the last five failed runs on the repo, none of which mention this test.

One observation, offered as a lead rather than a claim, since I have no macOS machine to test it on: the wake in that path is `write_stream.shutdown(Shutdown::Read)` (server.rs:1218). On Linux `shutdown(SHUT_RD)` reliably returns EOF to a reader already parked in `read()`; on macOS/BSD that has historically been less dependable for a reader parked in the kernel at the moment of the call. That shape would fit — macOS-only, timing-dependent, and rare — and if so the 30s deadline is not what is marginal. If it never recurs, ignore this; if it does, `Shutdown::Both` there is the cheap thing to try first.

PR #610, opened 2026-08-13
fix(ui): stop a rebuilt window renaming its workspace to a codename (#604)

### LINKED ISSUE #604: The workspace switcher shows the directory until the window learns its workspace's name
The chip at the top of the sidebar is the workspace switcher, so it must show the **workspace's name** — the one in the machine tree, which is what `tty7 ws ls` prints and what name-addressing resolves against. Today it shows the directory the first shell happened to start in, and then silently switches to the real name later.

**What happens**

1. Launch tty7 in a directory. The tree already holds a generated name for the new workspace (`machine.json` → `name: "pale-otter"`), but the chip reads `verify-main`.
2. Anything that pulls the whole tree — a reconnect, a relaunch, a rebuild after the daemon is killed — and the chip switches to `pale-otter` and stays there. `views.json` grows `"label": "pale-otter"`.

**Why**

`Tree::mutate(origin, …)` excludes the origin subscriber from its own deltas, so the window that asked for the workspace never receives the `WorkspaceCreated` delta carrying the name. Its mirror files the workspace unnamed, and the display falls back to the directory. The fallback stops applying the moment the mirror learns the name by any other route.

**What it should do**

Show the workspace's name from the first frame, and never change it under the user. The invariant to land: *the window that created a workspace knows its name as soon as it exists.* An explicit rename must still win; a workspace with no name at all still needs a sensible fallback.

**Why it matters beyond the flicker**: while the chip shows `verify-main`, the CLI does not know that name. `tty7 ws tree verify-main` answers "no workspace named 'verify-main'" because `resolve::workspace` matches `ws.name` only. The GUI and the CLI disagree about what the workspace is called, and the user is reading the wrong half off the screen.

---

*Note: this issue was originally filed with the diagnosis backwards — treating the directory name as correct and the codename as the intrusion. PR #610 implemented that reading and is being closed. The symptom described was real; the conclusion was not.*

l0ng-ai:
Fixes #604.

## What was happening

Two things, and the bug needs both of them.

**The codename was never shown to the window that asked for it.** Every
workspace a window creates is created with a generated name —
`fresh_workspace_name` mints it and `pull_or_create` / `pull_workspace` send it
with `WorkspaceCreate`. The tree mutation excludes the origin subscriber from
its own deltas, so the `WorkspaceCreated` delta carrying that name never comes
back, and `note_synced_workspace` files the workspace in the mirror as
`Workspace { id, ..default }` — unnamed. `display_name_of` preferred `ws.name`,
so up to that point the chip read the directory the shells were working in, and
that is the only reason it ever did.

The first time the window pulled the whole tree again, the codename arrived and
took the chip. A daemon restart does that (`resync_local_windows_from_tree` →
`hydrate` → `MachineGet` → `MachineMirrors::install`), but so does every local
reconnect (`LocalLink` refreshes the mirror on connect), every mirror re-pull
after a delta falls behind, and simply launching tty7 the next morning — a
restored window hydrates from the tree before it draws. `record_geometry` then
stamped the new name into `views.json` as the entry's label.

**The rebuild also cost the window its directory.** Pane records ride inside the
deltas a window is excluded from, so a pane it restored is in its mirror's tabs
with no record behind it — no cwd. The daemon closes that gap only when some
fact later *changes*, and a pane restored with the cwd it will keep changes
nothing. So after a rebuild `subject_path_of` answers `None`, and
`record_geometry` wrote that `None` back over the subject the entry remembered.

Verified in a dev instance rather than read off the code: on `664b766`, a fresh
config dir gives `machine.json` `name: "dusky-tern"` with the chip reading
`verify-main`; killing the GUI and relaunching it — daemon untouched — is enough
to leave the chip reading `dusky-tern` and `views.json` holding
`"label": "dusky-tern"`.

## The fix

- `codename::is_generated` recognises a name this codebase minted, including the
  numbered form `unique` falls back to.
- A generated name is treated as the placeholder it is. It still names the
  workspace in the tree and to the CLI — `tty7 ws ls` and `tty7 ws rename
  keen-marten …` are unchanged — but on screen only a name someone chose
  outranks the directory. A workspace with no directory to borrow still reads
  its codename rather than "Untitled", which is the list codenames were minted
  for.
- The directory is read through `subject_path`, which already falls back to the
  subject the entry remembers, and `display_hint` now reads both halves through
  it, so the save that follows a rebuild writes the directory back instead of
  dropping it.

Someone who names a workspace "wild-yak" by hand loses this coin flip and reads
their directory in the chip instead; 576 pairs of animal words seemed a cheap
thing to give up against every window renaming itself.

## Neighbouring cases

- **An explicitly renamed workspace keeps its name** through a rebuild — it is
  in the tree and is not a codename. Checked by hand: `tty7 ws rename pale-otter
  deploy`, chip reads `deploy`, still `deploy` after `kill -9` of the daemon.
- **Remote workspaces** go through the same create, the same mirror and the same
  `display_name`, so they are covered by the same change.
- **`views.json` quarantine-and-rebuild** is untouched: no field was added or
  removed, and an entry written by an older build still parses.

## Verification

A local debug build driven by hand in an isolated dev instance
(`--config-dir`, own daemon), plus the unit tests:

| | before | after |
|---|---|---|
| fresh launch, tree name `pale-otter` | `verify-main` | `verify-main` |
| `kill -9` the daemon, window rebuilt | `dusky-tern` | `verify-main` |
| quit and relaunch tty7 | `dusky-tern` | `verify-main` |
| renamed to `deploy`, then `kill -9` | — | `deploy` |

`a_rebuild_does_not_rename_a_workspace_to_its_codename` walks that whole
sequence against the mirror, including the rebuild that leaves the workspace's
pane with no record behind it. Two tests in `codename` pin the recogniser to
what the generator actually produces.

## Known follow-up, not fixed here

The mirror genuinely loses the pane records for panes its own window created —
`subject_path`'s memory papers over it for the workspace's directory, but the
mirror is wrong until something else re-pulls it. Worth its own issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A workspace no longer renames itself to a codename it was never shown.**
+  Every workspace a window creates is handed a generated name — `keen-marten` —
+  so the machine tree and `tty7 ws ls` have something to call it, while the
+  window itself reads the directory its shells are working in. The window is
+  left out of the deltas its own edits raise, so it never heard which codename
+  it got; the first time it pulled the whole tree again — a daemon restart, a
+  rebuild, or simply launching tty7 the next morning — the codename took the
+  workspace chip and was stamped into `views.json` from there, and that pull
+  had by then also cost the window the pane record it read the directory from.
+  A generated name is now treated as the placeholder it is: it still names the
+  workspace in the tree and to the CLI, but on screen only a name someone
+  actually chose outranks the directory, which the window now remembers across
+  a rebuild. A workspace with no directory to borrow still reads its codename
+  rather than "Untitled" (#604).
 - **A local daemon that dies and comes back no longer leaves a window of dead
   panes looking live** — from the client's side a killed daemon is
   indistinguishable from one whose shells all exited at once, so the window
--- a/crates/tty7-core/src/core/codename.rs
+++ b/crates/tty7-core/src/core/codename.rs
@@ -74,6 +74,32 @@ pub fn unique(taken: impl Fn(&str) -> bool) -> String {
     Names::new().unique(taken)
 }
 
+/// Does this read like a name minted here rather than typed by a person?
+///
+/// A client names every workspace it creates, so "this workspace has a name"
+/// and "somebody chose that name" are two different questions — and only the
+/// second one should outrank the directory a window's shells are working in.
+/// The tree cannot tell them apart, so the words themselves have to (#604).
+///
+/// Someone who names a workspace "wild-yak" by hand loses this coin flip and
+/// reads the directory in the chip instead. 576 pairs of animal words is a
+/// cheap thing to give up against every window renaming itself the next time
+/// its machine tree is pulled.
+pub fn is_generated(name: &str) -> bool {
+    // `unique` gives up on bare pairs after enough collisions and starts
+    // hanging a number off the end, so "keen-marten-7" is one of ours too.
+    let stem = match name.rsplit_once('-') {
+        Some((stem, count)) if !count.is_empty() && count.bytes().all(|b| b.is_ascii_digit()) => {
+            stem
+        }
+        _ => name,
+    };
+    matches!(
+        stem.split_once('-'),
+        Some((adjective, noun)) if ADJECTIVES.contains(&adjective) && NOUNS.contains(&noun)
+    )
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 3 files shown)


l0ng-ai:
Closing this — the direction is wrong, and the argument that settles it is one this PR never made.

The chip is the workspace switcher. Its job is to identify the workspace, so it has to show the workspace's *name* — the one in the tree. `tty7 ws ls` and name-addressing already use that name, so a chip showing the directory means the GUI and the CLI give two different answers to "what is this workspace called", and a user reading `verify-main` on screen cannot address it by that name. That is the real defect. The rename-on-restart in #604 is the symptom that made it visible.

So the codename is not a placeholder to be outranked, and `codename::is_generated` is a recogniser nobody needs. The fix belongs at the mirror instead: the window that created a workspace is never told the name the machine gave it, because a client is left out of the deltas its own ops raise and both create paths throw away the name the reply carries. Teach it that name and the codename is on screen from the first frame and never changes under anyone — which is what #604 actually asked for.

A replacement is on its way; I will link it here.

The pane-record half this PR turned up is real and survives the reversal — a window's mirror holds no `PaneRecord` for a pane that window created — so it is now #612 on its own.

Branch kept for now in case anything here is worth salvaging.

l0ng-ai:
Replacement is up: #613.

PR #630, opened 2026-08-14
feat(terminal): let the shell's own line editor own the prompt when asked (#624)

### LINKED ISSUE #624: Add native shell input mode: prompt editor bypasses ZLE and keeps stale cross-pane history
### Type

### What's going on?

tty7's inline prompt editor intercepts physical keyboard input whenever OSC 133 shell integration reports an active prompt. There is currently no global setting that lets users keep shell integration while sending prompt input directly to the shell's own line editor (zsh ZLE/readline/etc.).

This makes native shell keybindings appear broken and can make shared history look as if it is not shared between panes.

#### Reproduction

Environment:

- tty7 26.8.3 on macOS Apple Silicon
- zsh
- two panes using the same `HISTFILE`
- `SHARE_HISTORY` enabled
- `APPEND_HISTORY` and `INC_APPEND_HISTORY` disabled
- `per_pane_history: false`
- `history_search: false`
- Up is bound in zsh to `history-beginning-search-backward-end`

Steps:

1. Open two zsh panes, A and B.
2. In pane A, run a unique command such as `echo TTY7_SHARED_HISTORY_MARKER`.
3. In pane B, run `fc -RI` or `history` and verify that the command from pane A is present in zsh's history.
4. At a fresh prompt in pane B, type a matching prefix and press the physical Up key.

Actual behavior:

- The zsh history file and `history` output contain pane A's command, so shell history sharing itself works.
- Physical prefix + Up does not invoke the configured ZLE widget. It can replace the input with an unrelated command from tty7's own per-view history (for example, typing `h` then Up produced `q`).
- A temporary ZLE probe sees no physical `h` or Up input.
- Injecting the same input with `tty7 send` reaches the pane PTY, invokes ZLE, and finds the expected shared-history entry.

Expected behavior:

- Users should be able to select a native shell-input mode in which physical keyboard input at the prompt is passed to the PTY and handled by ZLE/readline, while optionally retaining OSC 133 integration for prompt/cwd/command lifecycle tracking.
- `history_search: false` should not leave users without a way to disable the rest of tty7's prompt-key interception.

#### Why the histories disagree

The two paths use different editors and different history state:

1. When `input_active()` is true, `TerminalView::on_key_down` calls `handle_editor_key` instead of writing the key to the PTY (`src/terminal/view.rs:1735`).
2. Physical Up calls tty7's `history_prev()` (`src/terminal/view.rs:2088`).
3. `history_prev()` ignores the current prefix and selects from the `TerminalView`'s local `history` vector (`src/terminal/view.rs:3702`).
4. Each `TerminalView` loads its history snapshot when the view is constructed (`src/terminal/view.rs:1159`). A command submitted in one view updates that view and appends to tty7's history store, but there is no cross-view invalidation/reload for already-open panes (`src/terminal/view.rs:3602`, `src/terminal/history.rs:228`).
5. `history_search: false` only hands Ctrl-R to the shell (`src/terminal/view.rs:2005`); it does not disable the inline editor or hand Up/printable input to the shell.
6. `tty7 send` bypasses the GUI editor and writes to an explicit server pane, which explains why the CLI control works.

`per_pane_history: false` is working as implemented: `TTY7_HISTFILE` is not injected and both zsh processes use the same real history file. The confusing behavior comes from tty7's separate in-memory prompt history.

#### Temporary workaround

Launch zsh through an executable that tty7 does not identify as zsh, so the prompt integration/editor does not arm:

```json
"shell": {
  "program": "/bin/sh",
  "args": [
    "-c",
    "exec env SHELL=/opt/homebrew/bin/zsh /opt/homebrew/bin/zsh -l"
  ]
}
```

New panes then use native ZLE and shared prefix history correctly. Existing panes must be recreated.

This is only a workaround: it also disables useful tty7 shell-integration features, and it relies on hiding the real shell executable from tty7.

Using a zsh entry under `custom_shells` is not sufficient in 26.8.3. `shell_integration::setup` unconditionally calls `setup_zsh()` for every recognized zsh program; unlike Bash/PowerShell/WSL, the zsh branch does not check `has_custom_args` (`crates/tty7-core/src/daemon/shell_integration.rs:1049`). This also appears inconsistent with the configuration documentation saying custom shells are launched exactly as written with no shell integration.

#### Proposed long-term solution

Please add a global, user-visible setting independent of `history_search` and `tab_completion`, for example:

```json
"shell_input_mode": "native"
```

or:

```json
"prompt_editor": false
```

In native mode:

- physical key events, macOS text/IME commits, and paste at a shell prompt should go to the pane PTY;
- the shell's own ZLE/readline keybindings and history traversal should own editing;
- OSC 133 shell integration may remain enabled for prompt boundaries, cwd tracking, exit status, and notifications;
- tty7's local CmdEditor, ghost suggestions, completion UI, Ctrl-R UI, and local Up/Down history should not intercept prompt input.

This likely needs a consistent gate around the local-editor branch in `TerminalView::on_key_down` and the text/IME/paste paths, not only a special case for Up.

If tty7's inline editor remains enabled, its history should additionally receive cross-view invalidation/reload when another pane appends a command. Prefix-aware Up behavior could also be configurable, but synchronization alone would not solve native ZLE compatibility.

Suggested acceptance tests:

- With native input mode enabled, physical `h` + Up reaches the PTY and invokes a zsh ZLE widget.
- The mode works while OSC 133 prompt/cwd tracking remains active.
- With two live panes and shared zsh history, a command from pane A is available to native ZLE in pane B after normal zsh import semantics.
- Existing inline-editor behavior is unchanged when the new mode is disabled.
- A recognized zsh entry can explicitly disable integration without using an executable-name wrapper.

### tty7 version (bugs)

26.8.3 (`main`/nightly commit `84f9d54` was also inspected)

### Platform (bugs)

macOS (Apple Silicon)

### Anything else?

The diagnosis was verified against two live panes and the 26.8.3 source. The key observation is that shell history was already correctly shared; physical GUI input and `tty7 send` were exercising different input owners.

Related but not duplicate: #182 documents that shell-level key handlers cannot work while CmdEditor owns the prompt, and #301 covers one swallowed macOS shortcut in a foreground TUI. This report asks for an explicit native prompt-input mode and covers stale cross-pane CmdEditor history.

biztex:
Closes #624 — the proposed `prompt_editor: false` global setting.

## What it does

A global `prompt_editor` setting, on by default so nothing changes for anyone who does not ask. Off, tty7's inline editor never arms at a prompt: physical keys, IME commits, mouse caret placement and paste all go straight to the PTY, and the shell's own line editor — ZLE, readline — owns the line. Up runs whatever the user bound there against the shell's own history rather than tty7's per-view copy, which goes stale the moment a sibling pane runs something. `tab_completion: false` and `history_search: false` each hand back one key; this hands back the prompt.

Shell integration is deliberately untouched — prompt marks, cwd tracking, command lifecycle and notifications keep working — because the gate sits in `input_inactive_reason`, the one predicate every editor entry point (key-down, `commit_text`, the mouse caret, paste, the input-bar render) already consults, not in the OSC 133 plumbing.

That predicate runs without an `App` handle, so the view mirrors the flag the way it already mirrors `mouse_reporting`: written at construction, rewritten by the Settings switch and by the config watcher. Flipping it reaches every open pane without a restart.

The switch lives at Settings → Input, answers settings search in all three locales, and is documented beside its siblings (`configuration.mdx`, `prompt.mdx`, a pointer from `history.mdx`).

## Tests

`cargo test -p tty7` 1522 passed, `-p tty7-core` 1126 passed. New: `prompt_editor_off_leaves_the_prompt_to_the_shell` — asserts a typed character and Up reach the PTY as raw bytes (`a`, `ESC [ A`) with no local overlay, and that flipping the setting back re-arms the editor at the same prompt. The acceptance point I could not verify here is a live macOS zsh + ZLE session; the byte-level behavior is proven at the PTY boundary.


l0ng-ai:
Thank you for this — closing in favour of #633, which landed the same setting in 72db26d. The two agreed on the design (the gate in `input_inactive_reason`, a mirrored field on the view because that predicate has no `App`, the config watcher refreshing live panes), so this is not a disagreement about the approach. What decided it was three things this branch does not cover:

**The cascade.** `tab_completion` and `history_search` are menus tty7 opens *inside* the editor, so with `prompt_editor: false` both keys already belong to the shell — `tab_pressed` returns at the gate before `complete_tab` ever reads the flag. Here both switches stay lit and clickable while doing nothing, with nothing on the page to say why. #633 greys those two rows, disables the switches and appends the reason, while leaving their stored values alone so they come back with the editor.

**`shell_owns_prompt()` does not learn the flag**, and that one matters beyond tidiness:

- `gap_holdable()` is `shell_active() && !on_alt_screen() && !shell_owns_prompt()`, so it stays true at a native prompt. On macOS a printable key defers to the IME and arrives through `commit_text` → `write_gap_text` → `hold_text`, which *holds* it: the first burst of typing in a pane reaches the PTY only when the 150 ms `HOLD_WINDOW` timer dumps it. Backspace takes the same route on every platform, through the hold branch in `on_key_down`. Nothing is lost — the ordering survives, because the next non-backspace key calls `release_hold()` first — but the machinery doing it exists solely to feed tty7's editor, which in this mode will never consume it.
- The typeahead record keeps accumulating for the same reason, and the thing that consumes it is `flush_typeahead()`, which writes `^U` to erase the line before moving the text into `cmd`. That cannot fire while the mode is on, but it is waiting on the way back out, on a line only ZLE has been editing.

(Read from the code paths against current `main` rather than reproduced — the tests there are unix-only.)

**A half-typed line.** Flipping the switch mid-line leaves the text in `cmd` with the input bar gone: it is neither on the shell's prompt nor on screen, and it reappears later, possibly mixed with what was typed in between. #633 hands it over the way an unknown chord does, so it is still on the prompt to finish under ZLE.

Two smaller ones: no CHANGELOG entry, and `SettingsPromptIntro` still promises only "menus".

One thing this branch had that #633 did not: the assertion that flipping the setting back on re-arms the editor at the same prompt. I took it — it is in the merged test.

#629 is untouched by this and is being judged on its own; it covers the other half of #624, and the argument in it (a `ZDOTDIR` swap changes which startup files run even without touching argv) is a better answer than the one I had reached.

PR #637, opened 2026-08-14
feat(shell): give Nushell panes OSC 7 cwd tracking and OSC 133 marks

ArnoChenFx:
Nushell panes were the only ones without shell integration: `shell_kind()` in `daemon::shell_integration` never matched `nu`, so tty7 spawned them with no OSC 7 cwd reports and could not follow `cd` — while pwsh, zsh, bash and fish all track the working directory today. This adds a Nushell integration with the same full OSC 133 prompt cycle (A/B/C/D) plus OSC 7 cwd reporting.

## Why a wrapper config.nu

Nushell has no environment variable that redirects its config file the way `ZDOTDIR` does for zsh, so the injection rides `nu --config <throwaway>`: a config.nu that sources the user's own config.nu back in first, then appends the hooks (user config first, so a wholesale `$env.config = {...}` cannot clobber them).

One constraint shaped the design: `source` is a **parse-time** construct in Nushell. It cannot be guarded at runtime, cannot take a variable, and naming a missing file is a hard parse error. The Rust side therefore resolves the user's config path with the same rules `$nu.default-config-dir` uses (`%APPDATA%\nushell` / `$XDG_CONFIG_HOME/nushell` / `~/.config/nushell`), checks it exists, and substitutes a literal — or a no-op comment on machines without a config.nu.

## What the hooks do

- `pre_prompt`: OSC 7 cwd (`%` → `%25`, `\` → `/`, leading slash for Windows drive paths so the daemon's `parse_osc7` round-trips), `133;A`, and `133;D` with `LAST_EXIT_CODE` — gated on a flag the `pre_execution` hook arms, so the very first prompt emits no bogus exit status.
- `pre_execution`: bare `133;C` (Nushell hooks receive no command text) and the D-gate flag.
- `prompt_indicator`: wrapped for `133;B` **only when the config defines one**. Recent Nushells (0.102+) draw a built-in prompt with their own indicator and B mark — and 0.114 removed `prompt_indicator` from the config schema — so replacing a missing indicator would erase the glyph. Overlapping marks with Nushell's built-in OSC 133 output are merged by the daemon's prompt-state machine (consecutive identical `at_prompt` states collapse).

## Scope notes

- Remote SSH and WSL panes are unchanged: their bootstraps are parsed by the far-end login shell / `sh`, which cannot carry a Nushell config, and nu-as-login-shell sessions would break on a heredoc bootstrap. No regression: they keep the plain login shell they had.
- The user's config.nu is read, never written.

### DIFF
--- a/crates/tty7-core/src/core/shells.rs
+++ b/crates/tty7-core/src/core/shells.rs
@@ -518,6 +518,11 @@ pub fn git_bash_path() -> Option<PathBuf> {
 }
 
 #[cfg(all(windows, test))]
+pub fn nushell_path() -> Option<PathBuf> {
+    find_in_path("nu.exe")
+}
+
+#[cfg(all(windows, test))]
 mod wsl_tests {
     #[test]
     fn the_default_distro_is_one_of_the_installed_ones() {
--- a/crates/tty7-core/src/daemon/shell_integration.rs
+++ b/crates/tty7-core/src/daemon/shell_integration.rs
@@ -654,8 +654,14 @@ if (($env.TTY7_SHELL_INTEGRATION? | default "") == "") {
     # OSC 7 cwd, so the daemon tracks the pane's directory across `cd`. The
     # payload is percent-decoded on the daemon side, so a literal `%` must be
     # escaped as %25, and a Windows drive path needs the leading slash that
-    # makes `C:/…` an absolute URI path (`/C:/…`).
-    let __tty7_path = ($env.PWD | str replace -a '\' '/' | str replace -a '%' '%25')
+    # makes `C:/…` an absolute URI path (`/C:/…`). Backslashes are separators
+    # only on Windows — a Unix path may legally contain one, so that
+    # translation is gated on the platform the way nu's own OSC 7 gates it.
+    let __tty7_path = if (($nu.os-info.name? | default '') == 'windows') {
+      ($env.PWD | str replace -a '\' '/' | str replace -a '%' '%25')
+    } else {
+      ($env.PWD | str replace -a '%' '%25')
+    }
     let __tty7_path = if ($__tty7_path | str starts-with '/') { $__tty7_path } else { '/' + $__tty7_path }
     do $__tty7_osc $"7;file://($env.COMPUTERNAME? | default 'localhost')($__tty7_path)"
     do $__tty7_osc "133;A"
@@ -872,10 +878,21 @@ fn setup_powershell() -> Option<Injection> {
 /// ZDOTDIR does for zsh, so the injection rides `--config` instead: a
 /// throwaway config.nu that sources the user's real one and then appends the
 /// OSC hooks. The `dir` is what makes the file disappear when the pane closes.
+///
+/// Trade-off, inherent to `--config`: `$nu.config-path` points at the wrapper,
+/// so `config nu` inside the pane edits a file that vanishes with it. The zsh
+/// path dodges the analogue by restoring ZDOTDIR after startup; Nushell's
+/// `$nu.*` paths are immutable, so this is accepted rather than fought.
 fn setup_nushell() -> Option<Injection> {
+    setup_nushell_with(nushell_user_config_path().as_deref())
+}
+
+/// Split from `setup_nushell` so a test can drive the whole wrapper against a
+/// config.nu of its own without writing into the real one.
+fn setup_nushell_with(user_config: Option<&Path>) -> Option<Injection> {
     let dir = throwaway_dir("tty7-nu-")?;
     let config = dir.join("config.nu");
-    std::fs::write(&config, nushell_config_script()).ok()?;
+    std::fs::write(&config, nushell_config_script_with(user_config)).ok()?;
 
     Some(Injection {
         env: HashMap::new(),
@@ -893,10 +910,6 @@ fn setup_nushell() -> Option<Injection> {
 /// it cannot be guarded at runtime or name a missing file — so the path is
 /// resolved and substituted here, and a machine without a config.nu gets a
 /// no-op line instead.
-fn nushell_config_script() -> String {
-    nushell_config_script_with(nushell_user_config_path().as_deref())
-}
-
 fn nushell_config_script_with(user_config: Option<&Path>) -> String {
     let source = match user_config {
         Some(path) => format!("source {}", nu_string_literal(&path.to_string_lossy())),
@@ -905,25 +918,61 @@ fn nushell_config_script_with(user_config: Option<&Path>) -> String {
     NUSHELL_INTEGRATION.replace("__TTY7_SOURCE_USER_CONFIG__", &source)
 }
 
-/// Where nu would load its config.nu from, resolved the way nu resolves it:
-/// `%APPDATA%\nushell` on Windows, `$XDG_CONFIG_HOME/nushell` (or
-/// `$HOME/.config/nushell`) elsewhere. Only a file that exists comes back —
-/// the wrapper's `source` cannot name one that is not there.
+/// Where nu would load its config.nu from. Only a file that exists comes
+/// back — the wrapper's `source` cannot name one that is not there.
 fn nushell_user_config_path() -> Option<PathBuf> {
-    #[cfg(windows)]
-    let dir = std::env::var_os("APPDATA")
-        .map(PathBuf::from)?
-        .join("nushell");
-    #[cfg(not(windows))]
-    let dir = {
-        let base = match std::env::var_os("XDG_CONFIG_HOME").filter(|v| !v.is_empty()) {
-            Some(xdg) => PathBuf::from(xdg),
-            None => PathBuf::from(std::env::var_os("HOME")?).join(".config"),
-        };
-        base.join("nushell")
-    };
-    let config = dir.join("config.nu");
+    let config = nushell_config_dir()?.join("config.nu");
     config.is_file().then_some(config)
+}
+
+/// nu's own config directory, resolved the way nu resolves it. Getting this
+/// wrong is silent and total: the wrapper reports "no config.nu to restore"
+/// for a user who has one, and `nu --config` then replaces their config with
+/// one that never sources it.
+///
+/// nu-path's `configurable_dir_path` consults `$XDG_CONFIG_HOME` on *every*
+/// platform, Windows included, and only when it is non-empty **and** absolute;
+/// anything else falls through to `dirs::config_dir()`.
+fn nushell_config_dir() -> Option<PathBuf> {
+    nushell_config_dir_from(
+        std::env::var_os("XDG_CONFIG_HOME").as_deref(),
+        platform_config_dir(),
+    )
+}
+
+fn nushell_config_dir_from(
+    xdg_config_home: Option<&std::ffi::OsStr>,
+    platform_default: Option<PathBuf>,
+) -> Option<PathBuf> {
+    let base = match xdg_config_home {
+        Some(xdg) if !xdg.is_empty() && Path::new(xdg).is_absolute() => PathBuf::from(xdg),
+        _ => platform_default?,
+    };
+    Some(base.join("nushell"))
+}
+
+/// `dirs::config_dir()` — what nu falls back to. Note this is **not** where
+/// tty7 keeps its own config (`~/.config/tty7` on macOS too); nu follows the
+/// platform convention, and the two are only the same directory on Linux.
+#[cfg(windows)]
+fn platform_config_dir() -> Option<PathBuf> {
+    // `dirs` asks Windows for FOLDERID_RoamingAppData; `%APPDATA%` is the
+    // same directory in every environment tty7 can spawn a pane in.
+    std::env::var_os("APPDATA")
+        .filter(|v| !v.is_empty())
+        .map(PathBuf::from)
+}
+
+#[cfg(target_os = "macos")]
+fn platform_config_dir() -> Option<PathBuf> {
+    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
+    Some(PathBuf::from(home).join("Library/Application Support"))
+}
+
+#[cfg(all(unix, not(target_os = "macos")))]
+fn platform_config_dir() -> Option<PathBuf> {
+    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
+    Some(PathBuf::from(home).join(".config"))
 }
 
 /// A Nushell string literal for `path`. Single quotes are fully literal in
@@ -1602,6 +1651,7 @@ mod tests {
         let mut out = Vec::new();
         let mut answered = 0usize;
         let mut typed = false;
+        let mut seen_fail = None;
         while std::time::Instant::now() < deadline {
             match rx.recv_timeout(std::time::Duration::from_millis(200)) {
                 Ok(chunk) => out.extend_from_slice(&chunk),
@@ -1644,8 +1694,17 @@ mod tests {
             // emit an unpaired D while drawing its first prompt, and breaking on
             // that would sample the transcript before the prompt cycle under test
             // has run at all.
-            if typed && text.contains(FAILED_COMMAND_MARK) {
-                break;
+            if typed && text.contains(FAILED_COMMAND_MARK) && seen_fail.is_none() {
+                seen_fail = Some(std::time::Instant::now());
+            }
+            // The D mark and the rest of the same prompt cycle (the cwd report,
+            // the A mark) are separate writes: breaking the moment the mark
+            // arrives can sample the transcript before the report does. Hold
+            // the pty open briefly so the cycle's tail lands.
+            if let Some(at) = seen_fail {
+                if at.elapsed() >= std::time::Duration::from_millis(500) {
+                    break;
+                }
             }
         }
         // Reap before asserting: a panic here would otherwise leave the shell
@@ -2675,7 +2734,10 @@ mod tests {
         assert_eq!(inj.args[0], "--config");
         assert_eq!(inj.args.len(), 2);
         let written = std::fs::read_to_string(&inj.args[1]).expect("config written");
-        assert_eq!(written, nushell_config_script());
+        assert_eq!(
+            written,
+            nushell_config_script_with(nushell_user_config_path().as_deref())
+        );
         assert!(inj.env.is_empty());
         assert!(!inj.replaces_argv);
         let _ = std::fs::remove_dir_all(&dir);
@@ -2689,8 +2751,9 @@ mod tests {
         }
         assert!(s.contains("7;file://"));
         assert!(
-            s.contains(r#"str replace -a '\' '/'"#),
-            "must slash Windows paths"
+            s.contains(r#"($nu.os-info.name? | default '') == 'windows'"#),
+            "the backslash translation must be gated on Windows — a Unix path \
+             may legally contain a literal backslash"
         );
         assert!(
             s.contains(r#"str replace -a '%' '%25'"#),
@@ -2754,6 +2817,178 @@ mod tests {
         );
     }
 
+    /// nu-path resolves the config dir from `$XDG_CONFIG_HOME` on every
+    /// platform, but only when it is non-empty *and* absolute. Both halves are
+    /// load-bearing: the original Windows arm ignored the variable outright,
+    /// and the original non-Windows arm accepted a relative value nu rejects.
+    #[test]
+    fn nushell_config_dir_follows_nu_s_own_resolution_rules() {
+        use std::ffi::OsStr;
+        // `/x` is not absolute on Windows (no prefix), so the fixtures have to
+        // be shaped for the host or the rule under test is never exercised.
+        let abs_xdg = if cfg!(windows) { r"C:\xdg" } else { "/xdg" };
+        let platform = || {
+            Some(PathBuf::from(if cfg!(windows) {
+                r"C:\platform"
+            } else {
+                "/platform"
+            }))
+        };
+        let want = |base: &str| Some(PathBuf::from(base).join("nushell"));
+
+        // Unset, empty, or relative: nu ignores it and takes the platform dir.
+        for ignored in [None, Some(OsStr::new("")), Some(OsStr::new("relative/dir"))] {
+            assert_eq!(
+                nushell_config_dir_from(ignored, platform()),
+                want(&platform().unwrap().to_string_lossy()),
+                "XDG_CONFIG_HOME {ignored:?} must not displace the platform dir"
+            );
+        }
+
+        // Non-empty and absolute: it wins, even where there is no platform dir.
+        assert_eq!(
+            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), platform()),
+            want(abs_xdg)
+        );
+        assert_eq!(
+            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), None),
+            want(abs_xdg)
+        );
+        assert_eq!(nushell_config_dir_from(None, None), None);
+    }
+
+    /// The macOS half of the same rule, stated as itself: nu follows the
+    /// platform convention there, which is *not* where tty7 keeps its own
+    /// config (`~/.config/tty7`). Reading `~/.config/nushell` on a Mac finds
+    /// nothing, and the wrapper then replaces a config it thinks is absent.
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn nushell_config_dir_is_application_support_on_macos() {
+        let home = PathBuf::from(std::env::var_os("HOME").expect("HOME"));
+        assert_eq!(
+            platform_config_dir(),
+            Some(home.join("Library/Application Support"))
+        );
+    }
+
+    /// Ask the real binary. A hand-copy of nu's resolution rules is only as
+    /// good as the reading behind it — this is the check that fails outright
+    /// if the two ever disagree on this machine.
+    #[cfg(unix)]
+    #[test]
+    fn nushell_config_dir_is_what_nu_itself_reports() {
+        let Some(nu) = test_nushell_path() else {
+            eprintln!("skipping: Nushell not installed");
+            return;
+        };
+        let xdg_absolute = std::env::temp_dir().join("tty7-nu-xdg");
+        let cases: [Option<&std::ffi::OsStr>; 4] = [
+            None,
+            Some(std::ffi::OsStr::new("")),
+            Some(std::ffi::OsStr::new("relative/dir")),
+            Some(xdg_absolute.as_os_str()),
+        ];
+        for xdg in cases {
+            let mut cmd = std::process::Command::new(&nu);
+            cmd.args(["-n", "-c", "$nu.default-config-dir"]);
+            match xdg {
+                Some(v) => cmd.env("XDG_CONFIG_HOME", v),
+                None => cmd.env_remove("XDG_CONFIG_HOME"),
+            };
+            let out = cmd.output().expect("run nu");
+            // A relative value makes nu warn before it answers, so take the
+            // last line, not the whole of stdout.
+            let stdout = String::from_utf8_lossy(&out.stdout);
+            let theirs = PathBuf::from(stdout.lines().last().unwrap_or_default().trim());
+            let ours = nushell_config_dir_from(xdg, platform_config_dir())
+                .expect("this machine has a config dir");
+            assert_eq!(
+                ours, theirs,
+                "tty7 and nu disagree on the config dir for XDG_CONFIG_HOME={xdg:?}"
+            );
+        }
+    }
+
+    /// End to end over a real pty: a user who *has* a config.nu must still
+    /// have it after tty7 points `nu --config` at the wrapper. `--config`
+    /// replaces the user's config rather than adding to it, so a wrapper that
+    /// fails to source it silently strips their prompt, aliases and
+    /// keybindings.
+    #[cfg(unix)]
+    #[test]
+    fn nushell_restores_the_user_config_and_still_reports_the_prompt_cycle() {
+        let Some(nu) = test_nushell_path() else {
+            eprintln!("skipping: Nushell not installed");
+            return;
+        };
+        let nu = nu.to_string_lossy().into_owned();
+        let home = tempfile::tempdir().expect("tempdir");
+        let user_config = home.path().join("config.nu");
+        // Shaped like a real config.nu: a wholesale `$env.config = {...}` (the
+        // conventional ending, and the one that would wipe hooks added before
+        // it), plus a marker only the user's own file could have set.
+        std::fs::write(
+            &user_config,
+            "$env.config = { show_banner: false }\n$env.TTY7_USER_CONFIG_LOADED = \"yes\"\n",
+        )
+        .expect("write user config");
+
+        let mut injection = setup_nushell_with(Some(&user_config)).expect("nushell integration");
+        // What `setup` does at every spawn boundary, and what this test
+        // bypasses by reaching for `setup_nushell_with` directly. Without it
+        // the wrapper is a no-op whenever the test itself runs inside an
+        // integrated pane — the guard sees the inherited `1` and skips the
+        // whole block, and the marks in the transcript are then Nushell's own.
+        injection
+            .env
+            .insert("TTY7_SHELL_INTEGRATION".to_string(), String::new());
+        let text = prompt_cycle_over_pty(
+            &nu,
+            &injection,
+            // One line: Nushell's line editor holds the tty in raw mode and
+            // drops what arrives while a command is running, so a second `\r`
+            // never reaches it and its prompt cycle never happens. `1 / 0` is
+            // the failure — a bare `false` in Nushell is a value.
+            b"print $\"probe=($env.TTY7_USER_CONFIG_LOADED? | default 'MISSING')\"; 1 / 0\r",
+            Some(home.path()),
+        );
+
+        assert!(
+            text.contains("probe=yes"),
+            "the user's own config.nu must survive --config; got:\n{text}"
+        );
+        // Assert on tty7's *own* marks, not on the substring `133;A`: stock
+        // Nushell emits its own OSC 133 cycle alongside these, so a bare
+        // substring passes even when the wrapper never ran. tty7's hooks
+        // terminate with BEL and Nushell's with ST, which tells them apart.
+        for mark in ["133;A\u{7}", "133;C\u{7}", FAILED_COMMAND_MARK] {
+            assert!(
+                text.contains(mark),
+                "Nushell must still report {mark:?} after sourcing the user config; got:\n{text}"
+            );
+        }
+        // Same reasoning for the cwd: the hook falls back to `localhost` for
+        // the host, while Nushell's native OSC 7 names the real one.
+        assert!(
+            text.contains("7;file://localhost/"),
+            "tty7's own OSC 7 must be among the reports; got:\n{text}"
+        );
+        assert_eq!(
+            last_osc7(&text),
+            home.path().canonicalize().expect("canonical tempdir"),
+            "the pane's reported cwd must be where nu actually is"
+        );
+    }
+
+    /// `shells::nushell_path` is Windows-only (nu has no fixed install path
+    /// there); on unix the binary is whatever `nu` resolves to on PATH.
+    #[cfg(unix)]
+    fn test_nushell_path() -> Option<PathBuf> {
+        std::env::split_paths(&std::env::var_os("PATH")?)
+            .map(|dir| dir.join("nu"))
+            .find(|p| p.is_file())
+    }
+
     #[cfg(windows)]
     #[test]
     fn nushell_reports_the_full_prompt_cycle_over_a_real_pty() {
@@ -2763,12 +2998,14 @@ mod tests {
         };
         let nu = nu.to_string_lossy().into_owned();
         let injection = setup(Some(&nu), &[], false).expect("nushell integration");
-        // `cd` to a second directory, then a command that fails — Nushell's
-        // `1 / 0` errors with exit status 1 (a literal `false` is just a value).
+        // One line: Nushell's line editor holds the tty in raw mode and can
+        // drop what arrives while a command is running, so a second `\r` may
+        // never reach it. `cd` first, then a command that fails — `1 / 0`
+        // errors with exit status 1 (a literal `false` is just a value).
         let text = prompt_cycle_over_pty(
             &nu,
             &injection,
-            b"cd C:/Windows\r1 / 0\r",
+            b"cd C:/Windows; 1 / 0\r",
             Some(Path::new("C:/")),
         );
 
@@ -2787,7 +3024,7 @@ mod tests {
         assert_eq!(
             last_osc7(&text),
             PathBuf::from("C:/Windows"),
-            "`cd` must move the pane's reported cwd"
+            "`cd` must move the pane's reported cwd; got:\n{text}"
         );
     }
 }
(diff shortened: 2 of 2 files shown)


l0ng-ai:
Thanks for this — the injection design is sound (sourcing the user's config *before* appending hooks is exactly right, and the `nu_string_literal` quoting holds up). I reviewed it on macOS with stock Nushell 0.114.1 driven under a real pty, and found one blocker plus a question about scope.

## Blocker: on macOS the wrapper discards the user's entire `config.nu`

`nushell_user_config_path()` resolves non-Windows as `$XDG_CONFIG_HOME/nushell` else `$HOME/.config/nushell`. Nushell doesn't use that on macOS — `nu_path::nu_config_dir()` is `configurable_dir_path("XDG_CONFIG_HOME", dirs::config_dir)`, and `dirs::config_dir()` on macOS is `~/Library/Application Support`:

```
$ nu -n -c '$nu.default-config-dir'
/Users/thomas/Library/Application Support/nushell
```

So on a stock Mac the lookup misses, the wrapper emits `# no user config.nu to restore`, and `nu --config <throwaway>` then *replaces* the user's config with one that never sources it. Every macOS Nushell pane loses their prompt, aliases, custom commands and keybindings — silently. Demonstrated with a fake `$HOME` holding `~/Library/Application Support/nushell/config.nu`:

```
plain nu               -> probe=yes
tty7 wrapper (as-is)   -> probe=MISSING
```

Windows is unaffected: `%APPDATA%` *is* `dirs::config_dir()` there, so your testing was valid — this is purely the non-Windows arm.

## Same root cause, smaller blast radius: `XDG_CONFIG_HOME` on Windows

`configurable_dir_path` is not `cfg`-gated, so nu honours `XDG_CONFIG_HOME` on **every** platform — but only when it is non-empty **and absolute**. Two consequences:

- the Windows arm ignores it outright, so a Windows user with cross-platform dotfiles hits the same silent config wipe;
- the non-Windows arm accepts a *relative* value, which nu rejects (it warns and falls back).

Verified against the real binary:

```
default              -> ~/Library/Application Support/nushell
XDG=/tmp/xdgtest     -> /tmp/xdgtest/nushell        (absolute wins)
XDG=relative/dir     -> warns, falls back to default
XDG=""               -> falls back to default
```

## Patch

Here's the fix with tests, if it's useful — `cargo fmt` clean, `tty7-core` lib suite green (1142 passed). Take it or reimplement it, whichever you prefer.

The tests are the part I'd most encourage keeping: one of them asks the real `nu` binary for `$nu.default-config-dir` under four different `XDG_CONFIG_HOME` settings and asserts our resolution agrees, so this class of bug can't come back silently. I confirmed they fail against the current code (`left: /Users/thomas/.config/nushell`, `right: /Users/thomas/Library/Application Support/nushell`).

<details>
<summary>Diff against the current head of this branch</summary>

```diff
diff --git a/crates/tty7-core/src/daemon/shell_integration.rs b/crates/tty7-core/src/daemon/shell_integration.rs
index 8804e9f..929ff4c 100644
--- a/crates/tty7-core/src/daemon/shell_integration.rs
+++ b/crates/tty7-core/src/daemon/shell_integration.rs
@@ -873,9 +873,15 @@ fn setup_powershell() -> Option<Injection> {
 /// throwaway config.nu that sources the user's real one and then appends the
 /// OSC hooks. The `dir` is what makes the file disappear when the pane closes.
 fn setup_nushell() -> Option<Injection> {
+    setup_nushell_with(nushell_user_config_path().as_deref())
+}
+
+/// Split from `setup_nushell` so a test can drive the whole wrapper against a
+/// user config.nu of its own without writing into the real one.
+fn setup_nushell_with(user_config: Option<&Path>) -> Option<Injection> {
     let dir = throwaway_dir("tty7-nu-")?;
     let config = dir.join("config.nu");
-    std::fs::write(&config, nushell_config_script()).ok()?;
+    std::fs::write(&config, nushell_config_script_with(user_config)).ok()?;
 
     Some(Injection {
         env: HashMap::new(),
@@ -893,10 +899,6 @@ fn setup_nushell() -> Option<Injection> {
 /// it cannot be guarded at runtime or name a missing file — so the path is
 /// resolved and substituted here, and a machine without a config.nu gets a
 /// no-op line instead.
-fn nushell_config_script() -> String {
-    nushell_config_script_with(nushell_user_config_path().as_deref())
-}
-
 fn nushell_config_script_with(user_config: Option<&Path>) -> String {
     let source = match user_config {
         Some(path) => format!("source {}", nu_string_literal(&path.to_string_lossy())),
@@ -905,27 +907,62 @@ fn nushell_config_script_with(user_config: Option<&Path>) -> String {
     NUSHELL_INTEGRATION.replace("__TTY7_SOURCE_USER_CONFIG__", &source)
 }
 
-/// Where nu would load its config.nu from, resolved the way nu resolves it:
-/// `%APPDATA%\nushell` on Windows, `$XDG_CONFIG_HOME/nushell` (or
-/// `$HOME/.config/nushell`) elsewhere. Only a file that exists comes back —
+/// Where nu would load its config.nu from. Only a file that exists comes back —
 /// the wrapper's `source` cannot name one that is not there.
 fn nushell_user_config_path() -> Option<PathBuf> {
-    #[cfg(windows)]
-    let dir = std::env::var_os("APPDATA")
-        .map(PathBuf::from)?
-        .join("nushell");
-    #[cfg(not(windows))]
-    let dir = {
-        let base = match std::env::var_os("XDG_CONFIG_HOME").filter(|v| !v.is_empty()) {
-            Some(xdg) => PathBuf::from(xdg),
-            None => PathBuf::from(std::env::var_os("HOME")?).join(".config"),
-        };
-        base.join("nushell")
-    };
-    let config = dir.join("config.nu");
+    let config = nushell_config_dir()?.join("config.nu");
     config.is_file().then_some(config)
 }
 
+/// nu's own config directory, resolved the way nu resolves it — get this wrong
+/// and the wrapper reports "no config.nu to restore" for a user who has one,
+/// and `--config` then replaces their config with one that never sources it.
+///
+/// nu-path's `configurable_dir_path` consults `$XDG_CONFIG_HOME` on *every*
+/// platform, Windows included, and only when it is non-empty **and absolute**;
+/// anything else falls through to `dirs::config_dir()`.
+fn nushell_config_dir() -> Option<PathBuf> {
+    nushell_config_dir_from(
+        std::env::var_os("XDG_CONFIG_HOME").as_deref(),
+        platform_config_dir(),
+    )
+}
+
+fn nushell_config_dir_from(
+    xdg_config_home: Option<&std::ffi::OsStr>,
+    platform_default: Option<PathBuf>,
+) -> Option<PathBuf> {
+    let base = match xdg_config_home {
+        Some(xdg) if !xdg.is_empty() && Path::new(xdg).is_absolute() => PathBuf::from(xdg),
+        _ => platform_default?,
+    };
+    Some(base.join("nushell"))
+}
+
+/// `dirs::config_dir()`, which nu falls back to. Note this is **not** where
+/// tty7 keeps its own config: `default_config_dir` is `~/.config/tty7` on macOS
+/// too, whereas nu follows the platform convention.
+#[cfg(windows)]
+fn platform_config_dir() -> Option<PathBuf> {
+    // `dirs` asks Windows for FOLDERID_RoamingAppData; %APPDATA% is the same
+    // directory in every environment tty7 can spawn a pane in.
+    std::env::var_os("APPDATA")
+        .filter(|v| !v.is_empty())
+        .map(PathBuf::from)
+}
+
+#[cfg(target_os = "macos")]
+fn platform_config_dir() -> Option<PathBuf> {
+    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
+    Some(PathBuf::from(home).join("Library/Application Support"))
+}
+
+#[cfg(all(unix, not(target_os = "macos")))]
+fn platform_config_dir() -> Option<PathBuf> {
+    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
+    Some(PathBuf::from(home).join(".config"))
+}
+
 /// A Nushell string literal for `path`. Single quotes are fully literal in
 /// Nushell, so they are the default; a path that contains an apostrophe (legal
 /// on Windows) falls back to double quotes with backslash escaping.
@@ -2675,7 +2712,10 @@ mod tests {
         assert_eq!(inj.args[0], "--config");
         assert_eq!(inj.args.len(), 2);
         let written = std::fs::read_to_string(&inj.args[1]).expect("config written");
-        assert_eq!(written, nushell_config_script());
+        assert_eq!(
+            written,
+            nushell_config_script_with(nushell_user_config_path().as_deref())
+        );
         assert!(inj.env.is_empty());
         assert!(!inj.replaces_argv);
         let _ = std::fs::remove_dir_all(&dir);
@@ -2754,6 +2794,177 @@ mod tests {
         );
     }
 
+    /// nu-path resolves the config dir from `$XDG_CONFIG_HOME` on every
+    /// platform, but only when it is non-empty *and absolute*. Both halves are
+    /// load-bearing: the old Windows arm ignored the variable outright, and the
+    /// old non-Windows arm accepted a relative value nu rejects.
+    #[test]
+    fn nushell_config_dir_follows_nu_s_own_resolution_rules() {
+        use std::ffi::OsStr;
+        // `/x` is not absolute on Windows (no prefix), so the fixtures have to
+        // be shaped for the host or the rule under test is never exercised.
+        let abs_xdg = if cfg!(windows) { r"C:\xdg" } else { "/xdg" };
+        let platform = || {
+            Some(PathBuf::from(if cfg!(windows) {
+                r"C:\platform"
+            } else {
+                "/platform"
+            }))
+        };
+        let want = |base: &str| Some(PathBuf::from(base).join("nushell"));
+
+        // Unset, empty, or relative: nu ignores it and takes the platform dir.
+        for ignored in [None, Some(OsStr::new("")), Some(OsStr::new("relative/dir"))] {
+            assert_eq!(
+                nushell_config_dir_from(ignored, platform()),
+                want(&platform().unwrap().to_string_lossy()),
+                "XDG_CONFIG_HOME {ignored:?} must not displace the platform dir"
+            );
+        }
+
+        // Non-empty and absolute: it wins, even where there is no platform dir.
+        assert_eq!(
+            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), platform()),
+            want(abs_xdg)
+        );
+        assert_eq!(
+            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), None),
+            want(abs_xdg)
+        );
+        assert_eq!(nushell_config_dir_from(None, None), None);
+    }
+
+    /// The macOS half of the same rule, stated as itself: nu follows the
+    /// platform convention there, which is *not* where tty7 keeps its own
+    /// config (`~/.config/tty7`). Reading `~/.config/nushell` on a Mac finds
+    /// nothing, and the wrapper then replaces a config it thinks is absent.
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn nushell_config_dir_is_application_support_on_macos() {
+        let home = PathBuf::from(std::env::var_os("HOME").expect("HOME"));
+        assert_eq!(
+            platform_config_dir(),
+            Some(home.join("Library/Application Support"))
+        );
+    }
+
+    /// Ask the real binary. A hand-copy of nu's resolution rules is only as good
+    /// as the reading behind it — this is the check that fails outright if the
+    /// two ever disagree on this machine.
+    #[cfg(unix)]
+    #[test]
+    fn nushell_config_dir_is_what_nu_itself_reports() {
+        let Some(nu) = test_nushell_path() else {
+            eprintln!("skipping: Nushell not installed");
+            return;
+        };
+        let xdg_absolute = std::env::temp_dir().join("tty7-nu-xdg");
+        let cases: [Option<&std::ffi::OsStr>; 4] = [
+            None,
+            Some(std::ffi::OsStr::new("")),
+            Some(std::ffi::OsStr::new("relative/dir")),
+            Some(xdg_absolute.as_os_str()),
+        ];
+        for xdg in cases {
+            let mut cmd = std::process::Command::new(&nu);
+            cmd.args(["-n", "-c", "$nu.default-config-dir"]);
+            match xdg {
+                Some(v) => cmd.env("XDG_CONFIG_HOME", v),
+                None => cmd.env_remove("XDG_CONFIG_HOME"),
+            };
+            let out = cmd.output().expect("run nu");
+            // A relative value makes nu warn before it answers, so take the
+            // last line, not the whole of stdout.
+            let stdout = String::from_utf8_lossy(&out.stdout);
+            let theirs = PathBuf::from(stdout.lines().last().unwrap_or_default().trim());
+            let ours = nushell_config_dir_from(xdg, platform_config_dir())
+                .expect("this machine has a config dir");
+            assert_eq!(
+                ours, theirs,
+                "tty7 and nu disagree on the config dir for XDG_CONFIG_HOME={xdg:?}"
+            );
+        }
+    }
+
+    /// End to end over a real pty: a user who *has* a config.nu must still have
+    /// it after tty7 points `nu --config` at the wrapper. `--config` replaces
+    /// the user's config rather than adding to it, so a wrapper that fails to
+    /// source it silently strips their prompt, aliases and keybindings.
+    #[cfg(unix)]
+    #[test]
+    fn nushell_restores_the_user_config_and_still_reports_the_prompt_cycle() {
+        let Some(nu) = test_nushell_path() else {
+            eprintln!("skipping: Nushell not installed");
+            return;
+        };
+        let nu = nu.to_string_lossy().into_owned();
+        let home = tempfile::tempdir().expect("tempdir");
+        let user_config = home.path().join("config.nu");
+        // Shaped like a real config.nu: a wholesale `$env.config = {...}` (the
+        // conventional ending, and the one that would wipe hooks added before
+        // it), plus a marker only the user's own file could have set.
+        std::fs::write(
+            &user_config,
+            "$env.config = { show_banner: false }\n$env.TTY7_USER_CONFIG_LOADED = \"yes\"\n",
+        )
+        .expect("write user config");
+
+        let mut injection = setup_nushell_with(Some(&user_config)).expect("nushell integration");
+        // What `setup` does at every spawn boundary, and what this test bypasses
+        // by reaching for `setup_nushell_with` directly. Without it the wrapper
+        // is a no-op whenever the test itself runs inside an integrated pane —
+        // the guard sees the inherited `1` and skips the whole block, and the
+        // marks in the transcript are then Nushell's own, not tty7's.
+        injection
+            .env
+            .insert("TTY7_SHELL_INTEGRATION".to_string(), String::new());
+        let text = prompt_cycle_over_pty(
+            &nu,
+            &injection,
+            // One line, not two: Nushell's line editor holds the tty in raw
+            // mode and drops whatever arrives while a command is running, so a
+            // second `\r` never reaches it and its prompt cycle never happens.
+            // `1 / 0` is the failure — a bare `false` in Nushell is a value.
+            b"print $\"probe=($env.TTY7_USER_CONFIG_LOADED? | default 'MISSING')\"; 1 / 0\r",
+            Some(home.path()),
+        );
+
+        assert!(
+            text.contains("probe=yes"),
+            "the user's own config.nu must survive --config; got:\n{text}"
+        );
+        // Assert on tty7's *own* marks, not on the substring `133;A`: stock
+        // Nushell emits its own OSC 133 cycle alongside these, so a bare
+        // substring passes even when the wrapper never ran. tty7's hooks
+        // terminate with BEL and Nushell's with ST, which tells them apart.
+        for mark in ["133;A\u{7}", "133;C\u{7}", FAILED_COMMAND_MARK] {
+            assert!(
+                text.contains(mark),
+                "Nushell must still report {mark:?} after sourcing the user config; got:\n{text}"
+            );
+        }
+        // Same reasoning for the cwd: the hook hardcodes `localhost`, while
+        // Nushell's native OSC 7 names the real host.
+        assert!(
+            text.contains("7;file://localhost/"),
+            "tty7's own OSC 7 must be among the reports; got:\n{text}"
+        );
+        assert_eq!(
+            last_osc7(&text),
+            home.path().canonicalize().expect("canonical tempdir"),
+            "the pane's reported cwd must be where nu actually is"
+        );
+    }
+
+    /// `shells::nushell_path` is Windows-only (nu has no fixed install path
+    /// there); on unix the binary is whatever `nu` resolves to on PATH.
+    #[cfg(unix)]
+    fn test_nushell_path() -> Option<PathBuf> {
+        std::env::split_paths(&std::env::var_os("PATH")?)
+            .map(|dir| dir.join("nu"))
+            .find(|p| p.is_file())
+    }
+
     #[cfg(windows)]
     #[test]
     fn nushell_reports_the_full_prompt_cycle_over_a_real_pty() {
```

</details>

## Two things about the existing pty test

- **It likely never runs.** `nushell_reports_the_full_prompt_cycle_over_a_real_pty` is `#[cfg(windows)]` and returns early when `nushell_path()` is `None`; CI's Windows runner doesn't have `nu` installed, so it skips. Worth knowing before relying on it as the safety net.
- **The second command doesn't execute.** It writes `cd C:/Windows\r1 / 0\r` in one go. Reedline holds the tty in raw mode and drops what arrives while a command is running, so on macOS the second line never ran and no `133;D;1` was ever emitted. I had to put both on one line (`print ...; 1 / 0`) to get a failing-command cycle at all. Possibly platform-dependent, but worth checking on Windows too.

## Scope question — how much of this does stock Nushell already do?

This isn't a blocker, just something I'd like to settle before 300 lines of injected config becomes ours to maintain. Stock 0.114.1 with **no config at all**, under a pty, already emits:

```
7;file://host/Users   ST     133;A;k=i;click_events=1 ST   133;B ST
133;C ST              2;/tmp BEL                           133;D;1 ST
7;file://host/tmp     ST   <- cwd tracked across `cd`
```

`$env.config.shell_integration` defaults to `{osc2: true, osc7: true, osc8: true, osc9_9: false, osc133: true, osc633: true}`, and tty7 already accepts all of it: `OscTokenizer` finishes on ST as well as BEL (`core/osc.rs`), `identifier_could_match` keys on the id before the first `;` so `133;A;k=i;click_events=1` passes, and `handle_osc133` matches on `rest.first()`. Nushell's own OSC 7 also does the Windows `\`→`/` and leading-slash work (`nu-cli/src/repl.rs`).

With the wrapper active I see both sets of marks in the same transcript — tty7's BEL-terminated ones next to Nushell's ST-terminated ones. That's harmless (A/B/D all set `at_prompt = true` and `OscSniffer::feed` collapses consecutive equal states), but it does mean the net gain over stock nu looks small: mainly the `%`→`%25` escaping in OSC 7 that nu's native path lacks (nu percent-encodes only controls, so a directory named `a%20b` mis-decodes in tty7 today), plus coverage for users who deliberately set `shell_integration` off.

So: is there a case you hit where stock nu *didn't* report, that I'm not reproducing? If the `%25` escaping is the real target, that's a much smaller change and I'd happily take it on its own. Happy to be shown I'm wrong here — you've clearly spent more time in nu than I have.

## Smaller notes, if this moves forward

- `--config` makes `$nu.config-path` the throwaway file, so `config nu` — the standard way Nushell users edit their config — opens a temp file whose edits vanish with the pane. Inherent to `--config` (the zsh path dodges the analogue by restoring `ZDOTDIR` after startup); worth a comment acknowledging the trade-off rather than a fix.
- `str replace -a '\' '/'` runs unconditionally, so a Unix path containing a literal backslash (legal on macOS/Linux) is reported as the wrong cwd. nu's native OSC 7 gates this on `cfg!(windows)`; `$nu.os-info.name == 'windows'` would be the equivalent here.

### What checked out clean

The user config is sourced before the hooks and survives a wholesale `$env.config = {...}`; hook closures propagate `$env` correctly (`__tty7_cmd_active` arms and `hide-env` clears, `133;D;0` then `133;D;1` after a failure, no bogus D on the first prompt); `--config` doesn't disturb `env.nu`, `autoload/*.nu`, or relative `source` inside the user's config; `prompt_indicator` is absent from 0.114's schema so the wrap is correctly skipped; the hooks skeleton works for a config declaring only `env_change`; and the `shell_kind` change is narrowly `"nu"`.

ArnoChenFx:
> Thanks for this — the injection design is sound (sourcing the user's config _before_ appending hooks is exactly right, and the `nu_string_literal` quoting holds up). I reviewed it on macOS with stock Nushell 0.114.1 driven under a real pty, and found one blocker plus a question about scope.
> 
> ## Blocker: on macOS the wrapper discards the user's entire `config.nu`
> `nushell_user_config_path()` resolves non-Windows as `$XDG_CONFIG_HOME/nushell` else `$HOME/.config/nushell`. Nushell doesn't use that on macOS — `nu_path::nu_config_dir()` is `configurable_dir_path("XDG_CONFIG_HOME", dirs::config_dir)`, and `dirs::config_dir()` on macOS is `~/Library/Application Support`:
> 
> ```
> $ nu -n -c '$nu.default-config-dir'
> /Users/thomas/Library/Application Support/nushell
> ```
> 
> So on a stock Mac the lookup misses, the wrapper emits `# no user config.nu to restore`, and `nu --config <throwaway>` then _replaces_ the user's config with one that never sources it. Every macOS Nushell pane loses their prompt, aliases, custom commands and keybindings — silently. Demonstrated with a fake `$HOME` holding `~/Library/Application Support/nushell/config.nu`:
> 
> ```
> plain nu               -> probe=yes
> tty7 wrapper (as-is)   -> probe=MISSING
> ```
> 
> Windows is unaffected: `%APPDATA%` _is_ `dirs::config_dir()` there, so your testing was valid — this is purely the non-Windows arm.
> 
> ## Same root cause, smaller blast radius: `XDG_CONFIG_HOME` on Windows
> `configurable_dir_path` is not `cfg`-gated, so nu honours `XDG_CONFIG_HOME` on **every** platform — but only when it is non-empty **and absolute**. Two consequences:
> 
> * the Windows arm ignores it outright, so a Windows user with cross-platform dotfiles hits the same silent config wipe;
> * the non-Windows arm accepts a _relative_ value, which nu rejects (it warns and falls back).
> 
> Verified against the real binary:
> 
> ```
> default              -> ~/Library/Application Support/nushell
> XDG=/tmp/xdgtest     -> /tmp/xdgtest/nushell        (absolute wins)
> XDG=relative/dir     -> warns, falls back to default
> XDG=""               -> falls back to default
> ```
> 
> ## Patch
> Here's the fix with tests, if it's useful — `cargo fmt` clean, `tty7-core` lib suite green (1142 passed). Take it or reimplement it, whichever you prefer.
> 
> The tests are the part I'd most encourage keeping: one of them asks the real `nu` binary for `$nu.default-config-dir` under four different `XDG_CONFIG_HOME` settings and asserts our resolution agrees, so this class of bug can't come back silently. I confirmed they fail against the current code (`left: /Users/thomas/.config/nushell`, `right: /Users/thomas/Library/Application Support/nushell`).
> 
> Diff against the current head of this branch
> ```diff
> diff --git a/crates/tty7-core/src/daemon/shell_integration.rs b/crates/tty7-core/src/daemon/shell_integration.rs
> index 8804e9f..929ff4c 100644
> --- a/crates/tty7-core/src/daemon/shell_integration.rs
> +++ b/crates/tty7-core/src/daemon/shell_integration.rs
> @@ -873,9 +873,15 @@ fn setup_powershell() -> Option<Injection> {
>  /// throwaway config.nu that sources the user's real one and then appends the
>  /// OSC hooks. The `dir` is what makes the file disappear when the pane closes.
>  fn setup_nushell() -> Option<Injection> {
> +    setup_nushell_with(nushell_user_config_path().as_deref())
> +}
> +
> +/// Split from `setup_nushell` so a test can drive the whole wrapper against a
> +/// user config.nu of its own without writing into the real one.
> +fn setup_nushell_with(user_config: Option<&Path>) -> Option<Injection> {
>      let dir = throwaway_dir("tty7-nu-")?;
>      let config = dir.join("config.nu");
> -    std::fs::write(&config, nushell_config_script()).ok()?;
> +    std::fs::write(&config, nushell_config_script_with(user_config)).ok()?;
>  
>      Some(Injection {
>          env: HashMap::new(),
> @@ -893,10 +899,6 @@ fn setup_nushell() -> Option<Injection> {
>  /// it cannot be guarded at runtime or name a missing file — so the path is
>  /// resolved and substituted here, and a machine without a config.nu gets a
>  /// no-op line instead.
> -fn nushell_config_script() -> String {
> -    nushell_config_script_with(nushell_user_config_path().as_deref())
> -}
> -
>  fn nushell_config_script_with(user_config: Option<&Path>) -> String {
>      let source = match user_config {
>          Some(path) => format!("source {}", nu_string_literal(&path.to_string_lossy())),
> @@ -905,27 +907,62 @@ fn nushell_config_script_with(user_config: Option<&Path>) -> String {
>      NUSHELL_INTEGRATION.replace("__TTY7_SOURCE_USER_CONFIG__", &source)
>  }
>  
> -/// Where nu would load its config.nu from, resolved the way nu resolves it:
> -/// `%APPDATA%\nushell` on Windows, `$XDG_CONFIG_HOME/nushell` (or
> -/// `$HOME/.config/nushell`) elsewhere. Only a file that exists comes back —
> +/// Where nu would load its config.nu from. Only a file that exists comes back —
>  /// the wrapper's `source` cannot name one that is not there.
>  fn nushell_user_config_path() -> Option<PathBuf> {
> -    #[cfg(windows)]
> -    let dir = std::env::var_os("APPDATA")
> -        .map(PathBuf::from)?
> -        .join("nushell");
> -    #[cfg(not(windows))]
> -    let dir = {
> -        let base = match std::env::var_os("XDG_CONFIG_HOME").filter(|v| !v.is_empty()) {
> -            Some(xdg) => PathBuf::from(xdg),
> -            None => PathBuf::from(std::env::var_os("HOME")?).join(".config"),
> -        };
> -        base.join("nushell")
> -    };
> -    let config = dir.join("config.nu");
> +    let config = nushell_config_dir()?.join("config.nu");
>      config.is_file().then_some(config)
>  }
>  
> +/// nu's own config directory, resolved the way nu resolves it — get this wrong
> +/// and the wrapper reports "no config.nu to restore" for a user who has one,
> +/// and `--config` then replaces their config with one that never sources it.
> +///
> +/// nu-path's `configurable_dir_path` consults `$XDG_CONFIG_HOME` on *every*
> +/// platform, Windows included, and only when it is non-empty **and absolute**;
> +/// anything else falls through to `dirs::config_dir()`.
> +fn nushell_config_dir() -> Option<PathBuf> {
> +    nushell_config_dir_from(
> +        std::env::var_os("XDG_CONFIG_HOME").as_deref(),
> +        platform_config_dir(),
> +    )
> +}
> +
> +fn nushell_config_dir_from(
> +    xdg_config_home: Option<&std::ffi::OsStr>,
> +    platform_default: Option<PathBuf>,
> +) -> Option<PathBuf> {
> +    let base = match xdg_config_home {
> +        Some(xdg) if !xdg.is_empty() && Path::new(xdg).is_absolute() => PathBuf::from(xdg),
> +        _ => platform_default?,
> +    };
> +    Some(base.join("nushell"))
> +}
> +
> +/// `dirs::config_dir()`, which nu falls back to. Note this is **not** where
> +/// tty7 keeps its own config: `default_config_dir` is `~/.config/tty7` on macOS
> +/// too, whereas nu follows the platform convention.
> +#[cfg(windows)]
> +fn platform_config_dir() -> Option<PathBuf> {
> +    // `dirs` asks Windows for FOLDERID_RoamingAppData; %APPDATA% is the same
> +    // directory in every environment tty7 can spawn a pane in.
> +    std::env::var_os("APPDATA")
> +        .filter(|v| !v.is_empty())
> +        .map(PathBuf::from)
> +}
> +
> +#[cfg(target_os = "macos")]
> +fn platform_config_dir() -> Option<PathBuf> {
> +    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
> +    Some(PathBuf::from(home).join("Library/Application Support"))
> +}
> +
> +#[cfg(all(unix, not(target_os = "macos")))]
> +fn platform_config_dir() -> Option<PathBuf> {
> +    let home = std::env::var_os("HOME").filter(|h| !h.is_empty())?;
> +    Some(PathBuf::from(home).join(".config"))
> +}
> +
>  /// A Nushell string literal for `path`. Single quotes are fully literal in
>  /// Nushell, so they are the default; a path that contains an apostrophe (legal
>  /// on Windows) falls back to double quotes with backslash escaping.
> @@ -2675,7 +2712,10 @@ mod tests {
>          assert_eq!(inj.args[0], "--config");
>          assert_eq!(inj.args.len(), 2);
>          let written = std::fs::read_to_string(&inj.args[1]).expect("config written");
> -        assert_eq!(written, nushell_config_script());
> +        assert_eq!(
> +            written,
> +            nushell_config_script_with(nushell_user_config_path().as_deref())
> +        );
>          assert!(inj.env.is_empty());
>          assert!(!inj.replaces_argv);
>          let _ = std::fs::remove_dir_all(&dir);
> @@ -2754,6 +2794,177 @@ mod tests {
>          );
>      }
>  
> +    /// nu-path resolves the config dir from `$XDG_CONFIG_HOME` on every
> +    /// platform, but only when it is non-empty *and absolute*. Both halves are
> +    /// load-bearing: the old Windows arm ignored the variable outright, and the
> +    /// old non-Windows arm accepted a relative value nu rejects.
> +    #[test]
> +    fn nushell_config_dir_follows_nu_s_own_resolution_rules() {
> +        use std::ffi::OsStr;
> +        // `/x` is not absolute on Windows (no prefix), so the fixtures have to
> +        // be shaped for the host or the rule under test is never exercised.
> +        let abs_xdg = if cfg!(windows) { r"C:\xdg" } else { "/xdg" };
> +        let platform = || {
> +            Some(PathBuf::from(if cfg!(windows) {
> +                r"C:\platform"
> +            } else {
> +                "/platform"
> +            }))
> +        };
> +        let want = |base: &str| Some(PathBuf::from(base).join("nushell"));
> +
> +        // Unset, empty, or relative: nu ignores it and takes the platform dir.
> +        for ignored in [None, Some(OsStr::new("")), Some(OsStr::new("relative/dir"))] {
> +            assert_eq!(
> +                nushell_config_dir_from(ignored, platform()),
> +                want(&platform().unwrap().to_string_lossy()),
> +                "XDG_CONFIG_HOME {ignored:?} must not displace the platform dir"
> +            );
> +        }
> +
> +        // Non-empty and absolute: it wins, even where there is no platform dir.
> +        assert_eq!(
> +            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), platform()),
> +            want(abs_xdg)
> +        );
> +        assert_eq!(
> +            nushell_config_dir_from(Some(OsStr::new(abs_xdg)), None),
> +            want(abs_xdg)
> +        );
> +        assert_eq!(nushell_config_dir_from(None, None), None);
> +    }
> +
> +    /// The macOS half of the same rule, stated as itself: nu follows the
> +    /// platform convention there, which is *not* where tty7 keeps its own
> +    /// config (`~/.config/tty7`). Reading `~/.config/nushell` on a Mac finds
> +    /// nothing, and the wrapper then replaces a config it thinks is absent.
> +    #[cfg(target_os = "macos")]
> +    #[test]
> +    fn nushell_config_dir_is_application_support_on_macos() {
> +        let home = PathBuf::from(std::env::var_os("HOME").expect("HOME"));
> +        assert_eq!(
> +            platform_config_dir(),
> +            Some(home.join("Library/Application Support"))
> +        );
> +    }
> +
> +    /// Ask the real binary. A hand-copy of nu's resolution rules is only as good
> +    /// as the reading behind it — this is the check that fails outright if the
> +    /// two ever disagree on this machine.
> +    #[cfg(unix)]
> +    #[test]
> +    fn nushell_config_dir_is_what_nu_itself_reports() {
> +        let Some(nu) = test_nushell_path() else {
> +            eprintln!("skipping: Nushell not installed");
> +            return;
> +        };
> +        let xdg_absolute = std::env::temp_dir().join("tty7-nu-xdg");
> +        let cases: [Option<&std::ffi::OsStr>; 4] = [
> +            None,
> +            Some(std::ffi::OsStr::new("")),
> +            Some(std::ffi::OsStr::new("relative/dir")),
> +            Some(xdg_absolute.as_os_str()),
> +        ];
> +        for xdg in cases {
> +            let mut cmd = std::process::Command::new(&nu);
> +            cmd.args(["-n", "-c", "$nu.default-config-dir"]);
> +            match xdg {
> +                Some(v) => cmd.env("XDG_CONFIG_HOME", v),
> +                None => cmd.env_remove("XDG_CONFIG_HOME"),
> +            };
> +            let out = cmd.output().expect("run nu");
> +            // A relative value makes nu warn before it answers, so take the
> +            // last line, not the whole of stdout.
> +            let stdout = String::from_utf8_lossy(&out.stdout);
> +            let theirs = PathBuf::from(stdout.lines().last().unwrap_or_default().trim());
> +            let ours = nushell_config_dir_from(xdg, platform_config_dir())
> +                .expect("this machine has a config dir");
> +            assert_eq!(
> +                ours, theirs,
> +                "tty7 and nu disagree on the config dir for XDG_CONFIG_HOME={xdg:?}"
> +            );
> +        }
> +    }
> +
> +    /// End to end over a real pty: a user who *has* a config.nu must still have
> +    /// it after tty7 points `nu --config` at the wrapper. `--config` replaces
> +    /// the user's config rather than adding to it, so a wrapper that fails to
> +    /// source it silently strips their prompt, aliases and keybindings.
> +    #[cfg(unix)]
> +    #[test]
> +    fn nushell_restores_the_user_config_and_still_reports_the_prompt_cycle() {
> +        let Some(nu) = test_nushell_path() else {
> +            eprintln!("skipping: Nushell not installed");
> +            return;
> +        };
> +        let nu = nu.to_string_lossy().into_owned();
> +        let home = tempfile::tempdir().expect("tempdir");
> +        let user_config = home.path().join("config.nu");
> +        // Shaped like a real config.nu: a wholesale `$env.config = {...}` (the
> +        // conventional ending, and the one that would wipe hooks added before
> +        // it), plus a marker only the user's own file could have set.
> +        std::fs::write(
> +            &user_config,
> +            "$env.config = { show_banner: false }\n$env.TTY7_USER_CONFIG_LOADED = \"yes\"\n",
> +        )
> +        .expect("write user config");
> +
> +        let mut injection = setup_nushell_with(Some(&user_config)).expect("nushell integration");
> +        // What `setup` does at every spawn boundary, and what this test bypasses
> +        // by reaching for `setup_nushell_with` directly. Without it the wrapper
> +        // is a no-op whenever the test itself runs inside an integrated pane —
> +        // the guard sees the inherited `1` and skips the whole block, and the
> +        // marks in the transcript are then Nushell's own, not tty7's.
> +        injection
> +            .env
> +            .insert("TTY7_SHELL_INTEGRATION".to_string(), String::new());
> +        let text = prompt_cycle_over_pty(
> +            &nu,
> +            &injection,
> +            // One line, not two: Nushell's line editor holds the tty in raw
> +            // mode and drops whatever arrives while a command is running, so a
> +            // second `\r` never reaches it and its prompt cycle never happens.
> +            // `1 / 0` is the failure — a bare `false` in Nushell is a value.
> +            b"print $\"probe=($env.TTY7_USER_CONFIG_LOADED? | default 'MISSING')\"; 1 / 0\r",
> +            Some(home.path()),
> +        );
> +
> +        assert!(
> +            text.contains("probe=yes"),
> +            "the user's own config.nu must survive --config; got:\n{text}"
> +        );
> +        // Assert on tty7's *own* marks, not on the substring `133;A`: stock
> +        // Nushell emits its own OSC 133 cycle alongside these, so a bare
> +        // substring passes even when the wrapper never ran. tty7's hooks
> +        // terminate with BEL and Nushell's with ST, which tells them apart.
> +        for mark in ["133;A\u{7}", "133;C\u{7}", FAILED_COMMAND_MARK] {
> +            assert!(
> +                text.contains(mark),
> +                "Nushell must still report {mark:?} after sourcing the user config; got:\n{text}"
> +            );
> +        }
> +        // Same reasoning for the cwd: the hook hardcodes `localhost`, while
> +        // Nushell's native OSC 7 names the real host.
> +        assert!(
> +            text.contains("7;file://localhost/"),
> +            "tty7's own OSC 7 must be among the reports; got:\n{text}"
> +        );
> +        assert_eq!(
> +            last_osc7(&text),
> +            home.path().canonicalize().expect("canonical tempdir"),
> +            "the pane's reported cwd must be where nu actually is"
> +        );
> +    }
> +
> +    /// `shells::nushell_path` is Windows-only (nu has no fixed install path
> +    /// there); on unix the binary is whatever `nu` resolves to on PATH.
> +    #[cfg(unix)]
> +    fn test_nushell_path() -> Option<PathBuf> {
> +        std::env::split_paths(&std::env::var_os("PATH")?)
> +            .map(|dir| dir.join("nu"))
> +            .find(|p| p.is_file())
> +    }
> +
>      #[cfg(windows)]
>      #[test]
>      fn nushell_reports_the_full_prompt_cycle_over_a_real_pty() {
> ```
> 
> ## Two things about the existing pty test
> * **It likely never runs.** `nushell_reports_the_full_prompt_cycle_over_a_real_pty` is `#[cfg(windows)]` and returns early when `nushell_path()` is `None`; CI's Windows runner doesn't have `nu` installed, so it skips. Worth knowing before relying on it as the safety net.
> * **The second command doesn't execute.** It writes `cd C:/Windows\r1 / 0\r` in one go. Reedline holds the tty in raw mode and drops what arrives while a command is running, so on macOS the second line never ran and no `133;D;1` was ever emitted. I had to put both on one line (`print ...; 1 / 0`) to get a failing-command cycle at all. Possibly platform-dependent, but worth checking on Windows too.
> 
> ## Scope question — how much of this does stock Nushell already do?
> This isn't a blocker, just something I'd like to settle before 300 lines of injected config becomes ours to maintain. Stock 0.114.1 with **no config at all**, under a pty, already emits:
> 
> ```
> 7;file://host/Users   ST     133;A;k=i;click_events=1 ST   133;B ST
> 133;C ST              2;/tmp BEL                           133;D;1 ST
> 7;file://host/tmp     ST   <- cwd tracked across `cd`
> ```
> 
> `$env.config.shell_integration` defaults to `{osc2: true, osc7: true, osc8: true, osc9_9: false, osc133: true, osc633: true}`, and tty7 already accepts all of it: `OscTokenizer` finishes on ST as well as BEL (`core/osc.rs`), `identifier_could_match` keys on the id before the first `;` so `133;A;k=i;click_events=1` passes, and `handle_osc133` matches on `rest.first()`. Nushell's own OSC 7 also does the Windows `\`→`/` and leading-slash work (`nu-cli/src/repl.rs`).
> 
> With the wrapper active I see both sets of marks in the same transcript — tty7's BEL-terminated ones next to Nushell's ST-terminated ones. That's harmless (A/B/D all set `at_prompt = true` and `OscSniffer::feed` collapses consecutive equal states), but it does mean the net gain over stock nu looks small: mainly the `%`→`%25` escaping in OSC 7 that nu's native path lacks (nu percent-encodes only controls, so a directory named `a%20b` mis-decodes in tty7 today), plus coverage for users who deliberately set `shell_integration` off.
> 
> So: is there a case you hit where stock nu _didn't_ report, that I'm not reproducing? If the `%25` escaping is the real target, that's a much smaller change and I'd happily take it on its own. Happy to be shown I'm wrong here — you've clearly spent more time in nu than I have.
> 
> ## Smaller notes, if this moves forward
> * `--config` makes `$nu.config-path` the throwaway file, so `config nu` — the standard way Nushell users edit their config — opens a temp file whose edits vanish with the pane. Inherent to `--config` (the zsh path dodges the analogue by restoring `ZDOTDIR` after startup); worth a comment acknowledging the trade-off rather than a fix.
> * `str replace -a '\' '/'` runs unconditionally, so a Unix path containing a literal backslash (legal on macOS/Linux) is reported as the wrong cwd. nu's native OSC 7 gates this on `cfg!(windows)`; `$nu.os-info.name == 'windows'` would be the equivalent here.
> 
> ### What checked out clean
> The user config is sourced before the hooks and survives a wholesale `$env.config = {...}`; hook closures propagate `$env` correctly (`__tty7_cmd_active` arms and `hide-env` clears, `133;D;0` then `133;D;1` after a failure, no bogus D on the first prompt); `--config` doesn't disturb `env.nu`, `autoload/*.nu`, or relative `source` inside the user's config; `prompt_indicator` is absent from 0.114's schema so the wrap is correctly skipped; the hooks skeleton works for a config declaring only `env_change`; and the `shell_kind` change is narrowly `"nu"`.

Fixed!

PR #639, opened 2026-08-14
Retire to the tray on window close; cold start no longer stalls on stale daemon files

ArnoChenFx:
Two problems shared a root: the daemon outlived every window, and nothing could stop it gracefully. This PR fixes the lifecycle on one side and makes the startup path robust on the other.

## Window lifecycle

- **Tray-persist**: closing the last window now retires the app to the tray instead of quitting (`QuitMode::Explicit`). The tray menu restores the most recent workspace and offers one exit path: **Quit**, which — after the existing confirmation that protects running shells — stops the daemon and exits the app. The redundant "Quit and Stop Server" menu item is gone; the palette and keybinding Quit behave identically. No exit path leaves an orphaned daemon behind a dead tray icon.
- **Pathless handoff**: a pathless launch (e.g. double-clicking `tty7-app`) while a GUI is already registered sends `GuiOpen(None)` to it and exits — the windowed instance activates, the windowless (tray) instance restores its workspace. No second process, no second tray icon.
- **Single tray per process**: the tray subsystem (dispatch loop + icon backend) initializes once; rebuilding a window no longer creates a duplicate icon.

## Cold-start robustness

After a force-killed daemon, stale `daemon.port`/`control.port` files made every cold start pay the OS's loopback refusal delay (up to 2 s on some machines) and, on a reused silent port, a 2 s handshake timeout plus a 6 s graceful-stop poll. Now:

- Liveness connects are bounded to **500 ms** (`CONNECT_TIMEOUT`); the version handshake times out in **1 s**.
- An unresponsive daemon is reaped by its recorded pid (bounded) instead of being polled through a 6 s graceful stop.
- A **dead recorded pid skips the TCP probes entirely** (`recorded_daemon_is_dead`): the GUI's `ensure_running`, the new daemon's endpoint check, and the control-listener occupancy check. The last one also had a latent failure mode: a stale control.port whose port was reused by a silent acceptor would make the new daemon refuse to start its control listener and exit.
- Stale cleanup now also removes the leftover `control.port`.

## Verification

Measured on Windows (cold start, no daemon running):

| Scenario | Before | After |
|---|---|---|
| Clean cold start | 0.6 s | ~0.4 s |
| Stale daemon files after force-kill | 4.5 s | **0.4 s** |
| Stale port reused by a silent acceptor | 8.75 s | **0.36 s** |
| Close window → tray → reopen | app quit, daemon orphaned | tray persists, one icon, window restores |

End-to-end scripted checks cover: stale-file cold start, silent-acceptor
cold start, window-close → tray retirement, second-launch handoff, and
tray-icon uniqueness across the full lifecycle.

### DIFF
--- a/crates/tty7-core/src/daemon/server.rs
+++ b/crates/tty7-core/src/daemon/server.rs
@@ -527,7 +527,12 @@ fn report_conpty_host() {
 fn run_with(registry: Arc<Registry>) -> anyhow::Result<()> {
     crate::daemon::control::server_started();
 
-    if transport::endpoint_exists() {
+    // A stale daemon.port whose recorded daemon is gone cannot belong to a
+    // live server: skip the probe (which would pay the OS's refusal delay on
+    // the dead port) and let the bind below overwrite the file. A live
+    // recorded daemon still gets the connect — the singleton seat is held by
+    // this process, so it can only be a foreign server worth refusing.
+    if transport::endpoint_exists() && !crate::daemon::spawn::recorded_daemon_is_dead() {
         match transport::connect() {
             Ok(_) => {
                 anyhow::bail!(
--- a/crates/tty7-core/src/daemon/spawn.rs
+++ b/crates/tty7-core/src/daemon/spawn.rs
@@ -155,10 +155,22 @@ fn is_reapable_daemon_name(name: &str) -> bool {
 /// out the refusal delay on the stale `control.port` that `ensure_running`
 /// clears later. The probe skips itself instead.
 pub fn recorded_daemon_is_dead() -> bool {
-    let Some(pid) = pidfile::read() else {
+    recorded_daemon_is_dead_with(pidfile::read(), daemon_process_alive)
+}
+
+/// The rule alone, so it can be tested without a pidfile on disk — pinning a
+/// config directory would mean an env var, and that is process-global state
+/// every other test running beside this one would inherit.
+///
+/// Note what a missing pidfile answers: **not dead**, because nothing here
+/// knows otherwise. Callers must not read that as "alive" — a refused connect
+/// is still the authority, and `ensure_running` keeps its own stale branch for
+/// exactly that.
+fn recorded_daemon_is_dead_with(recorded: Option<u32>, alive: impl Fn(u32) -> bool) -> bool {
+    let Some(pid) = recorded else {
         return false;
     };
-    pid > 4 && pid != std::process::id() && !daemon_process_alive(pid)
+    pid > 4 && pid != std::process::id() && !alive(pid)
 }
 
 #[cfg(windows)]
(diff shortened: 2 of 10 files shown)


l0ng-ai:
Reviewed this on macOS. The shape is right — the pidfile really is the cheap authority the TCP probes were missing, and the measured numbers hold up. I pushed one commit to this branch (`e966eef`) rather than sending it back and forth; revert or rework it freely if you disagree with any of it.

Three things:

**1. `ensure_running` lost the refused-connect cleanup.** The restructure moved the old `else` under the new liveness check:

```rust
let mut stale = recorded_daemon_is_dead();
if !stale {
    if let Ok(mut stream) = transport::connect() {
        match ... {}
    }
    // connect refused: falls through, stale stays false
} else {
    stale = true;
}
```

So a refused connect no longer reaps or clears `daemon.port` whenever `recorded_daemon_is_dead()` is false — and it is false for two cases the pidfile has no evidence about:

- **no pidfile at all.** `server.rs` writes `daemon.port` in `transport::bind()` and the pidfile ~6 lines later in `pidfile::write_current()`; a daemon that dies in between leaves the port file with no pid to judge it by. Same if the pidfile write ever fails.
- **a recycled pid.** After a force-kill the recorded pid can belong to something unrelated, which reads as alive.

Both then keep the stale endpoint and hand it to the spawn poll below, which pays the refusal delay on every 50 ms iteration — the exact cost this PR removes elsewhere. Restored the branch.

I also split the rule into `recorded_daemon_is_dead_with(recorded, alive)` and added a test for it, mirroring the injection you used for `forward_open_path_with`. It states the trap explicitly: **a missing pidfile answers "not dead", which is not the same as "alive"** — deliberately no config-dir pinning, since that means an env var and every test running in parallel inherits it.

**2. The windowless tray Quit skipped the confirmation.** The comment reasoned that "the confirmation exists to ask whether the panes behind a window should end, and there is no window" — but the confirmation is about shells, not windows. Its own body says *"anything still running in your shells is terminated"*, and this PR's new sentence promises the opposite for the retire path: *"Closing the window only retires the app to the tray; the shells keep running."*

Put together, the windowless state this PR introduces is exactly the one where a full session is running with no window in front of it — and the tray's only exit killed all of it with no prompt. Changed it to restore the window and deliver the action there, so `quit_stop_sessions` prompts as it does everywhere else. Factored the delivery tail into `deliver()` so both paths share it. If the window cannot be opened at all, the bare stop remains — a tray whose only exit is inert would strand the app — but it logs a warning first.

**3. `show_tray_icon` is a request, not an outcome.** `retire_to_tray` read the config alone, but `Backend::create` can fail for the whole run and the loop gives up after `MAX_ATTEMPTS` with just a log line. A Linux session with no StatusNotifier host is the ordinary way there. Config on + backend down + last window closed = a process with no window and no icon: unreachable, and still holding the daemon.

You'd already written the guard for the *config-off* case in the poll loop; this is the same hazard through the other door. Added an `ICON_UP` flag set from `Backend::create`'s actual result, and gated `retire_to_tray` on it.

---

`cargo fmt` clean; `tty7-core` 1143 and `tty7-app` 1538 tests pass locally on macOS.

Two other notes, no action needed:

- The Windows leg was red on the previous head with `send_then_capture_round_trip` (`cli_e2e.rs:606`) — the capture held only pwsh's startup sequence, so the shell was still coming up. A rerun cleared it; it's a flake, not yours. Worth knowing it looks alarmingly like a real regression on a PR that touches spawn and transport.
- I couldn't exercise the Windows-only paths or a real tray-backend failure here, so `CONNECT_TIMEOUT`, the control-listener skip and the measured timings are taken on your testing plus CI.

PR #645, opened 2026-08-15
Cover the whole window with the SSH password prompt's scrim, and name the machine in a failed reconnect

l0ng-ai:
Two fixes to what an SSH connection shows while it is failing: the password prompt now dims the whole window, and a failed reconnect names the machine instead of a config UUID.

| | Before | After |
|---|---|---|
| Password prompt scrim | Stopped at the terminal area — the title bar and tab strip stayed lit, so the prompt read as "only the terminal is busy" | Covers the whole window, like the switcher, the command palette and the worktree prompt |
| Password prompt position | 48px below the top of the terminal area | 120px below the top of the window (`CARD_TOP`), aligned with the other window-level cards |
| Reconnect failure for a profile target | `Reconnecting to java-box… — last failure: could not reach 51d32f65-e669-496b-8f7d-e09cb4991cb6: Broken pipe` | `…could not reach java-box: Broken pipe` |
| Reconnect failure for a deleted profile | Bare UUID | The deleted-profile placeholder, never the UUID (#485) |

## Why

Both are the same class of bug — a detail that leaks out of the layer it belongs to.

The prompt was parented to `body_area`, which is the box below the title bar, so `inset_0` could never reach the chrome. The worktree prompt was hoisted to the window root for this exact reason and left a comment saying so; the SSH prompt was the one that got missed.

The label came from `RemoteTarget`'s `Display`, and `RemoteTarget::Profile` carries nothing but a config UUID — the human name lives in `Config::ssh_profiles`, which `tty7-core` cannot see. The strip's own half of the sentence resolves the name through the live config, so a single line named the same machine two different ways. `remote_connect::target_label` already does that resolution and was added for this in #485; the reconnect path just never called it.

## Testing

- `cargo build -p tty7`, `cargo fmt --check` clean.
- Both symptoms were reported from a manual run of an isolated dev instance (`--config-dir`), on a profile-backed host whose connection dropped mid-session. The fixed build has not been re-checked by hand yet.
- No new tests: both changes are render-tree placement and a call-site swap to a function that already has its own coverage.

### DIFF
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -7292,10 +7292,14 @@ pub(crate) fn new_terminal(
         owner,
         font_size,
     };
+    // The same leak the reconnect banner had: this name is read out as
+    // "Connecting to {machine}…" and "Could not reach {machine}", and a
+    // `Profile` target spells itself as its config UUID (#485). The pane's own
+    // notifications already resolve it through the live config.
     let machine = spawn
         .workspace
         .as_ref()
-        .map(|w| w.target.to_string())
+        .map(|w| crate::ui::remote_connect::target_label(cx, &w.target))
         .unwrap_or_else(|| t(L10nKey::AppLocalServerName).to_string());
     let pending = cx.new(|cx| crate::ui::pending_pane::PendingPane::new(machine, spawn, cx));
     cx.subscribe_in(
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -6474,9 +6474,6 @@ impl Render for Tty7App {
             )
             .child(body)
             .when_some(self.pane_landing(window, cx), |this, el| this.child(el))
-            .when_some(self.render_ssh_prompt_overlay(window, cx), |this, el| {
-                this.child(el)
-            })
             .when_some(ssh_status, |this, el| this.child(el))
             .when_some(self.render_remote_workspace_strip(cx), |this, el| {
                 this.child(el)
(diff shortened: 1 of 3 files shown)


l0ng-ai:
Reviewed both halves and pushed one follow-up fix (9c8fafa).

**Fixed — the same UUID leak, one surface over.** `new_terminal` builds a connecting pane's machine name from `w.target.to_string()` (src/ui/app.rs), which is the bare config UUID for a `Profile` target — exactly what this PR removed from the reconnect banner. That name is read out twice, as "Connecting to {machine}…" and "Could not reach {machine}", so opening a pane on a profile-backed machine that is slow or down showed the UUID in both. It now goes through `remote_connect::target_label`, the same call the pane's own notification title already uses (src/terminal/view.rs).

**Checked and left alone.**
- The scrim's new parent is the window root, so `absolute().inset_0()` resolves against the whole window, and `CARD_TOP` puts it where the switcher, the palette and the worktree prompt sit. The card keeps its own `occlude()`, `track_focus` and Escape handler, so hoisting it changes nothing about focus or key dispatch.
- New paint order: the prompt now draws over the settings overlay and under the switcher/palette. Both read right — the prompt is the thing blocking the connection, and the switcher is what you would reach for to leave.
- The scrim still does not `occlude()`, so clicks reach the tab strip under it. That is deliberate: the overlay hides itself as soon as another pane takes focus, so occluding would leave a prompt in tab 3 trapping the mouse in tab 3.
- `target_label` vs the strip's `route_label` only diverge for a deleted profile, and an unresolvable route is parked as `RouteLost` before `launch_attempt` is ever reached, so the two halves of the sentence cannot disagree in practice.

**Verification:** `cargo fmt --all --check` clean, `cargo check --workspace --all-targets` clean (two pre-existing warnings, both already on main), `cargo test --workspace` all green (1538 + 1160 + 129 + the integration suites, 0 failed). Not re-checked by hand in a running instance — reproducing a password prompt needs a password-auth SSH server.

PR #646, opened 2026-08-15
Restyle the in-app notification to sit in tty7's own visual language

l0ng-ai:
Restyles the in-app notification so it reads as something the OS put on screen rather than a web dialog bolted onto the UI. All the work is in the `gpui-component` fork; this PR is the lock bump that picks it up.

The settings page has no bordered, hard-shadowed box anywhere in it — lists group by whitespace, hierarchy comes from type size and grey. The notification was the one element still built like a Bootstrap alert.

## Layout

| | Before | After |
|---|---|---|
| Status icon | `absolute`, hardcoded `top: 18px`; text column reserved room with `pl_6` | Sibling in the row, in a box one `text_sm` line tall pinned to the top |
| Icon on a wrapped message | Stayed at 18px — visibly off the first line, and CJK wraps early | Centred on the first line at any line count |
| Close button | `absolute` at `top_1 right_1`, sitting on the corner radius | Sibling at the end of the row, mirroring the icon |
| Action button | `mr_3p5` to clear the absolutely positioned close button | No margin needed — the close button is a sibling now |

## Surface

| | Before | After |
|---|---|---|
| Border | 1px drawn border, both themes | Light: hairline shadow (1px spread, no blur). Dark: keeps the real border |
| Shadow | `shadow_md`, then `shadow_xl` | Two soft layers — wide and faint for the lift, tight to seat the edge |
| Radius | `radius_lg` (8px) | 12px |
| Padding | 14 / 16 | 16 / 18 |
| Width | 448px | 400px |

A dark theme keeps the drawn border on purpose: its surface sits close to the app behind it, and a black shadow on a dark ground is invisible, so nothing else would separate the two.

## Type

| | Before | After |
|---|---|---|
| Title | `text_sm` semibold | `text_sm` medium |
| Message under a title | `text_sm`, same ink as the title, no gap | 13px, muted, 4px below the title |
| Message with no title | `text_sm` | Unchanged — full size, full contrast |
| Status icon | 16px | 14px |

Colour alone is a weak rank: two runs at the same size on adjacent lines still read as one paragraph with the back half faded. Size is what says "this is the detail".

The no-title case is deliberately left alone — most tty7 call sites are `window.push_notification(text, cx)`, which sets only a message. There it *is* the text, and shrinking and greying the only thing on the card would be pure loss.

## Why the fork and not a wrapper here

`Notification` implements `Styled` and its render calls `refine_style`, so width, radius, shadow and padding *could* have been overridden from tty7. The layout could not: the icon's absolute positioning, the close button's placement, and the title/message type ranking are all internal. Doing half the change here and half there would have split one visual decision across two repos, and every one of ~30 call sites would have had to opt in.

## Also included

`e31b77b fix(list): make set_query run the search it promises` — unrelated, landed on the fork's `tty7` branch from another change while these commits were in flight.

## Testing

- `cargo check -p gpui-component` after each fork commit.
- `cargo test -p gpui-component notification` — 4 existing tests pass. They cover dismissal/id logic; none of this is layout, which has no assertions in this codebase.
- Manual: isolated dev instance (`--config-dir /tmp/t7v`), Settings → SSH → import from `~/.ssh/config`, which produces a success notification with both a title and a wrapped three-line message. Verified icon-to-first-line alignment, the close button clear of the radius, and the title/message ranking. Light theme only — the dark branch of the border/shadow split is unverified by eye.
- Every build was checked with `strings` against the linked `gpui-component` rev before launching, because 43 worktrees share this repo's `target/` and a stale `tty7-core` artifact from another worktree can make `cargo build` report `Finished` while silently linking old code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3303,7 +3303,7 @@ dependencies = [
 [[package]]
 name = "gpui-component"
 version = "0.5.2"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#00c9a7d3927f461ed35725a64b4eaac2a4ecd23b"
 dependencies = [
  "aho-corasick",
  "anyhow",
@@ -3386,7 +3386,7 @@ dependencies = [
 [[package]]
 name = "gpui-component-assets"
 version = "0.5.1"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#00c9a7d3927f461ed35725a64b4eaac2a4ecd23b"
 dependencies = [
  "anyhow",
  "gpui",
@@ -3400,7 +3400,7 @@ dependencies = [
 [[package]]
 name = "gpui-component-macros"
 version = "0.5.1"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#00c9a7d3927f461ed35725a64b4eaac2a4ecd23b"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3303,7 +3303,7 @@ dependencies = [
 [[package]]
 name = "gpui-component"
 version = "0.5.2"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#2264ff991da93bddfaa0780b5fc070d0ed0a36a5"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
 dependencies = [
  "aho-corasick",
  "anyhow",
@@ -3386,7 +3386,7 @@ dependencies = [
 [[package]]
 name = "gpui-component-assets"
 version = "0.5.1"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#2264ff991da93bddfaa0780b5fc070d0ed0a36a5"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
 dependencies = [
  "anyhow",
  "gpui",
@@ -3400,7 +3400,7 @@ dependencies = [
 [[package]]
 name = "gpui-component-macros"
 version = "0.5.1"
-source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#2264ff991da93bddfaa0780b5fc070d0ed0a36a5"
+source = "git+https://github.com/l0ng-ai/gpui-component?branch=tty7#e48db6d4eb49528958cb70983f9cde6478930a28"
 dependencies = [
  "proc-macro2",
  "quote",


l0ng-ai:
## Review

I reviewed the fork range `2264ff9..e48db6d` (5 commits, 2 files: `crates/ui/src/notification.rs` +74/-27, `crates/ui/src/list/list.rs` +9/-0) rather than just the lock line. The fork's `tty7` head was exactly the pinned SHA, and the lock bump moved only the three `gpui-component` crates — no dependency drag-along. The layout rework is right: the icon and close button as flow siblings in a first-line-height box is a genuine fix for the old `absolute` `top: 18px`, the close button no longer sits on the corner radius, and dropping `mr_3p5` from the action button follows correctly from the close button becoming a sibling. `NotificationList` has no fixed width, so 448 → 400 doesn't strand anything.

### One defect, fixed and re-pinned

**The restyle wrote its new type ranking in pixels, and this app's rem is a user setting.**

`src/ui/app.rs:6383` sets `window.set_rem_size(px(config.ui_font_size))` on every window root — that is what makes the interface font-size setting reach the chrome — and `UI_FONT_SIZE_MIN/MAX` (`crates/tty7-core/src/core/config.rs:1052`) are **12 and 24**. gpui-component does the same thing itself in `root.rs:541`. So `text_sm()` is not 14px; it is anywhere from 10.5px to 21px. Against that, four of the new values froze:

- `first_line_box()` at `h(px(23.))` is the line box of `text_sm` **at rem 16 only**. At rem 24 the first line is ~34px and the icon and close button sit ~5px above its centre; at rem 12 it is ~17px and they sit below. That is precisely the off-centre glyph the box was introduced to prevent, reappearing at a different font size.
- The message at a flat `px(13.)` under a `text_sm` title narrows the rank as the rem grows and **inverts it below rem 15** — the "smaller, quieter detail" set *larger* than the title above it, which is the opposite of the stated intent.
- The status glyph at `px(14.)` stops tracking, even though `Icon`'s own default is the inherited text size (`icon.rs:150`).
- `w(px(400.))` and `px(px(18.))` froze while `py_4()` did not, so at rem 24 the card is padded taller than it is wide, and 400px holds roughly half the characters it was measured to hold.

This is invisible at the default 16, which is why the manual pass didn't catch it — the PR notes that verification was at the default size.

Fixed upstream in `l0ng-ai/gpui-component@00c9a7d`, which rewrites each of those as the same number over 16 in rems. **At the default rem nothing moves at all** — the look that was verified by eye is bit-identical — and everything tracks away from it. Radius and the shadow layers stay in pixels: neither is type, and the radius was already a pixel token. `Cargo.lock` here is re-pinned to `00c9a7d`, hand-edited so the diff is still exactly the three `source` lines.

### Notes, not blockers

- **Radius 12 is now the only 12 in the app.** `src/ui/theme.rs:780` sets `t.radius = px(8.)` and leaves `radius_lg` at gpui-component's 8, so every other surface is 8px. The old code read `cx.theme().radius_lg`; the new one hardcodes 12 and no longer follows the token. Deliberate per the PR table, but it does make the notification the odd surface out.
- **The text column shrinks more than the card does.** The close button used to be `absolute` and cost the text nothing; as a flow sibling it permanently reserves its width plus a `gap_3`, on top of the icon column. Combined with 448 → 400 the usable text width drops by roughly a quarter, so messages wrap earlier than the width table suggests. Arguably still a net win — text no longer runs under a button that appears on hover — but worth knowing.
- **Dark theme is still unverified by eye**, as the PR says. The border/shadow split is the branch with the most design risk and I could not get a screenshot of it in this session (see below).
- **`e31b77b` (`list: make set_query run the search it promises`) is inert here.** The fix reads correctly — `set_value` suppresses `InputEvent` by design, so the search never ran — but nothing in this repo or in the fork calls `List::set_query`, so it ships unexercised.

### Verification

- `cargo check --workspace --all-targets --locked` — clean (exit 0), against the re-pinned `00c9a7d`.
- `cargo fmt --all` — clean, no diff.
- `cargo test -p gpui-component notification` — 4 passed. (They cover dismissal and id logic; none of this is layout.)
- `Cargo.lock` does not drift when cargo re-resolves. Worth flagging: `cargo update -p gpui-component --precise` is **not** the right way to move this pin — it downgrades `windows-sys` 0.61.2 → 0.60.2/0.48.0 and `libloading` 0.8.9 → 0.7.4 across 17 packages. Editing the three `source` lines directly is the clean path.
- No screenshot: the machine is at 91% disk with other builds running, so a full GUI build wasn't worth the risk. The rem fix is provably a no-op at the default size, so the pixel-exact look the PR verified is unchanged.

l0ng-ai:
Correction to the last line of my review: the binary did build fine (`cargo build --bin tty7-app`, exit 0), so disk wasn't the blocker. The actual blocker is that **Screen Recording is not granted to the host tty7** — `CGPreflightScreenCaptureAccess()` returns `false`, which means `screencapture` silently returns bare wallpaper rather than the dev window, and the grant can't be added without quitting the app this session runs inside.

So the visual state is unchanged from what the PR describes: **light theme verified by eye by the author at the default font size; dark theme and every non-default `ui_font_size` unverified by eye.** The rem fix in `00c9a7d` is provably a no-op at rem 16, so it cannot have disturbed the appearance that *was* checked — but if anyone can run a dev instance with Screen Recording granted, the two shots worth taking are a dark-theme notification (the border/shadow split) and a titled notification at `ui_font_size` 24 and 12.

PR #647, opened 2026-08-15
feat(new-tab): reach saved SSH hosts from the New Tab button

l0ng-ai:
Opening a saved SSH host from the main window took the workspace switcher, New Workspace…, a host dropdown, pick, Enter (#438). The New Tab menu — which the `+` already opened — knew about nine shells and nothing else. It now knows about hosts too.

## Before / After

| | Before | After |
|---|---|---|
| Open a saved host from the main window | Workspace switcher → New Workspace… → host dropdown → pick → Enter | New Tab `+` → the host. One click |
| New Tab `+` menu | Installed shells only | Shells, then an **SSH** section: up to six saved hosts by frecency, each with its `user@host:port` underneath |
| A host saved without a name | — | Leads with its address instead of drawing a blank row |
| Reaching the hosts that did not fit | — | **All SSH Hosts…** closes the section and opens the palette with `ssh` already typed *and applied* |
| With no hosts saved at all | — | That row becomes **Add SSH Host…** and opens the host form — a list of hosts is no place to send someone who has none |
| Taking a menu row with ⌥ held | — | Opens beside the current pane instead of in a new tab; the menu says so in its footer |
| ⌘T | New tab | Unchanged — still one keystroke, no menu |

## Why filtering is not in the menu

A `~/.ssh/config` with two dozen hosts is ordinary, so six named rows is not the whole answer — but a search field cannot go inside a `PopupMenu`, which holds the keyboard for its own navigation. The branch that worked around it had to turn the menu into a popover carrying the command palette's own list, and that read far too heavy hanging off a button in the chrome; it was reverted.

So the menu names the few worth naming and hands the rest to the palette, which already filters better than a menu could. Every saved host is a `SSH: {name}` command, so seeding one word is the entire list, frecency-ranked, with the cursor left where the next keystroke narrows it.

## gpui-component bump

`ListState::set_query` wrote through `InputState::set_value`, which deliberately suppresses `InputEvent` so that filling a field from code cannot be mistaken for the user typing and loop back into whatever wrote it. Correct for a plain text field, wrong for this one — the list's search runs off exactly that event, so a seeded query appeared in the box while the list went on showing every row. The one thing the method's name promises was the one thing it did not do.

Fixed in the fork (`l0ng-ai/gpui-component@e31b77b`, branch `tty7`): dispatch the same `Change` the keystroke path would. Only the `Cargo.lock` rev moves here.

## Testing

- `cargo test -p tty7 --bin tty7-app` — 1541 passed, 0 failed.
- New unit tests: the menu stops naming hosts before it becomes a list (`MENU_HOSTS` cap holds with ten saved); a host saved without a name still says where it goes; only ⌥ turns a row into a split.
- `cargo fmt --check` clean.
- Manually verified in an isolated instance (`--config-dir`) seeded with nine hosts: the SSH section lists six, the unnamed host shows `root@build.lan`, **All SSH Hosts…** opens the palette pre-filtered to the SSH group, typing narrows further, and clearing the query returns the full command list.


l0ng-ai:
Reviewed and pushed fixes to `04b952e`. Five things, all inside the feature's own scope.

**The menu could not scroll.** `PopupMenu` only sets `scrollable` for itself on the `OwnedMenuItem` path; a menu built row by row never does, and `max_h` is only applied when it is scrollable. This menu is `1 + shells + 2 + hosts + 1 + 2` rows, and `shells` is whatever the machine has *plus* whatever the user put in `custom_shells` — no ceiling. Past the window height the rows that fall off the bottom are exactly the SSH section this PR adds. Marked it `.scrollable(true)`, which is what `scm/panel.rs` already does for its long menu.

**An unnamed host said the same thing twice.** `menu_hosts` used the endpoint as the display name *and* kept it as the note, so the row rendered `root@build.lan   root@build.lan`. The note is now dropped whenever it would only repeat the name — which also covers a host named after its own address, the shape Quick Connect saves (`profile.name = to_connect_string(&profile)` in `ssh_connect.rs`). Test updated, and a second one added for the Quick Connect shape.

**Long rows pushed instead of cutting.** `menu_row` had a bare label and note with no `min_w_0`/`truncate`. `PopupMenu::max_width` defaults to 500px, so a descriptive host name beside a long `user@host:port` overflows it. Both halves now truncate, the name keeping whatever the endpoint leaves — matching how the sidebar and switcher cut their own labels.

**The rows were assembled on every frame.** `new_tab_button` called `new_tab_menu_rows(cx)` at render time, and the tab strip redraws on every frame a terminal paints. That is `ssh_profiles_by_frecency` — clone every saved profile, sort them — per frame, per `+` button, for a menu nobody had opened. `DropdownMenuPopover` invokes its builder once when the popover opens and again after each dismissal, and hands it an `App`; with the weak handle already in hand there is nowhere it cannot read `Tty7App` from. Moved the read into the builder. Cheaper, and a host saved in Settings while the window sat still is in the list the next time the `+` is pressed.

**Cargo.lock carried sixteen unrelated edges backwards.** Alongside the intended `gpui-component` rev bump, the lock walked `anstream`, `dirs-sys`, `rustix`, `tempfile`, `quinn-udp`, `stacker` and ten others from `windows-sys 0.61.2` back to `0.60.2` (and `d3d12` from `libloading 0.8.9` to `0.7.4`, `winapi-util` to `windows-sys 0.48.0`). Both versions exist on `main`, so this was a stale resolve, not a requirement. Rebuilt the lock from `main`'s and moved only the three `gpui-component` source lines; `cargo check` leaves it alone. The pinned rev `e31b77b` is verified reachable from `l0ng-ai/gpui-component@tty7`.

**i18n.** All four new keys are present in en/ja/zh and all four are used, so `every_key_is_translated_in_every_locale` was already satisfied — but the ja `TabMenuSplitHint` read `{key} を押しながらで分割`, and `〜ながら` cannot take `で`. Now `{key} を押しながら選ぶと分割`, which stays neutral between clicking a row and confirming it from the keyboard. On the zh side, `全部 SSH 主机…` → `所有 SSH 主机…` (`所有` is what this file uses before a noun: `展开所有分组`, `所有内容`), and `按住 {key} 分屏` → `按住 {key} 可分屏打开`, matching the existing `按住 Shift 可让操作保持本地`. CJK lengths are all comfortably inside the 240px minimum.

Left alone deliberately: a profile deleted while the menu is open still just no-ops in `connect_ssh_profile_at`, and the frecency bump still happens before the spawn can fail — both are the pre-existing behaviour of the New Tab path and unchanged by this PR. The `cwd` handed to a split SSH pane is the same one `open_native_ssh_tab` has always passed. `PaletteView::seeded` subscribing after the seed is correct as written: the root list emits nothing that needs handling during construction.

**Verification** (own worktree, own `CARGO_TARGET_DIR`): `cargo fmt --all --check` clean; `cargo check --workspace --all-targets` clean; `cargo test -p tty7 --bin tty7-app` — 1542 passed, 0 failed.

I could not verify this visually. Screen Recording is not granted to the host app on this machine (`CGPreflightScreenCaptureAccess()` is `false`), so every `screencapture` returns bare wallpaper, and System Events/AX times out against the GPUI process, leaving no way to drive or see the menu. An isolated dev instance seeded with nine hosts — including an unnamed one, a very long one, and a CJK-named one — did launch and stay up cleanly on this branch with an empty log. The rendering claims above rest on reading `PopupMenu`'s own layout code, not on seeing it.

l0ng-ai:
Rebased onto `main` after #646 landed. The two PRs both moved the `gpui-component` pin and conflicted on `Cargo.lock`: this branch pinned `e31b77b`, which turned out to be a strict ancestor of the `00c9a7d` that #646 merged — three commits behind, exactly the notification restyle. Merging as-is would have silently reverted it.

Resolved to main's pin, so `Cargo.lock` is now byte-identical to `main` and this PR carries no lock change at all — the rev it wanted is already there.

Re-verified on the new base: `cargo fmt --all --check` clean, `cargo check --workspace --locked` clean with no lock drift, `cargo test -p tty7 --bin tty7-app` 1542 passed / 0 failed.

PR #651, opened 2026-08-15
feat(tabs): drag a tab in as a pane, and a pane out as a tab

### LINKED ISSUE #621: 【建议】一个页面不同分屏支持不同终端/ssh连接
### Type

### What's going on?

建议：
一个页面不同分屏支持不同终端/ssh连接

预期动作：
左侧边栏的会话支持将一个会话拖拽到另一个会话上去，合并会话（合并后同一个页面左右或上下分屏显示不同终端）

背景：
1、调测需要在本地和SSH连接中不断交互，当前没法显示在一个面板中操作，需要不停的切换会话
2、一个项目使用 PowerShell 编码的同时，需要时不时切到 git bash 执行 dev 的 shell 脚本，切会话不如放到一个会话中切换分屏方便

### tty7 version (bugs)

_No response_

### Platform (bugs)

_No response_

### Anything else?

_No response_

l0ng-ai:
A tab can be dragged into another tab's layout to become a pane of it, and a pane can be dragged onto the tab bar to become a tab of its own — closes #621.

## What changes

| | Before | After |
|---|---|---|
| Drag a tab over the panes | Nothing; the strip kept previewing a reorder | The landing lights up over the layout; dropping merges the tab in as a pane and the tab it came from is gone |
| Drag a *split* tab over the panes | — | Its panes arrive still arranged as they were, taking one share of the row or column they joined |
| Drop a tab on a pane's middle | — | Splits that pane along its longer side (a tab has nothing here to trade places with, which is what a pane's middle means) |
| Drag a pane's grip onto the strip or the sidebar | Nothing | A caret shows which two tabs it would go between; dropping gives the pane a tab of its own there |
| Drag the last pane of a tab onto the strip | — | No caret, no drop: it is a tab of its own already |
| Click a tab | Switches on mouse **down** | Switches on the **click** (press + release) |
| Drag a tab to reorder it | The tab became active as it was picked up | Reordering leaves the active tab alone |

Nothing restarts in either direction: the panes are moved, not respawned, so a shell mid-command, an SSH session or a coding agent mid-turn carries straight on.

## Why activation moved to the click

Both surfaces activated a tab on mouse-down, which makes the merge impossible to express: pressing the tab you want to drag puts *it* on screen, so by the time the pointer reaches the panes there is no other tab left to drop it into. gpui drops a pending click when a drag starts, so the press still keeps the title bar from dragging the window, a plain click still switches, and a drag no longer does.

## The machine tree had to follow

Two ops the tree could already express, in an order it could not.

```mermaid
flowchart LR
  subgraph merge["tab dropped into a tab"]
    A["tab A: 1 | 2"] -->|PaneMove 1 → beside 3| B["tab B: 3 | 1"]
    A -->|"PaneMove 2 → beside 1"| C["tab B: 3 | 1 / 2"]
    C --> D["tab A closes itself,<br/>having nothing left"]
  end
  subgraph detach["pane dropped on the strip"]
    E["tab A: 1 | 2"] -->|"PaneClose 2 — tree only"| F["tab A: 1"]
    F -->|"TabCreate, seed = pane 2"| G["tab B: 2"]
  end
```

- **Panes that change tabs are told as `PaneMove`**, one per round, instead of the old tab being closed and a new one rebuilt around its panes. The daemon already handled a cross-tab move, including closing the source tab once it was empty; the client's diff had no way to ask for one.
- **Tabs the machine already has are reconciled before new ones are created.** A pane leaving for a tab of its own is a removal from one tab and an addition to another; the machine refuses to register a pane it already holds, so building the new tab first got the op refused and desynced the window. Dropping a pane to the *left* of its own tab was enough to hit it.

## Testing

Unit tests:

- `pane.rs` — a grafted tab splits the pane it landed on, joins a row as one of its columns, is a band at an edge, and hands its panes back intact when refused.
- `pane_drag.rs` — a tab has no middle to trade places in; a graft's landing is the patch its panes fill between them; which gap on the strip or the sidebar a pane is nearest, and where the caret stands in it.
- `tree_sync.rs` — merging a tab emits the two `PaneMove`s and takes the emptied tab with them; a pane leaving for a tab of its own gives it up before the new tab asks for it.
- Full suite: 1553 pass.

Manual, in an isolated dev instance (`--config-dir`, own daemon):

- Tab → pane: from the sidebar and from the strip, onto a pane's side, its middle, and the tab's outer edge; a split tab dragged in whole; dragging back over the strip to get the reorder instead.
- Pane → tab: onto several gaps in the strip and past the last chip; a round trip out and back with the pane's banner still on screen, unrestarted.
- Click to switch, and double-click on the strip's empty space to maximize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **A tab can be dropped into another tab, as a pane of it** (#621). Drag a tab
+  by its chip or by its sidebar row, out over the panes, and it lands where the
+  highlight says — the same reading as dragging a pane, minus the middle, which
+  for a tab means "split this pane the way it is longest" rather than "trade
+  places". A tab that was itself split arrives with its panes still arranged the
+  way you left them and takes one share of the row or column it joined. Nothing
+  restarts on the way over: a shell mid-command, an SSH session, an agent
+  halfway through a turn all carry on, and only the tab they were in goes away.
+  Picking a tab up no longer switches to it, so the tab you drop into is the one
+  you were already looking at; a plain click still switches.
+
+- **And back out again: a pane dragged onto the tab bar becomes a tab of its
+  own** (#621). Take a pane by its grip up to the strip — or out to the sidebar,
+  wherever the tabs are — and a caret says which two tabs it would go between.
+  The last pane in a tab is offered nothing, being a tab of its own already.
+
 - **Give the prompt back to the shell** — a new **Settings → Input → Prompt →
   Prompt editor** switch (`prompt_editor` in `config.json`, on by default).
   Turned off, tty7 stops editing the shell prompt: every keystroke there —
--- a/docs/window/sidebar.mdx
+++ b/docs/window/sidebar.mdx
@@ -58,6 +58,12 @@ Drag a row to reorder it within its group, or drag a whole group header to move
 the group. A row cannot be dragged into a different group: a tab's group comes
 from its working directory, so `cd` is what moves it.
 
+Drag a row out over the panes instead and it stops being a session of its own:
+it lands as a pane of the tab on screen, wherever the highlight says. Dragging a
+pane the other way — by its grip, onto the sidebar — gives it a row of its own,
+between whichever two the caret lands between. See
+[making one tab a pane of another](/window/tabs-and-splits#making-one-tab-a-pane-of-another).
+
 ## Naming
 
 Almost no tab has a name of its own, so the sidebar falls back:
--- a/docs/window/tabs-and-splits.mdx
+++ b/docs/window/tabs-and-splits.mdx
@@ -82,6 +82,35 @@ Where you drop it decides what happens:
 The landing lights up while you drag, and only ever lights up when the drop
 would actually change the layout.
 
+## Making one tab a pane of another
+
+Two sessions you keep switching between belong in one tab. Drag a tab — by its
+chip in the tab bar, or by its row in the sidebar — out over the panes and drop
+it where you want it: it becomes a pane of the tab you are looking at, and the
+tab it came from is gone.
+
+The landing reads exactly as it does for a pane, minus the middle: a tab has
+nothing here to trade places with, so a pane's middle means "split it the way it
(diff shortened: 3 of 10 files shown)


l0ng-ai:
Reviewed the whole change — the design holds up. Two notes and one fix pushed.

**Fixed (f6bd96f):** moving activation from the press to the release means every
click inside a chip or a sidebar row now reaches the row itself, and
gpui-component's `Button` does not `stop_propagation` on a click it handled. So
one click on a tab's close button ran `close_tab(i)` and *then* `activate(i)` —
`i` by that point naming whichever tab had slid into the closed one's slot, so
closing a background tab moved the active tab somewhere nobody asked for. The
rename field had the same hole: it held the *press* back but not the click, so
clicking into it switched away from the tab being renamed and pulled the focus
out of the input. Both now hold the click where they handle it.

**Checked and fine:**

- `on_click` is primary-button only (`on_aux_click` carries the rest), so the
  chip's context menu still opens without activating.
- `stop_propagation` on the chip's `on_mouse_down` does not eat the click:
  gpui registers the click machinery *after* the user's mouse-down listeners
  and dispatches bubble-phase listeners in reverse, so `pending_mouse_down` is
  recorded before the guard runs. The press still stays off the title bar.
- The `PaneClose` + `TabCreate` pair a detach emits is bookkeeping only —
  `Machine::pane_close` just edits the tree and hands back the ids; only the CLI
  follows it with `hang_up_removed_panes`. Nothing is killed, so the "nothing
  restarts" claim holds on this path too, and the edge-graft fallback in
  `a_tab_grafted_above_a_whole_layout_still_converges` re-registers the same
  pane ids rather than respawning them.
- `migrate_panes` terminates: every round moves one more pane into the tab that
  wants it, and `arriving()` stops naming it once it is there.
- `zone_at` refuses a pointer outside the pane area, so a tab dragged over the
  strip or the sidebar still reorders rather than merging.

**Not blocking, worth knowing:** the strip's drop band is the whole title bar,
so releasing a pane over the New Tab button or the window controls detaches it
at the end of the list. That reads as intended ("after everything"), just wider
than the chips themselves.

The earlier red Windows job was the intermittent `Test`-step hang this repo's
own CI comments document — it passed on a re-run of the same commit.

PR #664, opened 2026-08-16
fix: a correctness and polish sweep across the CLI, daemon, terminal and docs

l0ng-ai:
A correctness and polish sweep across the CLI, daemon, terminal, SSH/SFTP, config and docs — 183 commits, 145 files, +8807/−1943, with no new features and no wire-format changes.

Every commit is a self-contained fix with its reasoning in the message. This body groups them by area so a reviewer can accept them in blocks rather than one at a time.

**Composition:** 93 `fix`, 48 `docs`, 23 `test`, 16 `refactor`, 2 `diag`, 1 `chore`.

---

## CLI output and exit codes (45 commits)

| | Before | After |
|---|---|---|
| `events` rendering a layout delta | `{delta:?}` — a Rust struct literal on a line the docs call an event | All 13 deltas print prose; the catch-all `{other:?}` arm is gone |
| `events` when the server closes the stream | Exited 0, though the contract says it "blocks forever" | Exits 1 + stderr; stdout untouched so NDJSON readers see no novel line |
| `agents` table AGENT column | Lowercased `{:?}`, so `OhMyPi` printed `ohmypi` against a slug of `omp` | Uses `slug()` — one name, not three |
| A long name in any table | Widened every row | Cell is bounded by columns; the last column is exempt |
| A name or error containing escapes | Could drive the reader's terminal | Sanitised at both sites |
| `tab new --cwd <nonexistent>` | Took the flag, never checked it, ran elsewhere | Errors |
| `procs` on a pane that does not exist | Reported an idle pane | Errors |
| `machine ls --json` | Omitted the machine it runs on | Includes it |
| `wait --timeout` expiring | Recommended the flag it was already given | Names the likely cause |
| A refused split / unfiled pane | Left a live pane nothing was holding | Cleaned up |
| `run --keep` | Spawned before checking the inherited workspace | Checks first |
| Failure lines | `1 pane(s)` | `1 pane` |
| Server down | `No such file or directory` | Says the server is down |
| `server start` | Named a pid that had already exited | Names the live one |

## Daemon and server lifecycle

| | Before | After |
|---|---|---|
| `SIGTERM` during startup | Blocked after threads existed, so the shutdown save never ran | Blocked before any thread exists |
| A `kill -9`'d daemon | Left the server unable to start again | Recovers |
| Shell-integration throwaway dirs | Only zsh's were swept; a killed daemon leaked the rest | Every shell's, including after a kill |
| Every CLI call | Logged as a failed read | Quiet |
| A broken config | Said so once per pane spawn | Says it once |
| Log lines | Undated, timezone unstated | Dated, and says the timestamp is UTC |

## Pane lifecycle and tree sync (4 fixes + 2 diagnostics)

| | Before | After |
|---|---|---|
| A layout rewrite | Left orphan shells running | Ends them |
| A pane the tree hands back | Was not hung up | Hung up |
| Ending a pane | Asked one window | Asks every window |
| `kill` on a pane | Could not reach what ran inside it | Reaches it |
| A killed command | Reported as an honest exit 1 | Reported as killed |

## Terminal and rendering safety

| | Before | After |
|---|---|---|
| A filename containing `\n` (`touch $'a\nb'`) | gpui broke the row in two; `text_ellipsis` stopped holding and the rows below slid down | Folded to `a↵b` **at the two draw sites only** — `join`/`rename`/`remove` still get the raw name |
| The same name in the SFTP panel | Same hazard, from a filesystem this window has even less say over | Folded at the entry row and the transfer row; all `remote_join` call sites still use the raw name |
| Eliding a label with `👨‍👩‍👧` / `❤️` / `🇨🇳` | Cut by `char`, so the joiner, the variation selector or one regional indicator was lost | Cuts on grapheme-cluster boundaries in all four functions |
| A terminal title with control characters | Drawn as-is | Folded |
| `1 MiB − 1 byte` | `1024.0K` next to a `1.0M` one byte along | `1023.9K` → `1.0M`, no `1024.0` |
| A tab stop or mode change | Parked the cursor | Does not |
| A bare scheme (`https:`) | Offered as a clickable link | Not offered |
| OSC 52 | Inherited "no clipboard reads" | Policy pinned explicitly |
| PNG decode ceiling | Inherited from the decoder | Owned, and the path is tested |

## SSH and SFTP

| | Before | After |
|---|---|---|
| A typed port of 22 vs an alias with its own `Port` | Alias won | The typed port wins |
| `Include` with a relative path | Resolved against the including file's dir | Resolves against `~/.ssh`, wherever the line sits |
| `Host=name` (the `=` spelling OpenSSH accepts) | Not read | Read |
| A failed shell probe | Disabled integration for good | Recovers |
| SFTP transfer at 99.6% | Printed `100%` beside a row still saying Running | Floored — `100` only when the bytes really are through |
| A finished download | Could replace a file it did not pick | Cannot |

## git, gitignore and SCM

| | Before | After |
|---|---|---|
| `.git/info/exclude` and the global excludes file | Not read | Read |
| A path under an excluded directory | Could be re-included | Cannot (matches git) |
| A repo path containing `\n` | Shifted every answer | Handled |
| Discard / reset / amend confirmations in Chinese or Japanese | `t(Cancel)` was not a cancel answer, so **Escape did nothing** on the dialogs guarding the operations that lose work; Cancel also sat in the position that means "do it" everywhere else | Both go through `confirm_answers`; Escape works in every locale and the positions match the rest of the app |
| A commit-losing op | Inherited the discard question | States what it needs validated and whether it can lose work |

## Config, UI and i18n

| | Before | After |
|---|---|---|
| `sidebar_width: 120` | Accepted by `sanitize`, kept in the file, drawn at 180 — the file said one thing, the window showed another | One range: `SIDEBAR_WIDTH_MIN`/`RIGHT_PANEL_WIDTH_MIN` live in `core::config`, and the two widget constants are *defined from* them. Docs corrected to 180–2000 / 216–2000 |
| The retired `theme` key | Written back into `config.json` | Dropped |
| Keys nothing reads | Silent | Named |
| A whitespace-only workspace/tab name (`ws new "  "`) | Stored; `ws ls` showed a blank cell that was not unnamed | Normalised in `MachineStore`, so both front doors and any future one agree |
| `chevrons-up-down.svg` | Fell through to the stock 2.0 chevron beside 2.1 neighbours | Serves this repo's own; the test now walks `assets/icons` instead of a hand-written list |
| Chinese UI | Called a pane a "panel", called the background process a "daemon" | Corrected |
| Eight i18n keys read by nothing | Present | Deleted with their translations |

## Security

- Three more secrets redacted from `Debug` output.
- CLI names and errors can no longer drive the reader's terminal.
- Four `russh` advisories documented, two of which reach a client pre-auth; the pinned fork predates upstream's fixes and the rebase has nowhere to land yet. **Documented, not fixed** — see `docs(deps)` commits.

## Docs, comments and lint (48 + 1)

A previous comment-stripping pass had removed a large number of invariants and why-notes. These commits restore them, each re-verified against the code rather than reinstated from history: seven cross-file invariants, eight backpressure/probe invariants in `pane`, four each in `view` and `machine`, three wire-dialect contracts in `control`. Several published claims were also corrected — `--json` behaviour that scripts branch on, `wait` idioms that never return, `events`' missing schema and a promised pane-exit event nothing sends.

`chore(lint)` puts clippy on the CI gate and clears the ~200 findings behind it. `e530d5f` gates rustdoc in CI.

---

## Verification status

Verified without a GUI, on this branch:

- Full suite green: `cargo test -p tty7 --bin tty7-app` → **1590 passed, 0 failed**.
- `ws new "  "` → `ws ls` shows `-`; `ws new "  web  "` → `web`. End-to-end against a real daemon in an isolated config dir.
- Panel-width drift is structurally impossible: `MIN_SIDEBAR_WIDTH = core::config::SIDEBAR_WIDTH_MIN`, `MIN_WIDTH = core::config::RIGHT_PANEL_WIDTH_MIN`.
- The `one_line` fold is wired at draw sites only — 2 in `file_tree.rs` (both inside `.child(...)`), 3 in `sftp.rs`; every `remote_join` call site still passes the raw `entry.name`, so no `↵` can reach a path.
- Both SCM confirmations go through `confirm_answers` (`scm/actions.rs:127`, `:316`), and the rule is pinned by a test using the Chinese labels.

**Still wants a human or an agent with screen access.** These are appearance, not correctness — the correctness half is covered above:

1. File tree with `touch $'a\nb'` — how `↵` reads in a row, and that the column stays aligned.
2. A long filename ending in `👨‍👩‍👧` — where the ellipsis lands.
3. Sidebar with `sidebar_width: 120` in the config — the window should draw 180.
4. A discard/reset/amend dialog with `gui_language: zh-CN` — Escape should close it.
5. The tab-strip chevron at 2.1 vs the stock 2.0 stroke (sub-pixel; low value).

## Known issue, not fixed in this PR

`terminal::remote::tests::resize_updates_size_and_notifies_daemon` (`src/terminal/remote.rs:3297`) **fails when run alone** — 3/3 — panicking on the `assert_scratch_config_dir` guard at `core/config.rs:36`. It passes in a full run only because `set_config_dir` is first-wins and a sibling pins the dir first. It is missing the `crate::core::config::pin_test_config_dir()` call that the other nine tests in that module have.

This is the same defect class that `ff6a132` fixed for `ui::sftp`; that commit's claim that "the other test modules were checked and already do" has this one hole. The fix is one line before 3298. Left out deliberately so this PR contains only reviewed commits.

## Testing

- `cargo test -p tty7 --bin tty7-app`: 1590 passed, 0 failed, 1 ignored.
- `cargo test -p tty7-core`: green.
- Workspace-name normalisation verified end-to-end through a real daemon on an isolated config dir, not just in unit tests.
- A dev instance was launched on this branch with `--config-dir`/`TTY7_DATA_DIR` pointed at a scratch dir, seeded with out-of-range panel widths, `gui_language: zh-CN`, and a directory of hazard filenames (embedded newlines, ZWJ emoji, regional indicators). GUI screenshots were not possible in that session — macOS Screen Recording is not granted to the terminal's responsible process — so the five items listed above are the outstanding visual acceptance.


l0ng-ai:
## Reproducing the visual-acceptance setup

For whoever picks up the five outstanding visual items. This seeds every hazard in one launch.

```bash
SCRATCH=/tmp/tty7-accept          # anywhere outside the repo and ~/.config/tty7
mkdir -p "$SCRATCH/devcfg" "$SCRATCH/hazard"

# 1. Hazard filenames: embedded newlines, ZWJ emoji, variation selector,
#    regional indicators, and a long name whose tail is a ZWJ cluster.
cd "$SCRATCH/hazard"
touch $'a\nb.txt' $'report\nQ3\nfinal.md'
touch '👨‍👩‍👧-family.txt' '❤️-heart.txt' '🇨🇳-flag.txt'
touch 'a-very-long-filename-that-should-be-elided-somewhere-in-the-middle-not-inside-a-grapheme-cluster-👨‍👩‍👧-tail.txt'

# 2. Out-of-range panel widths + a non-English locale, in one config.
cat > "$SCRATCH/devcfg/config.json" <<'EOF'
{ "sidebar_width": 120, "right_panel_width": 150, "gui_language": "zh-CN" }
EOF

# 3. Launch isolated, opened straight on the hazard directory.
#    --config-dir is mandatory: without it the dev build attaches to the
#    user's real daemon and sessions. TTY7_DATA_DIR is belt and braces.
cargo build -p tty7
nohup env TTY7_DATA_DIR="$SCRATCH/devcfg" \
  ./target/debug/tty7-app --config-dir "$SCRATCH/devcfg" \
  --open-path "$SCRATCH/hazard" > "$SCRATCH/app.log" 2>&1 &
disown
```

A plain `... &` dies when the shell call returns — the daemon survives, so `ps` still shows a `--daemon` process and it looks like the GUI is up. Use `nohup` + `disown`.

To restart or clean up, scope the kill — never `pkill -x tty7`, which also kills an installed app the user is running:

```bash
pkill -f -- "--config-dir $SCRATCH/devcfg"
```

### What to look at

| # | Where | Correct |
|---|---|---|
| 1 | File tree (already open on the hazard dir) | `a↵b.txt` and `report↵Q3↵final.md` each occupy **one row**; rows below are not pushed down and the column stays aligned |
| 2 | Same panel, the long filename | The ellipsis never splits `👨‍👩‍👧` into loose people or tofu, and `🇨🇳` never degrades to a bare letter |
| 3 | Sidebar and right panel | Drawn at **180** and **216**, not the 120/150 in `config.json` |
| 4 | A tab in a git repo → discard / reset / amend confirmation | UI is Chinese; **Escape closes the dialog**, and the confirming button sits where it does in every other dialog in the app |
| 5 | Tab strip chevron | 2.1 stroke, matching its neighbours rather than the stock 2.0. Sub-pixel — lowest value of the five |

Items 1–4 have their correctness half covered by tests already; what is left is appearance and the end-to-end keypress. Item 5 is appearance only.

### Note on screenshots

On the machine this branch was developed on, macOS Screen Recording is not granted to the terminal's responsible process, so automated screenshots return bare wallpaper with no error — and no repackaging works around it, because TCC attributes every spawned process back to that same app. Two reliable checks that a capture is denied rather than empty: every window's `kCGWindowName` comes back blank except WindowServer's own `Menubar` and `StatusIndicator`, and the capture is always the desktop picture at full size.

If you hit the same wall, the zero-permission path is to ask the user to press ⇧⌘3 — the system screenshot UI carries its own grant and the PNG lands in `~/Desktop`.

l0ng-ai:
Closing in favour of splitting this up.

Three blockers as it stands:

1. **It does not compile on Windows.** `error[E0364]: `CONTROL_PORT_FILE` is only public within the crate, and cannot be re-exported` — 183 commits that never once built on the Windows target.
2. **The clippy gate this PR adds to CI fails on this PR's own code**: `target_scheme` is never used, an unneeded `return`, a collapsible `if`, a redundant closure.
3. **145 files / 183 commits / +8807−1943 in one PR is not reviewable.** The body itself argues for accepting it in blocks — that is the shape of a dozen PRs, not one.

The content is worth having. The SCM confirmation bug in particular (`t(Cancel)` is not a cancel answer in zh/ja, so Escape did nothing on the dialogs guarding work-losing operations) is a real defect, as are the grapheme-boundary truncation and the `1024.0K` formatting.

Suggested split, following this body's own section tables:

1. `chore(lint)` — the clippy gate on its own, plus the Windows visibility fix. Infrastructure first, so every later PR benefits from a green gate.
2. Then by area: CLI output and exit codes / daemon lifecycle / pane lifecycle and tree sync / terminal and rendering safety / SSH and SFTP / git, gitignore and SCM / config, UI and i18n / docs.

Each of those stands alone and can go green on its own.

Issue #580, opened 2026-08-13
Info panel: a remote pane's cwd is abbreviated against the local $HOME


l0ng-ai:
Found while reviewing #570, out of scope there.

`compact_path` abbreviates a pane's cwd with `~` using the **local** machine's `$HOME`, even when the pane lives on another host. A remote pane sitting in `/home/deploy/app` renders as `~/app` on a laptop whose home is `/home/deploy`, and does not when it is `/Users/thomas` — the `~` means the wrong machine either way.

This predates #570, but it is the same class of bug #568 (`d343fd8`) just fixed for file links ("`~` no longer borrows this machine's `$HOME` for a pane whose paths are elsewhere"), so the fix shape probably already exists there. #570 also widened the surface slightly by adding separator normalization and ASCII case folding to the comparison.

PR #680, opened 2026-08-18
fix(windows): normalize path separators before reveal and copy

WhiteDG:
`reveal_path` ("open folder") and copy-to-clipboard on Windows were handed raw paths whose separators `IShellFolder::ParseDisplayName` could not resolve, so opening a folder did nothing and copied paths were not what a shell would type.

A mixed-separator path — a forward-slash prefix joined with backslash entries — reaches `reveal_path` through two routes:

  - **The shell's PWD.** OSC 7 from Git Bash / MSYS bash reports the directory with `/`, and that string survives `Path::ancestors()` when the file tree walks up to find `.git`. So the file-tree root keeps the forward slashes, while the `read_dir` entries underneath come back in native form (backslash).
  - **`git rev-parse --show-toplevel`.** Git for Windows (MSYS2) prints `/` regardless of the calling shell (pwsh included). Both the SCM panel's `scm_repo_root` (`src/ui/scm/panel.rs`) and the worktree creation in `tty7-core` use it, so the root they hand downstream is `/`-prefixed and joins against backslash-joined entries to form paths like `D:/code/tty7\skills`.

Windows' `IShellFolder::ParseDisplayName` rejects that with `E_INVALIDARG` (0x80070057). gpui's `reveal_path` swallows the error (it only logs), so "open folder" silently no-ops. The same mixed-separator strings also make copy-to-clipboard produce paths the user has to retype before a shell will accept them.

## Changes

Add `native_separators(path) -> Cow<Path>` to `src/ui/path_display.rs`:

- On Windows, if the path contains `/`, rewrite every `/` to `\`.
- Otherwise (UNC `\\wsl$\…`, `\\?\…`, already-native path, non-Windows) return `Cow::Borrowed(path)` — **no allocation**.

The rewrite runs on the path's own UTF-16 code units (`OsStrExt::encode_wide` → `OsString::from_wide`), not on a `to_string_lossy` copy. A Windows filename may hold unpaired surrogates, which `to_string_lossy` replaces with `U+FFFD`; a path rebuilt from that names a *different* file, and because `reveal_path` only logs its failures that reads to the user as exactly the silent no-op this PR is removing. `/` and `\` are ASCII, so a code unit equal to either is that character and never half of a surrogate pair — which is what makes the swap safe one unit at a time.

`Cow<'_, Path>` rather than a `String`: `reveal_path` wants a `&Path` and gets one without a lossy step, and the borrowed arm keeps the common case allocation-free.

### Only local paths get re-spelled

`native_separators` answers "how does **this** machine spell a path", so it may only be applied to a path on this machine. A remote host's `/home/u/src` is already native over there; giving it this OS's separators would put `\home\u\src` on the clipboard, which names nothing on either machine. Every call site therefore sits behind the same locality check its Reveal neighbour already used — `paths_are_local` in the file tree, `repo.host == HostId::LOCAL` in the SCM panel, `local_cwd().is_some()` for a pane's cwd.

| Call site | Before | After |
|---|---|---|
| File tree: "Copy path" | Raw separators | Normalized when `paths_are_local` |
| File tree: "Reveal" (open folder) | `reveal_path` got the mixed path, failed silently | Normalized, opens |
| SCM panel: copy path / reveal | Same failure | Normalized when the repo is on `HostId::LOCAL` |
| Info panel cwd: copy / reveal | `copy: cwd.display()`, bare `reveal_path(&cwd)` | Both normalized when the pane's paths are local |
| "Copy working directory" (app menu + tab context menu) | Two inline copies, neither normalized | Both go through `tab_cwd_text`, normalized when local |
| SFTP download → reveal | `reveal_path` got the raw local path | Normalized, opens |

The two "Copy working directory" entries had each spelled the path their own way; folding them into one helper is what stops them drifting apart again.

## Testing

Unit tests in `src/ui/path_display.rs`:

- `native_separators_passes_through_a_path_with_nothing_to_fix` — no `/` → `Cow::Borrowed`, no allocation.
- `native_separators_rewrites_forward_slashes_to_backslashes_on_windows` — `D:/code/tty7\skills` and `D:/code/tty7` both normalized.
- `native_separators_leaves_unc_and_backslash_paths_alone` (windows) — UNC / `\\?\…` / already-native paths pass through borrowed.
- `native_separators_keeps_a_name_a_string_cannot_hold` (windows) — a lone `0xD800` high surrogate survives the rewrite, and the test also asserts the `to_string_lossy` round-trip this replaced really did destroy it.
- `native_separators_is_a_no_op_off_windows` (non-windows) — Unix and backslash-containing paths untouched.

`cargo fmt --all --check` clean. The `#[cfg(windows)]` tests are the Windows CI job's to run.

## Follow-up: the entry points are left as they are

This fix normalizes at the leaves — reveal and clipboard copy — not where the forward slashes enter. That is deliberate, and it does leave work behind: the file-tree root, the dirs handed to `host.watch` (`src/terminal/git_data.rs:944`), the root passed to `git worktree add` (`crates/tty7-core/src/core/worktree.rs`), and the `starts_with` comparison at `worktree.rs:74` all still hold mixed-separator paths, so prefix checks and dedup can still be wrong there — just not user-visible yet.

Normalizing at the entry points is a separate change because both of them are the wrong shape for a `cfg!(windows)` rewrite:

- **`git rev-parse --show-toplevel` (6 sites).** `crates/tty7-core/src/core/git/mod.rs:38` (`probe`), `git/status.rs:835` (`probe_status`), `git/diff.rs:426` (`probe_diff`), `worktree.rs:66` (`managed`), `worktree.rs:111` (`repo_dir`), `src/ui/scm/panel.rs:1147` (`scm_repo_root`) — plus `src/terminal/git_data.rs:1141` for `--git-dir`/`--git-common-dir`. Every one takes `&dyn Host`, and `RemoteHost::git` ships the command to `tty7-server` on another machine. Normalizing there by `cfg!(windows)` would mangle a Linux server's `/home/u/repo` into `\home\u\repo` whenever the *client* is Windows. The right primitive is the host-aware `host.separator()` (what `default_join` already uses), which is a different helper and means auditing all six sites — they also carry three different trim strategies and one is missing `--path-format=absolute`. That is a refactor with its own test surface, not part of this bug fix.
- **OSC 7.** One decode site, `parse_osc7` in `crates/tty7-core/src/daemon/pane.rs:2950`, but it is in the daemon and its result is wire-visible: `PaneState.cwd` → `DaemonMsg::Cwd`, the `ObservedFacts.cwd` / `PaneFacts.cwd` deltas, and `PaneInfo.cwd` on the control API. The forward-slash spelling is also deliberate today — the shell snippet at `daemon/shell_integration.rs:278` emits `pwd -W` output on purpose and `shell_integration.rs:2466` asserts `C:/Users/thoma/repo` survives the round trip. Changing it changes a daemon contract and its guard test, and needs its own compatibility story across client/daemon versions.

The leaves are, in fact, the one place `cfg!(windows)` is the correct question to ask: reveal and copy are guarded to local paths, so the target machine is known to be this one. Pushing normalization up the stack means switching to a host-aware separator first — worth doing, in a change scoped to it.

### DIFF
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -4241,22 +4241,29 @@ impl Tty7App {
         cx.notify();
     }
 
-    pub(crate) fn tab_cwd(
-        &self,
-        index: usize,
-        window: &Window,
-        cx: &App,
-    ) -> Option<std::path::PathBuf> {
-        self.tabs
-            .get(index)?
-            .pane
-            .focused_or_first(window, cx)
-            .and_then(|leaf| leaf.read(cx).cwd())
+    /// A tab's working directory, spelled for the clipboard.
+    ///
+    /// A cwd on this machine is re-spelled with this OS's separators, because
+    /// what goes on the clipboard is meant to be pasted into a shell and a
+    /// mixed-separator path is not one Windows will take. A remote pane's cwd
+    /// keeps the spelling its own machine uses — it is already native over
+    /// there. Both "Copy working directory" entry points (the app-menu action
+    /// and the tab context menu) come through here so the two cannot drift.
+    pub(crate) fn tab_cwd_text(&self, index: usize, window: &Window, cx: &App) -> Option<String> {
+        let leaf = self.tabs.get(index)?.pane.focused_or_first(window, cx)?;
+        let view = leaf.read(cx);
+        let cwd = view.cwd()?;
+        Some(match view.local_cwd().is_some() {
+            true => crate::ui::path_display::native_separators(&cwd)
+                .display()
+                .to_string(),
+            false => cwd.display().to_string(),
+        })
     }
 
     pub(crate) fn copy_active_cwd(&mut self, window: &Window, cx: &mut Context<Self>) {
-        if let Some(cwd) = self.tab_cwd(self.active, window, cx) {
-            cx.write_to_clipboard(gpui::ClipboardItem::new_string(cwd.display().to_string()));
+        if let Some(text) = self.tab_cwd_text(self.active, window, cx) {
+            cx.write_to_clipboard(gpui::ClipboardItem::new_string(text));
         }
     }
 
--- a/src/ui/file_tree.rs
+++ b/src/ui/file_tree.rs
@@ -2050,13 +2050,18 @@ impl Tty7App {
 
         menu = menu.separator().item(
             PopupMenuItem::new(t(L10nKey::FileTreeContextCopyPath)).on_click({
-                let p = p.clone();
+                // Only a path on this machine gets re-spelled: a remote
+                // host's paths are already native over there, and giving
+                // them this OS's separators would copy something that names
+                // nothing on either machine.
+                let text = match paths_are_local {
+                    true => crate::ui::path_display::native_separators(&p)
+                        .display()
+                        .to_string(),
+                    false => p.display().to_string(),
+                };
                 move |_, _window, cx| {
-                    cx.write_to_clipboard(gpui::ClipboardItem::new_string(
-                        crate::ui::path_display::native_separators(&p)
-                            .display()
-                            .to_string(),
-                    ));
+                    cx.write_to_clipboard(gpui::ClipboardItem::new_string(text.clone()));
                 }
             }),
         );
--- a/src/ui/file_tree.rs
+++ b/src/ui/file_tree.rs
@@ -2052,7 +2052,11 @@ impl Tty7App {
             PopupMenuItem::new(t(L10nKey::FileTreeContextCopyPath)).on_click({
                 let p = p.clone();
                 move |_, _window, cx| {
-                    cx.write_to_clipboard(gpui::ClipboardItem::new_string(p.display().to_string()));
+                    cx.write_to_clipboard(gpui::ClipboardItem::new_string(
+                        crate::ui::path_display::native_separators(&p)
+                            .display()
+                            .to_string(),
+                    ));
                 }
             }),
         );
@@ -2061,7 +2065,7 @@ impl Tty7App {
                 PopupMenuItem::new(crate::ui::right_panel::reveal_label()).on_click({
                     let p = p.clone();
                     move |_, _window, cx| {
-                        cx.reveal_path(&p);
+                        cx.reveal_path(&crate::ui::path_display::native_separators(&p));
                     }
                 }),
             );
--- a/src/ui/path_display.rs
+++ b/src/ui/path_display.rs
@@ -69,21 +69,55 @@ fn normalized(s: &str) -> String {
         .to_ascii_lowercase()
 }
 
-/// Normalizes a path's separators to what the local OS expects. On Windows the
-/// shell's `IShellFolder::ParseDisplayName` bails out with `E_INVALIDARG` on a
-/// mixed-separator path — a forward-slash prefix joined with backslash
-/// entries. The forward slashes get in from two routes: the shell's PWD
-/// (OSC 7 from Git Bash / MSYS bash reports `/`, and that string survives
-/// `Path::ancestors()` when the file tree walks up to find `.git`), and
-/// `git rev-parse --show-toplevel` from Git for Windows (MSYS2), which always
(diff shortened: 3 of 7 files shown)


l0ng-ai:
Thanks — the diagnosis here is the good part, and it's right. `git rev-parse --show-toplevel` under Git for Windows and OSC 7 from MSYS shells both report `/`-spelled roots, those get joined with backslash-spelled `read_dir` entries, and `ParseDisplayName` rejects the mixed result with `E_INVALIDARG` — which gpui's `reveal_path` only logs, so the action silently no-ops. That's a real bug and worth fixing.

One thing needs to change before this can land, though.

### Blocking: copy-path is normalized for remote paths too

At all three sites the *reveal* action is gated on locality but *copy* is not, and this PR normalizes both:

- `src/ui/scm/panel.rs:1668` — copy is unconditional; reveal sits behind `repo.host == HostId::LOCAL`, under a comment that reads "a remote repository's paths are not this filesystem's."
- `src/ui/file_tree.rs:2053` — copy is unconditional; reveal sits behind `paths_are_local`.
- `src/ui/right_panel.rs:689` — `copy:` is built unconditionally from `effective_cwd()`; `reveal:` is `local_cwd().is_some().then(...)`.

Concrete case: tty7 on Windows, SCM panel open on a remote Linux repo rooted at `/home/deploy/app`. Right-click a changed file → Copy Path. Before this PR the clipboard got `/home/deploy/app/src/main.rs`; after it, `\home\deploy\app\src\main.rs`, which no remote shell will take. Same for the file tree over a remote host and for the Info-panel cwd of an ssh'd pane.

The predicate you want is "this path lives on a Windows machine", not "this process runs on Windows" — which is the lesson `path_display.rs`'s own module doc records from #580/#568 ("only the caller knows which machine the path is on"). Either gate the three copy sites on the locality boolean the neighbouring reveal already uses, or thread the host into `native_separators`. A test pinning that a remote POSIX path comes back untouched would be good too — right now there's no way to express that, since the function has no host parameter.

### Blocking: rustfmt

One hunk, `src/ui/path_display.rs:235-241`:

```diff
-        for p in ["\\\\wsl$\\Ubuntu\\home", "\\\\?\\C:\\code", "C:\\code\\tty7"] {
+        for p in [
+            "\\\\wsl$\\Ubuntu\\home",
+            "\\\\?\\C:\\code",
+            "C:\\code\\tty7",
+        ] {
             let got = native_separators(Path::new(p));
             assert_eq!(got.as_ref(), Path::new(p), "{p:?}");
-            assert!(
-                matches!(got, Cow::Borrowed(_)),
-                "{p:?} should not allocate"
-            );
+            assert!(matches!(got, Cow::Borrowed(_)), "{p:?} should not allocate");
         }
```

Both edits are inside the `#[cfg(windows)]` test — rustfmt is purely syntactic and formats cfg'd-out code regardless, so this reproduces on macOS too if you're checking there.

### Suggestion: drop the lossy round-trip

`src/ui/path_display.rs:81-86` goes through `to_string_lossy()`. On Windows an `OsStr` is WTF-8, so a filename with an unpaired surrogate becomes U+FFFD and the resulting `PathBuf` names a different, nonexistent file. It needs the path to *also* contain a `/` to bite, so it's rare — but `path.components().collect::<PathBuf>()` re-spells with the native separator losslessly over `OsStr`, handles `\\?\` and `\\server\share` correctly by construction, and lets the `contains('/')` guard and the double `to_string_lossy()` go away.

### Minor

- Two other "copy the pane cwd" sites are untouched and will now be inconsistent with the right panel: `src/ui/app.rs:4259` (`copy_active_cwd`, backing `CopyWorkingDirectory`) and `src/ui/tab_strip.rs:1443` (tab context menu). They're currently correct-by-omission for the remote case, but wrong for a local Windows pane whose OSC 7 reports `/` — the exact scenario this PR is about.
- `src/ui/path_display.rs:14-16` says "Nothing feeds it back to an API: the Info panel's Copy Path and Reveal both carry the untouched `PathBuf`." This PR makes both halves false.

The four tests are the right ones — asserting borrow-vs-own rather than just equality is the detail I'd have asked for. Happy to take this once the copy sites are gated.

Issue #672, opened 2026-08-17
tty7 lost all of my tabs!


Infiniverse:
### Type

Bug — something doesn't work as expected

### What's going on?

I've just restarted the server after an upgrade, and I've been presented with an empty workspace - none of my tabs has been recovered.

Useful context:

- I had created my tabs using v27.8.2, then upgraded to v27.8.3. Tabs had been restored, but the server was still running v.27.8.2
- I had renamed my workspace from the default to 'Ignis'. 

I thought that this was a key feature. Was it something to do with the renaming?

### tty7 version (bugs)

v26.8.3 (server was v27.8.2)

### Platform (bugs)

macOS (Apple Silicon)

l0ng-ai:
Fixed on `main` — but not in any released build yet, so please don't upgrade expecting it to be gone.

This turned out to be three separate ways the tab set could get torn down, closed in stages:

- #564 — when a restart handoff fails, the window is now rebuilt from the machine tree instead of only recording a startup error.
- #608 — an abandoned `Adopt::Replace` debt no longer deletes the very tabs it was in the middle of pulling in.
- #681 — tabs a partial rebuild could not put up are now held back, so the next full sync stops reading them as "the user closed these" and sending `TabClose` for them.

On your hunch about the rename: that isn't it. A workspace is identified by a UUID and restore selects by that id — renaming writes `ws.name` and nothing else, and never touches the restore path.

All three are on `main` and will go out in the next release. v26.8.3, which is what you were running, still has the old code — so if you reproduce this on v26.8.3 that's expected, not a regression. Closing as fixed; please reopen if it happens on a build newer than the next release.

Issue #579, opened 2026-08-13
tree-sync: an abandoned Replace debt still lets a resync delete live pane records


l0ng-ai:
Found while reviewing #564 and #569 — reported independently by both reviews, so filing it on its own.

`take_rehydrate` (`src/ui/tree_sync.rs:839-857`) **abandons** an `Adopt::Replace` debt whenever the window already has tabs. That is the common case in exactly the scenarios the debt exists for:

- A window whose tabs are stale/dead (a refused handoff, a daemon that came back as a new process) always has tabs — the dead ones are the whole point.
- If the tree pull fails and the user then opens a tab themselves, the `Replace` debt is dropped, `start_prime` re-primes the real mirror, and the next `SyncScope::Full` diffs that one new tab against it — emitting `TabClose` for every other tab, which deletes those panes' records from the machine tree. The processes keep running and the GUI can never reach them again.

This is the same damage #554 describes, reached through a different door, and it applies to remote windows identically. It is pre-existing behaviour from #472 (`af3928d`, refined by `e4bd49c`), not introduced by either PR, which is why neither fix touched it — changing the rule affects the remote resyncs that share it.

Worth deciding deliberately: should a `Replace` debt survive the window having tabs, and if so, how does it avoid fighting a user who is actively opening tabs while the pull is out?

PR #681, opened 2026-08-18
fix(restart): fail a silent Attach, and hold the tabs a rebuild could not put up (#673)

### LINKED ISSUE #673: Restarting tty7 doesn't resume my sessions (they are all hanging)
### Type

### What's going on?

I have restarted tty7, but none of my existing Claude Code sessions has restarted properly. 

Image

They are all locked on resume.

I've tried resizing the window, in case its a termcap issue, but no luck.

Backgrounding isn't returning me to the underlying shell:

    Claude Code has been suspended. Run `fg` to bring Claude Code back.
    Note: ctrl + z now suspends Claude Code, ctrl + _ undoes input.
    ^Z
    ^Z

Control-C is not having any effect either. All terminals appear locked.

### tty7 version (bugs)

v26.8.4-nightly-202608161107

### Platform (bugs)

macOS (Apple Silicon)

### Anything else?

_No response_

biztex:
Fixes #673. Also closes what is left of #672 at HEAD (the rest was #554 and #579).

## The bug

A restart on nightly 26.8.4 came back with every restored coding-agent pane locked: Ctrl-Z printed its suspended message and never returned to a shell, Ctrl-C did nothing, no keystroke reached anything. Both this and #672's empty workspace are the same mistake — a restart's rebuild reporting a success it did not have.

**The locked panes are an `Attach` the client took on trust.** `attach_reply_prefix` reads far enough into the daemon's reply to tell an `Error` frame from a replay, and a read that timed out with *nothing* in the buffer fell through to the success branch: silence was read as "a quiet pane". But a quiet pane is never silent. `attach_subscriber` replays the pane's ring before the daemon reads a byte of our input, the ring always holds a segment, and every daemon build there has been queues a `Size` first — a pane that has printed nothing still answers with its geometry. So an `Attach` that produced no bytes in the whole wait is one nobody is serving: a daemon still mid-restart, or a socket some process holds open and will never read. Taken for an attach, it made `spawn_shell_terminal_in` report `restored = true` — the flag that skips the fresh spawn, the restored-screen banner and the agent's `--resume` — and `write` threw every encode error away, so Ctrl-C and Ctrl-Z went into that socket and vanished.

**The tabs that did not come back** are the rebuild's licence outrunning what it rebuilt. `tabs_from_session` drops any tab none of whose panes would start; `settle_hydration` then marked the window `informed` as long as *some* tab rebuilt, while the mirror still listed every tab. The next `sync_window` ran at `SyncScope::Full`, whose `diff` emits `TabClose` for every mirror tab not in `desired` — exactly the tabs that had failed to rebuild. A partial rebuild deleted from the machine the tabs it could not put up, panes and all.

## The fix

Zero bytes within the wait is now the failure it is, and the caller falls through to the path it already had for a pane that is gone — a fresh shell under the old screen, with the resume typed. Nothing changes on the wire. Trade-off stated plainly: a *slow* but real attach now costs a fresh shell over its screen and an orphaned shell in the daemon (the orphan sweep reports it), where before it cost a pane nobody could type into. `write` also stops swallowing a refused `Input`: logged once, and unless the reader was retired for a relink the pane is marked exited by the reader's own signal, so the window shows it as gone.

The un-rebuilt tabs are **held** rather than the licence withheld: `settle_rebuild` records the wanted ids the window is not showing, and `sync_window` carries them into `held`, whose contract in `diff` is already "mirror tabs the window cannot speak for — close nothing, don't reorder around them". Withholding the licence would take `TabClose` away from the whole window for as long as the failure stands (a tab whose shell is no longer on the machine fails every restart), resurrecting every close the user made in the meantime. The set is rewritten by the next rebuild and pruned against the mirror on every sync. The none-rebuilt guard is unchanged.

## Tests

`cargo test -p tty7` 1573 passed, `-p tty7-core` 1154 passed. New in `remote.rs`: `an_attach_nobody_answers_is_not_an_attach`, `a_size_frame_alone_is_a_live_attach`, `a_write_the_link_refuses_marks_the_pane_gone_once`, `a_write_on_a_retired_link_does_not_mark_the_pane_gone`. New in `tree_sync.rs`: `a_partial_rebuild_holds_the_tabs_it_could_not_put_up` and `the_next_sync_leaves_a_tab_the_rebuild_could_not_put_up_on_the_machine` (verified to fail with the `held.extend` line removed). Not reproduced live: the macOS mid-handoff race that makes the daemon silent for >2 s — the client fix is correct regardless of its cause.

### DIFF
--- a/src/terminal/mod.rs
+++ b/src/terminal/mod.rs
@@ -29,5 +29,5 @@ mod typeahead;
 pub mod view;
 
 pub(crate) use remote::notify_desktop;
-pub use remote::{PaneRoute, PaneWorkspace, RemoteTerminal};
+pub use remote::{PaneRoute, PaneWorkspace, RemoteTerminal, attach_unanswered};
 pub use size::TermSize;
--- a/src/terminal/remote.rs
+++ b/src/terminal/remote.rs
@@ -219,6 +219,10 @@ pub struct RemoteTerminal {
     /// flag under the term lock before every grid mutation, so once it is set
     /// the abandoned thread can only exit, never write.
     reader_quit: Arc<AtomicBool>,
+    /// Set by the first `Input` the link refused, so the loss is said once
+    /// rather than once per keystroke. Cleared when a relink installs a link
+    /// that has not refused anything yet.
+    input_lost: AtomicBool,
 }
 
 /// The workspace id a spawn carries, so the pane's shell gets `$TTY7_WS` and a
@@ -405,7 +409,36 @@ impl RemoteTerminal {
         let win = win_size(size, cell_w, cell_h);
 
         ClientMsg::Attach { pane_id, size: win }.encode(&mut stream)?;
-        let buffered = attach_reply_prefix(&mut stream, pane_id, attach_reply_wait(route))?;
+        let buffered = match attach_reply_prefix(&mut stream, pane_id, attach_reply_wait(route)) {
+            Ok(buffered) => buffered,
+            Err(e) if route.is_local() && attach_unanswered(&e) => {
+                // Silence on the attach socket has two readings, and only one
+                // of them is safe to act on. Ask the daemon on a fresh
+                // connection: if it answers `Version` there, it is up and
+                // serving, and the attach connection is one it will never
+                // serve — a socket some process holds open, or one from a
+                // listener no longer drained — so the verdict stands and the
+                // caller spawns fresh. If it does not answer there either, it
+                // is not serving anyone yet — mid-restart, mid-handoff — and
+                // a fresh pane spawned now would land on a live one the moment
+                // it comes up (its history carried across, its agent session
+                // resumed twice). There is no third path from a synchronous
+                // UI-thread call, so the attach still fails, but says which
+                // silence it was: the log line someone reads while diagnosing
+                // an orphaned shell must not claim the pane was gone.
+                //
+                // Only local routes probe: a remote attach already waits 15 s
+                // and a second routed connection is a second bridge process.
+                if local_daemon_answers() {
+                    return Err(e);
+                }
+                return Err(e.context(
+                    "the daemon answered nothing on a fresh connection either — it is not \
+                     serving yet (restarting?), so this pane may still be alive",
+                ));
+            }
+            Err(e) => return Err(e),
+        };
         let mut term = Self::from_stream_with(stream, size, buffered)?;
         term.route = route.clone();
         Ok(term)
@@ -480,6 +513,7 @@ impl RemoteTerminal {
         }
         self.reader_thread = Some(reader);
         self.reader_quit = quit;
+        self.input_lost.store(false, Ordering::SeqCst);
         self.route = route.clone();
         self.synced_size = false;
         self.resize(size, cell_w, cell_h);
@@ -577,6 +611,7 @@ impl RemoteTerminal {
             proxy,
             reader_thread: Some(reader_thread),
             reader_quit,
+            input_lost: AtomicBool::new(false),
         })
     }
 
@@ -1094,9 +1129,40 @@ impl RemoteTerminal {
         if bytes.is_empty() {
             return;
         }
-        if let Ok(mut writer) = self.writer.lock() {
-            let _ = ClientMsg::Input(bytes.into_owned()).encode(&mut *writer);
+        let Ok(mut writer) = self.writer.lock() else {
+            return;
+        };
+        if let Err(e) = ClientMsg::Input(bytes.into_owned()).encode(&mut *writer) {
+            drop(writer);
+            self.note_input_lost(&e);
         }
+    }
+
+    /// The link refused an `Input`. Every keystroke after the first would say
+    /// the same thing, so this side says it once; and unless the reader has
+    /// been retired for a relink or a release, the pane is marked exited the
+    /// way the reader marks it on EOF — it is the same socket, noticed from the
+    /// writing side first — so the window shows the pane as gone instead of
+    /// taking input into it that nothing will ever read. The reader still
+    /// raises its own `Exit` when it finds the same socket closed; the handler
+    /// is idempotent, so a link that is genuinely gone may be reported twice.
+    ///
+    /// This is hardening for a *closed* link, not the cure for #673: a socket
+    /// some process holds open and never reads accepts writes into its send
+    /// buffer, so they succeed and vanish until the buffer fills, and nothing
+    /// here fires. What stops that pane existing at all is `attach_on`
+    /// refusing to call a silent `Attach` attached.
+    fn note_input_lost(&self, err: &std::io::Error) {
+        if self.input_lost.swap(true, Ordering::SeqCst) {
+            return;
+        }
+        log::warn!("the daemon link stopped taking this pane's input: {err}");
+        if self.reader_quit.load(Ordering::SeqCst) {
+            return;
+        }
+        self.exited_flag.store(true, Ordering::SeqCst);
+        self.proxy.send_event(AlacEvent::Wakeup);
+        self.proxy.send_event(AlacEvent::Exit);
     }
 
     /// Whether the daemon behind this pane echoes a `DaemonMsg::Size` into the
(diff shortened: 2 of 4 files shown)


l0ng-ai:
Both bugs are real and the #673 diagnosis is well argued — I checked its central premise against the daemon and it holds: `ReplayRing::new` always seeds one segment (`crates/tty7-core/src/daemon/pane.rs:2300`), `seeded` falls back to it when empty, and `replay` emits a `Size` per segment (`pane.rs:2407`), so a quiet pane really does answer and silence really is anomalous. Version skew isn't a concern either — remote routing and Size-first replay landed together in 208454e2.

A few things before this lands.

### Blocking: the two new `write` tests only pass by luck of ordering

`src/terminal/remote.rs:3252` and `:3286` go through `RemoteTerminal::from_stream` → `Config::load()` (`remote.rs:508`), but neither calls `crate::core::config::pin_test_config_dir()`. The guard at `src/core/config.rs:36` then panics with *"would touch the real user config dir"* unless some other test in the same process pinned it first.

```
cargo test -p tty7 --bin tty7-app -- tree_sync:: remote::
```

fails both. Adding the one-line pin to each fixes it. CI is green only because a full run happens to pin first. (`agent_session_follows_daemon_status_reports` has the same gap, so this is following local precedent — but it adds two more instances.)

### Blocking: keep the none-rebuilt guard on the unfiltered count

`src/ui/tree_sync.rs:2096` — `filter_map(|t| t.tree_id)`. `SessionTab::tree_id` is `#[serde(skip)]` (`crates/tty7-core/src/core/session.rs:56`), so a disk-loaded `Session` has `None` on every tab. Today the only real caller passes `session_from_tree` output (`tree_sync.rs:1519`, always `Some`) so the guard is genuinely unchanged as you say — but if a disk session ever reaches this path, `wanted.is_empty()` goes true and a fully-failed rebuild gets the prune licence, which is #672 again. Keep the guard on `session.tabs.len()` and use the filtered vec only for the held set, or `debug_assert_eq!` the two lengths.

### Question: the 2 s local budget vs. the execve handoff

This is the part I'd most like your read on. Previously a silent Attach returned `Ok(vec![])` and `from_stream_with` handed the *same still-open socket* to the reader thread — so a daemon that was merely slow to accept (the execve handoff from #669 keeps the listener and its backlog across the exec) would serve the connection a moment later and the pane came back on its own. Now the socket is dropped at the budget and a fresh pane is spawned over a live one, which means:

- `spawn_on(..., RestoreFrom { pane_id })` makes the daemon run `history::carry(dead, id)` (`daemon/server.rs:705`), which **renames** the live pane's `$HISTFILE`, `.seed` and `.origin` onto the new pane — `daemon/history.rs:93` is explicit that it's for a *dead* pane. `restored_screen` also calls `scrollback::forget(pane_id)`.
- `restored == false` makes `session_to_pane` fire `agent_resume_command` (`src/ui/app.rs:7716`), so `claude --resume <id>` starts against a session the orphaned process still holds.

The orphan is then only reachable as a Close button in the switcher (`src/ui/switcher.rs:2008`); there's no re-adopt path.

I could **not** reproduce a >2 s handoff window, so this is unverified and may be purely theoretical — it depends on daemon init time on slower machines. But note the budget is only ever *paid* by a silent connection (a dead pane answers `Error` immediately, a live one answers `Size` immediately), so raising it costs nothing in the common case. Probing once with `ClientMsg::Version` on a fresh connection and only calling silence fatal if the daemon answers there would be stronger still. The counterweight is that local attaches run synchronously on the UI thread (`session_to_pane` → `new_terminal` → `spawn_shell_terminal_in` inside `handle.update`), so N panes × budget is frozen-window time. That tension deserves a sentence in the code, not just in the PR body.

### Please add a comment: the held set has no retry

`tree_sync.rs:318` — `diff` returns before the `TabMove` pass and the `WorkspaceSetActiveTab` op whenever `held` is non-empty, and `sync_window:1002` now feeds `not_rebuilt` into `held`. `not_rebuilt` is rewritten only by another `settle_rebuild`, which runs only from `finish_hydration`; a `desync`/re-prime goes through `start_prime`/`finish_prime` and never touches it, and every `hydrate(Adopt::IfEmpty)` short-circuits at `tree_sync.rs:2065` on a non-empty window. So: restart with 5 tabs, 1 fails → the user reorders tabs and works in tab 3 → none of that reaches the machine tree → the next restart restores the pre-drag order and the old active tab, for the rest of the session.

This state was already reachable (a `PendingPane` whose remote spawn failed stays `Connecting` forever, `app.rs:3074`), so the PR widens an existing hole rather than opening one — I'm not asking for a retry mechanism in this PR. But it should say out loud that the held set has no retry and that ordering/active-tab sync is suppressed while it stands.

### Minor

- The `write` half doesn't actually cover #673's reported symptom. `note_input_lost` fires only when `encode` *errors*; in the case the PR describes — a socket held open by a process that will never read — writes land in the send buffer and succeed, so keystrokes still vanish silently until ~64 KB accumulates, at which point a blocking `UnixStream::write` on the UI thread freezes the window rather than returning an error. The test shuts the write half, i.e. the closed case, not the reported one. The attach change is the real fix here; the write change is defensive hardening, and the PR body reads as if Ctrl-C-into-the-void is now caught.
- `src/terminal/view.rs:1012`: `"pane {id} is gone on its machine ({e:#}); spawning fresh"` is wrong for the timeout case — the pane may well still be there, and this is the line someone will read while diagnosing an orphan.
- "said once" is narrower than it reads: `remote.rs:1118` guards the writer's report, but the reader's `teardown` (`remote.rs:688`) raises its own `Exit` for the same dead socket, so a genuinely closed link still reports twice in production. Harmless — the handler is idempotent — and the test's `exits == 1` holds only because the peer socket stays open.
- `adopt_workspace` still pushes `AppTabsNotRestored` (`app.rs:1510`), which now understates things: the tabs are kept on the machine.
- Possible follow-up, pre-existing: `tabs_from_session` computes `active = session.active.min(len-1)` (`app.rs:7653`) where `session.active` indexes the *original* list, so dropping a tab ahead of the active one lands focus on the wrong tab. Not introduced here, but a partial rebuild is now a state you're meant to live in rather than one that self-corrects, so it's more visible.

### Checked and clean

No new panic path (the one `expect` in `diff` at `tree_sync.rs:327` sits behind `held.is_empty()`, and the change only makes `held` larger). A failed attach closes nothing — `note_input_lost` sets `exited_flag` but not `child_exited`, so `ChildExited` doesn't fire and the daemon pane survives; deliberate and correct. No double-add on a later successful rebuild. `adopt_relink` resets both flags and drains the event channel. The other four new tests pass.

biztex:
Thanks — this is a very useful review, and the `Version`-probe suggestion is better than anything I had.

Pushed as `aabcb9e`:

- **Both blockers fixed.** `pin_test_config_dir()` on both `write` tests (your `-- tree_sync:: remote::` filter now passes; it did fail before). The none-rebuilt guard is back on `session.tabs.len()`; the `tree_id` vec feeds only the held set, with a `debug_assert_eq!` on the lengths and a test case pinning that a session with tabs but no ids does *not* get the licence.
- **The 2 s budget.** I kept it — `a_local_attach_does_not_wait_as_long_as_a_remote_one` already pins it as a UI-thread bound, and as you say N silent panes pay it N times. Instead I took your probe: a silent local Attach is confirmed by asking `Version` on a fresh connection. If the daemon answers, it's up and the attach socket is one it will never serve, so the verdict stands. If it doesn't answer either, nobody is serving yet — I don't have a third path from a synchronous UI-thread call, so the attach still fails, but the error and the `view.rs` log now say "the daemon is not serving yet — this pane may still be alive" rather than "gone", since that is the line someone reads while diagnosing an orphan. Only local routes probe (a remote attach already waits 15 s and a second routed connection is a second bridge). The tension is now in a comment on `attach_reply_wait`, and there's a typed `AttachUnanswered` so callers can tell the three cases apart.
- **Held set:** `settle_rebuild`'s doc now states plainly that holding suppresses tab-order/active-tab sync, that nothing retries it short of the next rebuild, and that this widens the existing `Connecting`-pane hole rather than opening one. Agreed a retry / a way to close a held tab is separate work.
- **Minor:** the `write` docs and the commit message now say it's hardening for a *closed* link — not the held-open case, where the attach change is the fix — and that "once" is the writer's side only, the reader still raises its own `Exit`. I left `AppTabsNotRestored` and the `active` index as follow-ups since both predate this and the toast is shared with the launch path across three locales — happy to take either in a separate PR if you'd like.

Full suite green locally (1573 / 1154).

PR #682, opened 2026-08-18
fix(input): hand Ctrl+V to a full-screen program on the alternate screen (#677)

### LINKED ISSUE #677: windows 下的 vim/neovim ctrl-v还是粘贴
### Type

### What's going on?

我用的是scoop安装的最新版的tty7, vim和neovim也是scoop的最新版, 在window下在  powershell/cmd进入 vim/nvim后，按ctrl-v 会粘贴，这个键在vim系列里是块选择的作用。

作为对比
windows terminal删除 keybindings里的ctrl-v配置后, 以及wezterm/alacritty本来就没有没有这个问题
tty7我看到已经设置了ctrl+shift+v，但是ctrl-v还是会粘贴

### tty7 version (bugs)

26.8.3

### Platform (bugs)

Windows 11

### Anything else?
vim-9.2.258, neovim 0.12.4

biztex:
Fixes #677.

## The bug

In vim or neovim on Windows and Linux, Ctrl+V pasted the clipboard where the editor expected blockwise Visual mode. macOS was never affected, since Cmd+V is the paste chord there.

Ctrl+V was not a keybinding at all. `on_key_down` hands plain Ctrl+C, V and X to `handle_cmd_shortcut` off macOS, and of the three the "v" arm was the only unconditional one: Ctrl+C copies with a selection and otherwise falls through to SIGINT, Ctrl+X falls through outside the editor, but Ctrl+V always consumed, so SYN never reached the PTY — `input.rs` had the byte, unreachably — and an empty clipboard turned the key into nothing at all. #270 set the rule that off macOS ctrl-<letter> belongs to the terminal and anything sitting on one must fall through; Ctrl+V was the exception that had escaped it.

## The fix

The arm is now contextual like its neighbours. On the alternate screen it falls through, and `keystroke_to_bytes` sends 0x16 (or the CSI u form under the kitty protocol); off it Ctrl+V pastes exactly as before, and Cmd+V on macOS is untouched. The alternate screen is the gate rather than `input_active` because the editor is inactive whenever shell integration is missing or the prompt editor is off, and gating on that would take paste away from every such user; a program that has switched screens is precisely the case reported. Inside such a program paste is Ctrl+Shift+V, Shift+Insert or the right-click menu.

Two things ride along: the block did not exclude Shift, so Ctrl+Shift+C/V/X reached the hardcoded path whenever the keymap had nothing on them — the state rebinding Paste leaves behind, which #271 promised would retire Ctrl+Shift+V. Only unshifted chords enter the block now. And the right-click menu advertised Ctrl+C/X/V off macOS as though they were the bindings, next to a Select All row that already showed its hint on macOS only; the three rows take the same treatment.

One behavior change worth naming: off macOS, a TUI on the alternate screen (tmux, or an agent that runs full-screen) now receives `^V` on plain Ctrl+V rather than a paste — the intended alt-screen semantics, and what WezTerm and Alacritty do.

## Tests

`cargo test -p tty7` 1570 passed. New: `ctrl_v_on_the_alternate_screen_reaches_the_pty_as_syn` (end-to-end, `Some(vec![0x16])`), `ctrl_v_reaches_a_tui_on_the_alternate_screen` (`FallThrough`, and Cmd+V on the same screen still `Consumed`), `ctrl_v_pastes_off_the_alternate_screen`; the keymap paste test now asserts no default claims `ctrl-v` in the Terminal context. Sanity-checked: forcing the gate off fails the first and third. Not run on Windows locally; the path is `cfg!(not(target_os = "macos"))` and byte-identical on Linux, so Linux CI exercises it.

### DIFF
--- a/docs/reference/keyboard-shortcuts.mdx
+++ b/docs/reference/keyboard-shortcuts.mdx
@@ -53,6 +53,11 @@ it shows what *your* copy is bound to. This is the shipped default.
 | Accept ghost suggestion | <kbd>→</kbd> | same |
 | Completion menu | <kbd>⇥</kbd> | same |
 
+On Windows and Linux plain <kbd>Ctrl V</kbd> also pastes at a shell prompt.
+Inside a full-screen application — vim, less, tmux — it is passed through as
+the key, so it means what the application says it means (blockwise Visual mode
+in vim); paste there with <kbd>Ctrl ⇧ V</kbd> or <kbd>⇧ Insert</kbd>.
+
 ## Git and SSH
 
 | Action | macOS | Windows / Linux |
--- a/src/terminal/view.rs
+++ b/src/terminal/view.rs
@@ -1844,10 +1844,13 @@ impl TerminalView {
             }
         }
 
+        // Ctrl+Shift+C/V/X are the keymap's alone: rebinding Paste has to
+        // retire Ctrl+Shift+V, which it cannot if this path answers it too.
         if cfg!(not(target_os = "macos"))
             && m.control
             && !m.platform
             && !m.alt
+            && !m.shift
             && matches!(ks.key.as_str(), "c" | "v" | "x")
         {
             match self.handle_cmd_shortcut(ks, window, cx) {
@@ -1983,8 +1986,16 @@ impl TerminalView {
                 }
             }
             "v" => {
-                self.paste_from_clipboard(cx);
-                CmdKey::Consumed
+                // Off macOS Ctrl+V is a control code first: on the alternate
+                // screen the key is the program's (vim's blockwise select), the
+                // way Ctrl+C is SIGINT when there is nothing to copy. Cmd+V
+                // pastes anywhere.
+                if m.control && !m.platform && self.on_alt_screen() {
+                    CmdKey::FallThrough
+                } else {
+                    self.paste_from_clipboard(cx);
+                    CmdKey::Consumed
+                }
             }
             "a" => {
                 self.select_all_contextual(cx);
@@ -6139,16 +6150,16 @@ impl Render for TerminalView {
                     .menu_element_with_disabled(
                         Box::new(CopyText),
                         !has_selection,
-                        menu_row_with_hint(t(L10nKey::AppMenuCopy), Some("secondary-c")),
+                        menu_row_with_hint(t(L10nKey::AppMenuCopy), mac_only("secondary-c")),
                     )
                     .menu_element_with_disabled(
                         Box::new(CutText),
                         !has_selection,
-                        menu_row_with_hint(t(L10nKey::AppMenuCut), Some("secondary-x")),
+                        menu_row_with_hint(t(L10nKey::AppMenuCut), mac_only("secondary-x")),
                     )
                     .menu_element(
                         Box::new(PasteText),
-                        menu_row_with_hint(t(L10nKey::AppMenuPaste), Some("secondary-v")),
+                        menu_row_with_hint(t(L10nKey::AppMenuPaste), mac_only("secondary-v")),
                     )
                     .menu_element(
                         Box::new(SelectAll),
@@ -8355,6 +8366,27 @@ mod gpui_tests {
         panic!("the prompt report never reached the view");
     }
 
+    fn alt_screen_ready(
+        window: &gpui::WindowHandle<TerminalView>,
+        cx: &mut TestAppContext,
+        daemon: &mut UnixStream,
+    ) {
+        DaemonMsg::Output(b"\x1b[?1049h".to_vec())
+            .encode(daemon)
+            .unwrap();
+        for _ in 0..400 {
+            cx.run_until_parked();
+            if window
+                .update(cx, |view, _, _| view.on_alt_screen())
+                .unwrap()
+            {
+                return;
+            }
+            std::thread::sleep(std::time::Duration::from_millis(5));
+        }
+        panic!("the alternate-screen switch never reached the grid");
+    }
+
     #[gpui::test]
     fn an_agent_that_has_finished_its_turn_is_not_busy(cx: &mut TestAppContext) {
         use crate::core::cli_agent::{AgentSessionState, AgentStatus, CLIAgent};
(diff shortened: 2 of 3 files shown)


l0ng-ai:
The core fix is right, and correctly scoped. Confirming the diagnosis for anyone reading later: off macOS, plain Ctrl+C/V/X route into `handle_cmd_shortcut` (`src/terminal/view.rs:1848`); `c` and `x` are contextual and fall through to the PTY when there's nothing to copy, but `v` unconditionally pasted and returned `Consumed` — so the `0x16` byte in `src/terminal/input.rs:270` was dead code off macOS and vim never got blockwise-visual. macOS is genuinely untouched: the non-mac block is `cfg!`-gated, and `Cmd+V` enters through the `m.platform && !m.control` path where the new `m.control` test is false.

Two things I'd like addressed, one of them before merge.

### Please fix before merge: the replacement chord is now invisible exactly where it's needed

The three menu hints go from `Some("secondary-c"/"x"/"v")` to `mac_only(...)` at `src/terminal/view.rs:6153`, `6158`, `6163` — i.e. no hint at all off macOS. But `Cut` and `Select All` have no default binding off macOS while **Copy and Paste do**: `ctrl-shift-c` / `ctrl-shift-v` (`src/ui/keymap.rs:126`, `321`). So the blanket `mac_only` is right for two rows and wrong for two — on Windows/Linux the right-click menu now shows "Paste" with no hint at the precise moment the user's Ctrl+V stopped working in vim. `src/ui/palette.rs:216-221` has the same hole (`inline` returns `None` off macOS, and `CopyText | CutText | PasteText` early-return from the action lookup at `:297-314`), so there's no in-app surface left that teaches Ctrl+Shift+V.

Related: the PR body offers the right-click menu as the hatch, but neovim's default is `mouse=nvi` and `should_show_context_menu(mouse_mode, shift)` (`view.rs:6370`) suppresses the menu when a program owns the mouse — so in the program from #677, plain right-click opens nothing; it needs Shift+right-click.

One line each: pass `Some(if cfg!(target_os = "macos") { "secondary-v" } else { "secondary-shift-v" })` for Paste and the same shape for Copy, keeping `mac_only` for Cut and Select All. The docs page is accurate as written.

### Worth your answer: `!m.shift` is broader than #271 asked for

`src/terminal/view.rs:1852`. For `v` it's right. The collateral:

- **Ctrl+Shift+X**: `CutText` is bound nowhere — absent from `default_bindings()` and with no arm in `make_binding` — so the hardcoded path was its only handler. With the prompt editor active it now falls to `handle_editor_key`, whose unknown-Ctrl-chord fallback is `handoff_line_to_shell(&bytes)` (`view.rs:2172-2175`; `apply_readline_ctrl` has no `"x"` arm). So Ctrl+Shift+X on a half-typed command surrenders the inline editor and hands the line plus `^X` to the shell, where readline treats `^X` as a prefix key and waits for a second keystroke. Ctrl+X still cuts, so the loss itself is marginal — the handoff is the surprising part.
- **Ctrl+Shift+C** is only safe because `CopyText` holds `ctrl-shift-c` by default. A user who legally rebinds it (`"CopyText": "ctrl-alt-c"`) now gets `0x03`: in the editor `apply_readline_ctrl("c")` clears the line and writes SIGINT (`view.rs:2351-2357`); at a shell it kills the foreground job. Pressing what you think is Copy and killing a running build is worse than #271's stated goal, which names Paste only.

Narrowing to `!(m.shift && ks.key == "v")` gets #271's promise without either. If you'd rather keep the rule uniform, that's defensible — but then this half needs tests; there's currently no `ctrl-shift` keystroke anywhere in the `view.rs` test module.

### Accepting the trade-off, for the record

Ctrl+V stops pasting at a shell prompt inside any alt-screen wrapper. Concrete: user runs `tmux`, sits at a bash prompt *inside* it, presses Ctrl+V — tmux holds the alternate screen, so readline gets `0x16` = `quoted-insert` and the next keystroke is inserted literally (Ctrl+V then Enter puts a visible `^M` on the line). Same for `fzf` in full-screen mode, where `0x16` is just swallowed. That's the honest cost of the discriminator and it's roughly what WezTerm/Alacritty do by never binding Ctrl+V at all, so I'm not blocking on it — but it's why the hint change above matters more than it looks.

### Minor

- Test gaps worth closing since they're the behaviour a regression would hurt most: (a) Ctrl+Shift+V / Shift+Insert still paste **on the alternate screen**; (b) a user rebinding `PasteText` to `ctrl-v` still pastes there. Both hold by construction — gpui dispatches keymap actions before key listeners, and action listeners set `propagate_event = false` in the bubble phase, so `on_key_down` never runs — but they're unpinned.
- `ctrl_v_reaches_a_tui_on_the_alternate_screen` calls `handle_cmd_shortcut` with `key("cmd-v")`, a chord that never routes there off macOS. Fine as an arm-level unit test; worth a comment saying so.
- `m.control && !m.platform` at `:1993` — `!m.platform` is already guaranteed by both call sites, so it's defensive only.
- The comment's analogy is slightly off: Ctrl+C's discriminator is "is there a selection", Ctrl+V's is "is this the alt screen". Both defensible, but not "the same way" — Ctrl+C on the alternate screen still copies a tty7 selection rather than sending SIGINT.
- No CHANGELOG entry. `CHANGELOG.md:2462` ships "Ctrl+C, Ctrl+V and Ctrl+X are unchanged" (#270) and `:2471` "Ctrl+C/Ctrl+V still work as before" (#271); this narrows that on two platforms.
- Pre-existing, not yours: a pane stranded on the alt screen never gets Ctrl+V paste back, since `stale_mode_resets` (`src/terminal/remote.rs:1751`) clears `ALT_SCREEN` only on an OSC 133 prompt report. `ssh box` → `vim` → link drops → plain Ctrl+V silently sends `^V` until `reset`.

### Checked and clean

Bracketed paste is orthogonal — `paste()` reads `BRACKETED_PASTE` at `view.rs:2537` regardless of screen, so Ctrl+Shift+V into vim still arrives bracketed. Kitty protocol falls out correctly. AltGr on Windows is excluded by the pre-existing `!m.alt`. The alt-screen read is exact for remote panes.

l0ng-ai:
Thanks — the diagnosis is right and the fix reads correctly. Merging this.

A few notes on what comes after it; all of these are mine to do, nothing is asked of you here.

**On the alt-screen gate.** It is the right default, and it matches what WezTerm and Alacritty do. The one thing it does not leave behind is an escape hatch: after this, nobody can ask for Ctrl+V to paste inside a full-screen program, and nobody can ask for `^V` at the shell prompt either (readline's `quoted-insert` still loses to paste off the alternate screen). Warp keeps Ctrl+V pasting on Windows deliberately, but registers it as an *editable* binding — `terminal:alternate_terminal_paste`, "Alternate terminal paste" — with a comment saying it is modelled on Windows Terminal, "which also allows users to disable it, so we also make this an EditableBinding". That cohort is real, so the end state should be configurable rather than hardcoded in either direction. That is a follow-up, not a change to this PR.

**Follow-ups I'll take:**

1. `input.rs` is missing the rest of the C0 table. Ctrl+3..8 (ESC/FS/GS/RS/US/DEL) and Ctrl+/ (US) have no mapping at all, so `Ctrl-^` — vim's alternate file — and readline's undo are silent dead keys on every platform: gpui filters control characters out of `key_char` on all three backends, so the text fallback produces nothing either. Warp maps exactly these in `keystroke_to_c0_control_code`, with a dedicated regression test for `ctrl-/` because Vim/Neovim bind `<C-/>`.

2. `view.rs` also swallows plain Ctrl+1..9 off macOS with a bare `return`, immediately below the block you touched. Same shape as the bug you fixed, left over from when tabs lived on ctrl-digits — they are on alt-1..9 now, so nothing claims those chords off macOS and the block only deletes keys. It also sits *before* `keystroke_to_bytes`, so not even the kitty protocol gets through it.

3. Ctrl+V becomes a real, removable binding carrying an `alt_screen` key context, so the default stays exactly what you implemented here while the keymap can finally express it. This one has to wait for this PR to land, since it replaces the arm you changed.

4. `no_default_binding_sits_on_a_terminal_control_code` only walks the defaults, so a hand-edited `config.json` can still bind `ctrl-d` and swallow EOF without a word. Warp enforces the same rule as a runtime binding validator with an explicit, commented allowlist; ours should too.

Your tests survive all of that — they assert the bytes on the wire, not the code path.

PR #683, opened 2026-08-18
fix(linux): stop the compositor framing a window that draws its own title bar (#679)

### LINKED ISSUE #679: Linux/Wayland：自绘标题栏与原生边框同时显示（双边框）
### 类型

Bug

### 问题描述

在 Linux/Wayland 环境中，tty7 会同时显示自己的标题栏和 compositor 提供的原生边框，导致窗口出现双边框。

复现版本：tty7 v26.8.3。

tty7 在 `src/ui/windows.rs` 的 `window_options()` 中设置了自绘 `gpui_component::TitleBar` 对应的 `titlebar`，但没有显式设置 `window_decorations`。

GPUI 随后使用默认值：

```rust
platform_window
    .request_decorations(window_decorations.unwrap_or(WindowDecorations::Server));
```

因此 Wayland 收到：

```text
zxdg_toplevel_decoration_v1.set_mode(2)
```

`xdg-decoration-unstable-v1` 中，`client-side` 是 `1`，`server-side` 是 `2`。所以 compositor 又绘制了原生 server-side decoration，与 tty7 自己绘制的标题栏叠加，形成双边框。

### 预期行为

当 tty7 使用自绘标题栏时，窗口只应显示一套标题栏和边框。

### 可能的修复方式

在 tty7 的 `WindowOptions` 中显式请求 client-side decoration：

```rust
window_decorations: Some(gpui::WindowDecorations::Client),
```

Zed 已经在创建窗口时显式传入 `window_decorations`，默认配置也是 client-side，因此没有这个问题。

### 环境

- tty7：v26.8.3
- 平台：Linux
- 会话：Wayland

biztex:
Fixes #679 — the diagnosis in the report is exactly right.

## The bug

tty7 paints its own title bar through gpui-component's `TitleBar`, and the `WindowOptions` it opens with say as much (`appears_transparent`) — but say nothing about decorations. gpui reads a missing `window_decorations` as `WindowDecorations::Server` (`window.rs`: `request_decorations(window_decorations.unwrap_or(WindowDecorations::Server))`) and, on Wayland, sends `zxdg_toplevel_decoration_v1.set_mode(server_side)`, so a compositor that honours it draws a second title bar and border around the one the app already has.

## The fix

`window_options()` now asks for `WindowDecorations::Client`. Nothing new has to be painted for it — gpui-component's `Root` already wraps the window in `window_border()`, `bordered` defaults to `true`, and tty7 never calls `.bordered(false)` at any of its `Root::new` sites. Under `Decorations::Client` that border draws the 1px frame, the 12px shadow (Linux only; `SHADOW_SIZE` is `px(0.)` elsewhere), the resize hit bands and their cursors, starts the resize from its own `on_mouse_down`, and reports the inset through `set_client_inset`. Under `Decorations::Server` it degrades to a plain `div`. The request was the only piece missing. Right-click-for-window-menu lives in `TitleBar`, already gated on `is_linux && is_client_decorated`, and turns on with the same request.

The field is set without a `cfg`. `request_decorations` is an empty default on the `PlatformWindow` trait, and the only overrides in the pinned gpui are the Wayland, X11 and web backends — macOS and Windows never see it, and `window_decorations()` keeps answering `Decorations::Server` there. X11 turns the request into `_MOTIF_WM_HINTS` and falls back to server-side on its own when no compositor is present, so a bare X session still gets a window-manager frame; a reparenting WM under a compositor, which today is told in the same hints to decorate, gets the same fix as Wayland.

### Remembered bounds

Client-side decorations bring one follow-on Zed hit too: `ca9cee85e1` ("linux: Fix non-maximized Zed windows growing larger across sessions", #22301). The bounds tty7 remembers go back in through `WindowOptions::window_bounds`, which every backend reads as the **outer** rectangle — the whole surface, shadow included, under CSD. Both Linux backends put the shadow back on once the window is up, so reopening at the outer rectangle grows the window by twice the inset per launch:

- **Wayland** re-inflates in gpui itself: the first sized configure runs `compute_outer_size(state.inset(), …)`, which adds the inset back onto whatever geometry was asked for.
- **X11** re-inflates through the window manager. gpui only enables CSD there when a compositor is present *and* the WM advertises `_GTK_FRAME_EXTENTS` (`client_side_decorations_supported`), and a WM that advertises that atom honours it by keeping the *visible* frame put and treating the extents as shadow outside it. Zed's numbers in `ca9cee85e1` were all taken on X11 — +20px per session before, stable after.

So a `window_bounds_to_remember()` helper saves `inner_window_bounds()`, unconditionally, exactly as Zed still does at the pinned rev. It is a no-op wherever there is no inset to strip: `inner_window_bounds` defaults to `window_bounds` on the trait and neither the macOS nor the Windows backend nor gpui's `TestWindow` overrides it, and on X11 without a compositor `window_decorations()` answers `Server`, so the window border never calls `set_client_inset` and `last_insets` stays `[0, 0, 0, 0]`. Maximized and fullscreen are identical in both rectangles on both backends. `WindowState` round-trips unchanged.

*(The first revision of this PR split Wayland from X11 here, saving outer on X11 on the reading that it creates the window at the requested rectangle verbatim. Second commit undoes that: the citation's own measurements are the counter-example.)*

### Drop band

The one place that hardcoded the window's corner has to follow the frame: the pane-to-tab-strip drop band was a rectangle from `(0, 0)` spanning the full viewport width, which under CSD is the 12px shadow strip plus the top 28px of the 40px title bar — the bar's bottom 12px fell outside it — and which runs one shadow past the frame at each end. It now comes from a `strip_band(viewport, pad)` helper fed by `gpui_component::window_paddings(window)` — start at the padding, lose one padding at each end. `window_paddings` answers `Edges::all(0)` under `Decorations::Server`, and gpui's default `window_decorations()` is `Server` with no macOS or Windows override, so this is a no-op off Linux CSD. It clamps at zero: a surface narrower than its own shadow is only reachable mid-resize, but a negative width would hand `Bounds::contains` a rectangle that is inside out.

## Tests

`cargo test -p tty7 --bin tty7-app`: 1603 passed. Rebased onto `2cdc26f3`.

- `a_window_asks_the_compositor_for_client_side_decorations` — calls the real `window_options()`, asserts `Some(Client)` and the transparent title bar it stands in for.
- `an_undecorated_frame_leaves_the_strips_drop_band_where_it_was` — no padding, band unmoved from the corner.
- `a_client_side_frame_pulls_the_drop_band_in_by_the_shadow_on_both_sides` — the band's far edge lands on the frame, not on the surface.
- `a_drop_band_narrower_than_its_frame_collapses_instead_of_inverting`.

`window_bounds_to_remember` is deliberately untested: gpui's `TestWindow` overrides neither `inner_window_bounds` nor the decorations, so inner and outer are one rectangle in every test and any assertion would only restate the call.

## Not verified

There is no Linux session here, so everything below was read out of the pinned gpui (`d99a40a5`) and gpui-component (`070d1a2`) rather than run:

- Whether the reporter's compositor honours the mode switch at all — `xdg-decoration-unstable-v1` lets it refuse.
- How the 12px shadow and the 1px `window_border` colour read against the shipped themes.
- The edges of a maximized or tiled window, and where gpui-component drops the padding on the tiled sides.
- One quit-and-relaunch on X11 under a compositor, to watch the remembered geometry actually hold.
- FreeBSD runs the same backends with gpui-component's `SHADOW_SIZE` at zero; unexercised.

A look from someone on GNOME or KDE would be welcome.

### DIFF
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -293,6 +293,27 @@ pub(crate) fn title_bar_hug_offset() -> f32 {
     }
 }
 
+/// The bounds to remember for reopening this window at the same place.
+///
+/// The value goes back in through `WindowOptions::window_bounds`, which every
+/// backend reads as the *outer* rectangle, so outer is what to save — except
+/// on Wayland under client-side decorations, where the outer rectangle is the
+/// surface, shadow included, and the compositor's first sized configure adds
+/// the inset back onto whatever was asked for (`compute_outer_size`). Saving
+/// outer there grows the window by twice the shadow on every launch (the bug
+/// Zed fixed in ca9cee85e1); saving inner pre-deflates by exactly what the
+/// configure re-inflates, and the geometry holds still. X11 never re-inflates:
+/// it creates the window at the requested rectangle verbatim and its
+/// `inner_window_bounds` also shifts the origin by the inset, so saving inner
+/// there would shrink the window and walk it down-right by the shadow on
+/// every launch. macOS and Windows report no inset, so the two are the same.
+fn window_bounds_to_remember(window: &Window, cx: &App) -> Bounds<Pixels> {
+    match cx.compositor_name() {
+        "Wayland" => window.inner_window_bounds().get_bounds(),
+        _ => window.window_bounds().get_bounds(),
+    }
+}
+
 pub(crate) const WINDOW_MARK_SIZE: f32 = 20.;
 
 pub(crate) fn title_bar_drag(
@@ -1298,7 +1319,7 @@ impl Tty7App {
             settings: None,
             ssh_prompt: crate::ui::ssh_prompt::SshPromptState::new(cx),
             close_prompt_open: false,
-            window_bounds: window.window_bounds().get_bounds(),
+            window_bounds: window_bounds_to_remember(window, cx),
             workspace,
             workspace_rename: None,
             window_title: std::cell::RefCell::new(String::new()),
@@ -1322,8 +1343,8 @@ impl Tty7App {
         })
         .detach();
 
-        cx.observe_window_bounds(window, |this, window, _cx| {
-            this.window_bounds = window.window_bounds().get_bounds();
+        cx.observe_window_bounds(window, |this, window, cx| {
+            this.window_bounds = window_bounds_to_remember(window, cx);
         })
         .detach();
 
@@ -3817,6 +3838,11 @@ impl Tty7App {
         let vertical = matches!(cx.global::<Config>().tab_bar_position, TabBarPosition::Left)
             && !self.tabs.is_empty();
         let viewport = window.viewport_size();
+        // The pointer and the viewport are in the window's outer coordinates;
+        // under client-side decorations the content — title bar included —
+        // sits inside the frame padding, so the strip's band starts there and
+        // not at the window's corner. Zero padding everywhere else.
+        let pad = gpui_component::window_paddings(window);
         // The band each surface claims, read off the tabs it drew rather than
         // measured as an element of its own: the chips and the rows are the
         // only part of either surface a drop has anything to say about, and a
@@ -3833,8 +3859,8 @@ impl Tty7App {
                 .reduce(Pixels::max)?;
             Some(match axis {
                 Axis::Horizontal => Bounds {
-                    origin: point(px(0.), px(0.)),
-                    size: size(viewport.width, px(TITLE_BAR_HEIGHT)),
+                    origin: point(pad.left, pad.top),
+                    size: size(viewport.width - pad.left - pad.right, px(TITLE_BAR_HEIGHT)),
                 },
                 Axis::Vertical => Bounds {
                     origin: point(left, top - px(BAND_REACH)),
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -1,6 +1,6 @@
 use gpui::{
-    App, Axis, Bounds, Context, Entity, Focusable, Pixels, PromptLevel, Subscription, Window, div,
-    img, point, prelude::*, px, size,
+    App, Axis, Bounds, Context, Edges, Entity, Focusable, Pixels, PromptLevel, Size, Subscription,
+    Window, div, img, point, prelude::*, px, size,
 };
 use gpui_component::color_picker::{ColorPickerEvent, ColorPickerState};
 use gpui_component::input::{InputEvent, InputState};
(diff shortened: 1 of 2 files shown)


l0ng-ai:
The one-line change in `src/ui/windows.rs:717` is correct, minimal and well justified — I'd take it as is. I checked every factual claim in the body against the pinned gpui (`zed@1d217ee`) and gpui-component (`070d1a2`) and they all hold: `request_decorations` is an empty default that neither the macOS nor the Windows backend overrides (`gpui/src/platform.rs:700` — only x11/wayland/web override it), the default `window_decorations()` getter returns `Decorations::Server` (`platform.rs:704`), `SHADOW_SIZE` is `px(0.)` off Linux, and `Root::bordered` defaults true with tty7 never clearing it (`src/ui/windows.rs:249`). macOS and Windows are provably untouched.

The blockers are the two things riding along with it.

### Blocking: save and restore now use different coordinate spaces, and X11 has nothing to put the inset back

`src/ui/app.rs:1249`, `1278` move the *save* side to `inner_window_bounds()`, but the *open* side still feeds `WindowOptions.window_bounds` (`src/ui/windows.rs:704`), which every Linux backend consumes as the **outer** rectangle.

Wayland gets away with it: `compute_outer_size` (`wayland/window.rs:661`, `1690`) adds the inset back when the compositor sends a sized configure, so outer settles at inner+24 and the visible geometry is stable. That's exactly why the Zed fix works.

X11 has no such compensation. `inner_window_bounds` (`x11/window.rs:1339-1362`) subtracts `_GTK_FRAME_EXTENTS` **and shifts the origin by +left/+top**; `X11Window::new` creates the window at the requested rectangle verbatim (`x11/window.rs:473-509`), and `ConfigureNotify` writes the outer geometry straight back (`set_bounds`, `x11/window.rs:1236-1250`). So wherever `client_side_decorations_supported` is true — compositor present **and** `_GTK_FRAME_EXTENTS` in `_NET_SUPPORTED`, i.e. GNOME on Xorg and KDE Plasma X11, both of which qualify — the visible window should shrink 24px and creep 12px down-right on every launch, until `at_least_min_size` clamps it at 720×520 and the off-screen check re-centres it. `remember_window_size` defaults to true.

Before this PR X11 was stable (server decorations ⇒ zero insets ⇒ inner == outer), so this would be a regression the PR introduces rather than an inherited one.

**This is read from source, not observed** — I have no Linux box here, so please verify one quit/relaunch cycle on an X11 session with a compositor before we ship it. If it does shrink, `inner_window_bounds()` can't be used unconditionally; either read it only where the platform re-inflates it, or re-apply the size once the decoration mode resolves. Either way X11 belongs in the commit's "not verified here" list — it currently mentions Wayland only.

### Blocking: the pane→tab-bar drop band is hardcoded to the window origin

`src/ui/app.rs:3786-3788`:

```rust
Axis::Horizontal => Bounds {
    origin: point(px(0.), px(0.)),
    size: size(viewport.width, px(TITLE_BAR_HEIGHT)),   // 40.
},
```

Under client decorations the content box starts at (12,12) — `window_border` pads the whole tree, title bar included — while `window.mouse_position()` and `viewport_size()` stay in outer coordinates. So the band covers the transparent shadow strip above the frame and misses the bottom 12px of a 40px title bar: on KWin or sway, dragging a pane onto the lower third of the tab strip silently does nothing. The `Vertical` arm is fine, since it's derived from the drawn slots. One line — use the `top` already computed two lines above, or `gpui_component::window_paddings(window)`.

### Please correct the description

"what Zed sets unconditionally" isn't right: Zed reads a user setting, `WorkspaceSettings::window_decorations` (`settings_content/src/workspace.rs:133`, enum at `:349`), defaulting to Client and threaded into `WindowOptions` at `zed/src/zed.rs:386`.

Which raises the question of whether we want the same key. The case that wants it is sway and i3-on-Wayland: sway drops its own border for a CSD client (`border csd`), and gpui-component drops *its* border and shadow once the window is tiled on all four sides (`window_border.rs:110-116`, `168-176`), so a tiled tty7 ends up with no border at all and loses the focus indication every neighbouring window has. A `window_decorations: "client" | "server"` config key mirroring Zed's would cover that. Not a merge blocker — happy to do it as a follow-up.

### Minor

- Viewport-based width math is now 24px optimistic on Linux: `src/ui/tab_strip.rs:1523` (`strip_w = viewport.width - 114` reserves room for the window controls against the *outer* width, so `visible_chips` may keep a chip that gets clipped against the controls — the same overrun class the comment above it describes for macOS), plus `tab_sidebar.rs:118`, `right_panel.rs:307`, `tab_strip.rs:582`. Cosmetic, but a `window_paddings`-aware helper would be worth it if you're touching #2 anyway.
- Worth saying in the PR that the `app.rs` change is untestable in principle: gpui's `TestWindow` never overrides `inner_window_bounds`, so inner and outer are identical in every test. The `windows.rs` test pins the field it sets, which is as good as that half gets.
- With `window_blur` on, the Wayland blur region is the whole surface rather than the opaque area (`wayland/window.rs:1636-1650`), so KWin would blur 12px beyond the visible frame. Cosmetic, unverifiable here.
- `inner_window_bounds` returns the *outer* restore rect while maximized on both backends, so quitting maximized persists a value 24px larger than quitting windowed. Harmless, but it means the persisted number's meaning depends on window state.

### Checked and clean

The classic CSD failure — compositor refuses, window ends up with no title bar and no way to move or close it — is genuinely covered, and I couldn't name a compositor that triggers it. If the compositor answers server-side, gpui follows the configure event and flips `decorations` back (`wayland/window.rs:703-728`); X11 downgrades to Server on its own when no compositor is present (`x11/window.rs:1823-1833`); `window_border` degrades to a plain div; and gpui-component's `TitleBar` paints the drag area, double-click zoom, min/max/close and the right-click window menu on every non-macOS platform regardless of mode (`title_bar.rs:203-330`). KWin, sway, Hyprland and river all honour a client-side request and mutter is CSD-only, so treat the refusal path as theoretical.

The reverse doesn't adapt — under forced Server decorations tty7 keeps painting its own title bar and buttons, so such a compositor sees two. Zed hides its controls in that case (`platform_title_bar.rs:126-166`). Pre-existing, not a regression here, but it's the other half of the config key above.

biztex:
Thanks — both blockers confirmed and fixed in `8b968de` (rebased onto current main).

**Save/restore spaces.** You're right, and X11 is worse than "no compensation": its `inner_window_bounds` also shifts the origin by the inset, so the window would shrink *and* creep down-right per launch. There is a runtime backend signal, though — `cx.compositor_name()` is public on `App` and answers `"Wayland"` / `"X11"` / `""` elsewhere — so the save side is now a `window_bounds_to_remember` helper that saves inner on Wayland (pre-deflating exactly what the first sized configure re-inflates) and outer everywhere else, with the reasoning in its doc comment. I checked Zed's restore path while I was at it: it does *not* re-inflate on open either, so Zed carries the same X11 creep upstream under its default `client` setting — I couldn't find it filed, so "Zed does it" was never evidence the shape was right. X11 is now in the "not verified" list, with the one quit-and-relaunch you asked for.

**Drop band.** Fixed with `gpui_component::window_paddings(window)` — zero under server decorations, so nothing moves off CSD. Left the `Vertical` arm alone; it over-reaches into the bottom shadow rather than under-reaching, so it's cosmetic.

**Description.** Corrected to "what Zed defaults to (`window_decorations`, a user setting, overridable via `ZED_WINDOW_DECORATIONS`)". Agreed on the config key for sway/i3 and for the forced-server case where tty7 keeps painting its own buttons — happy to take that as a follow-up.

Also noted in the message: the `app.rs` half is untestable in principle (TestWindow overrides neither `inner_window_bounds` nor the decorations), and FreeBSD runs the same backends with the shadow at zero, unexercised. The viewport-width helper for the four sites you listed is a good follow-up too.

Full suite green (1582).

PR #688, opened 2026-08-19
fix(terminal): shape regional indicators as flags

### LINKED ISSUE #686: Linux：终端窗格无法正确渲染国旗 Emoji
## 问题类型

这是一个 Linux 终端窗格中的 Emoji 渲染问题。

## 问题描述

在 Linux 下，tty7 无法正常显示由 Regional Indicator 字符组成的国旗 Emoji。例如完整字符串 `🇨🇳` 在终端窗格中不能稳定渲染为一个中国国旗，可能显示为两个分离的区域指示符、字母样式的符号或其他不完整结果。

这不是字体未安装的问题。我的系统可以通过 Fontconfig 找到 `Noto Color Emoji`，tty7 的 Linux 默认字体回退链也包含该字体；问题出现在 tty7 将国旗序列交给字体引擎之前的终端 cell 分组和绘制逻辑。

## 复现环境

- 操作系统：Linux（NixOS）
- tty7 版本：26.8.3
- 主字体：`Maple Mono NF CN`
- 回退字体：`Noto Sans CJK SC`、`Noto Color Emoji`
- 当前上游源码提交：`f44b667639475a1ccb41b7381d737040f3c879a5`

## 复现步骤

在 tty7 的终端窗格中执行：

```sh
printf '🇨🇳 🇺🇸 🇯🇵\n'
printf '前🇨🇳后\n'
```

## 预期行为

每个国旗都应作为一个完整的彩色 Emoji 显示，并占用两个终端列；国旗前后的文字和多个连续国旗之间应保持正确对齐。

## 实际行为

国旗不能作为完整的彩色旗帜显示，或者显示为两个彼此分离的符号。周围文本的列位置也可能因此出现偏移。

## 源码分析

`🇨🇳` 不是一个 Unicode 码点，而是两个 Regional Indicator 字符：

```text
U+1F1E8 U+1F1F3
```

当前 `src/terminal/element.rs` 中的 `segment_row` 只会把带零宽附加字符的 cell，以及特定的泰文/老挝文组合，合并为一个绘制 cluster。普通的非 ASCII 字符如果后面没有宽字符 spacer，就会进入 `Solo` 分支。

随后 `paint_glyphs` 会对每个 `Solo` cell 单独调用 `shape_line`，也就是分别向字体引擎提交两个独立字符，而不是提交完整的 `🇨🇳` 序列。字体引擎因此没有机会依据 Regional Indicator 配对规则生成国旗字形。

相关代码：

- `src/terminal/element.rs` 中的 `segment_row`
- `src/terminal/element.rs` 中的 `RowSeg::Solo` 绘制路径
- `src/terminal/element.rs` 中对单个 cell 调用 `shape_line` 的逻辑

源码链接：

https://github.com/l0ng-ai/tty7/blob/f44b667639475a1ccb41b7381d737040f3c879a5/src/terminal/element.rs#L664-L742

https://github.com/l0ng-ai/tty7/blob/f44b667639475a1ccb41b7381d737040f3c879a5/src/terminal/element.rs#L945-L1065

## 与已有问题的区别

- #203 主要讨论 Windows 下普通 Emoji 的宽度和渲染问题。
- #450 主要修复侧栏或 Tab 标题在截断文本时按单个字符切割 grapheme cluster，导致国旗被截成半个的问题。
- 本问题发生在 Linux 终端窗格的正常文本绘制过程中，即使完整的 `🇨🇳` 没有被截断，也会因为两个 Regional Indicator 被分别绘制而无法合成国旗。

## 修改建议

建议在终端 cell 分组和绘制路径中增加 Regional Indicator 配对处理：

1. 识别相邻的两个 Regional Indicator 字符，并将它们作为一个 grapheme cluster 交给字体 shaping，而不是分别调用 `shape_line`。
2. 让该 cluster 在终端网格中保持两列宽度，同时正确处理光标移动、换行、重排、选择、复制和行尾边界情况。
3. 不要只修改字体回退链；`Noto Color Emoji` 已经能够覆盖这些字符，关键是需要把两个码点作为一个序列传给字体引擎。
4. 增加回归测试，至少覆盖：
   - `🇨🇳x`：国旗后面的字符不能左移；
   - `前🇨🇳后`：前后文本保持正确列位置；
   - `🇨🇳🇺🇸🇯🇵`：连续国旗分别显示为完整旗帜；
   - 混合普通 Emoji、CJK、国旗以及行尾/换行场景。

感谢维护者审阅。

green3sf:
## Summary

- shape adjacent, same-style Regional Indicator cells together so the font engine receives a complete flag sequence
- keep each flag cluster constrained to its two terminal columns without changing existing combining-mark and Sara Am clusters
- leave unmatched indicators and indicators separated by a style boundary on the existing solo-glyph path

Fixes #686.

## Tests

- `cargo fmt --check`
- `git diff --check`
- added regression coverage for a flag followed by text, CJK text around a flag, consecutive flags, an unmatched indicator, and a style boundary

The focused `cargo test --locked --offline --bin tty7-app segment_row_shapes_regional_indicator_pairs_as_two_column_flags` command stopped during dependency resolution because the pinned `russh` Git dependency is not present in the local Cargo cache. No third-party dependency or build script was downloaded or executed locally; the repository's GitHub Actions matrix can perform the full locked build and test run.

### DIFF
--- a/src/terminal/element.rs
+++ b/src/terminal/element.rs
@@ -643,7 +643,7 @@ enum RowSeg {
         col: usize,
         cells: usize,
         text: String,
-        wide_base: bool,
+        double_width: bool,
     },
 }
 
@@ -654,6 +654,23 @@ fn push_cell(text: &mut String, cell: &RenderCell) {
 
 fn is_sara_am(c: char) -> bool {
     matches!(c, '\u{0E33}' | '\u{0EB3}')
+}
+
+fn is_regional_indicator(c: char) -> bool {
+    ('\u{1F1E6}'..='\u{1F1FF}').contains(&c)
+}
+
+fn regional_indicator_pair_at(row: &[RenderCell], col: usize) -> Option<&RenderCell> {
+    let first = row.get(col)?;
+    let second = row.get(col + 1)?;
+    (!first.spacer
+        && first.marks.is_none()
+        && is_regional_indicator(first.c)
+        && !second.spacer
+        && second.marks.is_none()
+        && is_regional_indicator(second.c)
+        && GlyphStyle::of(first) == GlyphStyle::of(second))
+    .then_some(second)
 }
 
 fn sara_am_at(row: &[RenderCell], col: usize) -> Option<&RenderCell> {
@@ -709,13 +726,24 @@ fn segment_row(row: &[RenderCell]) -> Vec<RowSeg> {
                 col,
                 cells,
                 text,
-                wide_base,
+                double_width: wide_base,
             });
             col += cells;
             continue;
         }
         if !cell.c.is_ascii_graphic() {
-            if col + 1 < row.len() && row[col + 1].spacer {
+            if let Some(second) = regional_indicator_pair_at(row, col) {
+                let mut text = String::with_capacity(2);
+                text.push(cell.c);
+                text.push(second.c);
+                segs.push(RowSeg::Cluster {
+                    col,
+                    cells: 2,
+                    text,
+                    double_width: true,
+                });
+                col += 2;
+            } else if col + 1 < row.len() && row[col + 1].spacer {
                 segs.push(RowSeg::Wide {
                     start: col,
                     cells: 2,
@@ -732,7 +760,7 @@ fn segment_row(row: &[RenderCell]) -> Vec<RowSeg> {
                     col,
                     cells: 2,
                     text,
-                    wide_base: false,
+                    double_width: false,
                 });
                 col += 2;
             } else {
@@ -1033,12 +1061,12 @@ fn paint_glyphs(
                     col,
                     cells,
                     text,
-                    wide_base,
+                    double_width,
                 } => (
                     col,
                     cells,
                     SharedString::from(text),
-                    (cells == 2).then(|| geom.cell_width * if wide_base { 2. } else { 1. }),
+                    (cells == 2).then(|| geom.cell_width * if double_width { 2. } else { 1. }),
                     cells == 1,
                 ),
             };
@@ -2638,7 +2666,7 @@ mod tests {
             col,
             cells,
             text: text.to_string(),
-            wide_base: false,
+            double_width: false,
         }
     }
 
@@ -2647,7 +2675,7 @@ mod tests {
             col,
             cells,
             text: text.to_string(),
-            wide_base: true,
+            double_width: true,
         }
     }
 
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Closing in favour of #691, which fixes the same issue.

The diagnosis here is right, and the shape of the fix is the same one: join an adjacent Regional Indicator pair into a single `Cluster` so `shape_line` sees both codepoints and the font gets a chance to form the flag ligature. Three details decided it.

**1. `double_width: true` pins the second glyph outside the clip.**

That field feeds `force_width` in `shape_line`, and gpui's `apply_force_width_to_layout` uses it as the stride between successive *base* glyphs — `forced_x = glyph_pos * force_width`. The two indicators of a flag are one column apart, not two. With `2 * cell_width`, the second base glyph is pinned to `x = 2 * cell_width`, which is exactly the edge of the cluster's clip (`seg_clip_width` = `cells * cell_width` = two columns), so it is clipped away.

It makes no visible difference when the ligature forms — one glyph, pinned at 0 either way. It does when the font has no flag for that pair and each indicator falls back to its lettered box: one half vanishes. #691 leaves `wide_base: false`, so the second box stays inside its own column.

**2. The `GlyphStyle` equality check drops the flag whenever the halves differ.**

`GlyphStyle` carries `fg`, and the two cells do differ in real situations. `invert_cursor_cell` rewrites the fg of whichever cell the cursor sits on, so parking the cursor on the second indicator splits the flag into two lettered boxes; so does any per-character colouring — lolcat-style output, gradient TUI headers — that lands a different SGR on each half.

#691 takes the first cell's style, the same rule the existing SARA AM cluster already follows. Losing one half's fg costs little here: colour emoji glyphs ignore it, and selection washes and cursor fills are painted in a separate pass, so both still show across the whole pair.

**3. The check sits after the marks branch.**

Because `regional_indicator_pair_at` requires `marks.is_none()` on both cells, a zero-width mark on either half (VS16, for instance) sends the first indicator into the marks branch as a one-cell cluster and leaves the second as a `Solo` — the flag breaks apart again. #691 puts the check ahead of the marks branch and builds the text with `push_cell`, so a mark on either half rides along inside the pair.

Two smaller points: the `wide_base` → `double_width` rename reworks the meaning of an existing field ("the base is a double-width character") to accommodate the new use, and the tests in this PR were never executed locally.

Thanks for digging into this — the root-cause analysis was correct, and having two independent patches to compare made the pinning question much easier to settle.

PR #735, opened 2026-08-26
feat(sidebar): rename group headers from the context menu (#718)

### LINKED ISSUE #718: [Idea] Rename sidebar group headers
### Type

### What's going on?

Sidebar groups are titled from the last path component, uppercased. Two clones, worktrees, or folders that share a name all read as `TTY7`. Tabs and workspaces can already be renamed; the group header cannot.

I'd like to rename a group from its header's context menu, with the name surviving restart. Clearing it should restore the derived title. Groups that were never renamed stay as they are.

### Anything else?

Nearby but not the same: #281 is the tab title under a group, #620 is how groups are keyed. This is only the header string. I have a small patch if this shape is one you'd take.

cloudy-liu:
Fixes #718.

## The problem

Sidebar group titles are derived from their root paths. That makes similarly named worktrees hard to distinguish and gives the Scratch group no way to carry project context.

## The change

- Add a Rename Group action to each sidebar group's context menu.
- Edit the title inline in the sidebar.
- Persist overrides in `sidebar_group_names`, keyed by the group's root path. Scratch uses the empty key.
- Remove the override when the submitted name is empty so the derived title returns.
- Add English, Japanese, and Chinese labels for the action.

The config field has a serde default, so existing config files continue to load unchanged.

## Tests

- `cargo test --locked -p tty7-core sidebar_group_names_round_trip_and_default_empty`
- `cargo test --locked -p tty7 --bin tty7-app group_rename`

### DIFF
--- a/crates/tty7-core/src/core/config.rs
+++ b/crates/tty7-core/src/core/config.rs
@@ -206,6 +206,12 @@ pub struct Config {
     pub sidebar_grouping: SidebarGrouping,
     #[serde(default = "default_true")]
     pub sidebar_diff_preview: bool,
+    /// Custom sidebar group titles, keyed by the group's root path — `""`
+    /// stands for the Scratch group. A group without an entry keeps the name
+    /// derived from its path, and an entry is removed (not blanked) when the
+    /// user commits an empty rename, so the derived name comes back.
+    #[serde(default)]
+    pub sidebar_group_names: HashMap<String, String>,
     #[serde(default, deserialize_with = "de_lenient")]
     pub notify_on_command_finish: NotifyMode,
     pub check_for_updates: bool,
@@ -605,6 +611,7 @@ impl Default for Config {
             scm_graph_expanded: false,
             sidebar_grouping: SidebarGrouping::Repo,
             sidebar_diff_preview: true,
+            sidebar_group_names: HashMap::new(),
             notify_on_command_finish: NotifyMode::Unfocused,
             check_for_updates: true,
             update_channel: UpdateChannel::default(),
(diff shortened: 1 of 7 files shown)


l0ng-ai:
Thanks for this — the code is clean and the want behind it is real. I'd like to
hold it, though, because I think it lands on the wrong layer. Wrote the long
version up as #756; the short version:

**A sidebar group has nothing to hang a name on.** It isn't a level of the
hierarchy — `sidebar_group_keys()` recomputes it every frame from the first
leaf's cwd, and `Tab.sidebar_group` is a sticky cache of that guess, not a
declaration. So the name has to be keyed on the path, which is what
`sidebar_group_names` does, and every rough edge follows from that: rename a
directory or move a worktree and the name is orphaned; the group goes away and
its config entry stays forever with no GC; you can't name a project before
opening a tab in it.

The fix I'd rather make is to give the project a real identity inside the
workspace (`Project { id, name, root }`, `Tab.project: Option<ProjectId>`), with
cwd probing demoted to auto-filing. Then the name lives on the entity, moving the
project is an edit to `root`, and "Open Project" becomes an action that exists.

One correction on the motivation while I'm here: two worktrees sharing a name do
**not** both render as `TTY7`. `group_names()` (`src/ui/tab_sidebar.rs:1406`)
already disambiguates collisions by walking up a component at a time, so they
show as `025/TTY7` and `026/TTY7`. The part of #718 that does hold is the other
one — wanting a name with nothing to do with the path — and that's the part #756
is about.

I'm holding rather than closing because the schema work in #756 is a small epic
and might not land soon. If you'd rather ship this as an interim, say so and I'll
take it with two changes: `BTreeMap` instead of `HashMap` (a `HashMap` reorders
the config file on every write once there's more than one entry), and a note on
the field that it's slated for removal. The reason I lean against that is the
stored data — once the field ships, users have names in it, and the eventual
`Project` migration has to reverse-map paths back to entities.

Two things in the current diff worth keeping either way, since they're right
regardless of which layer wins:

- The group header had no context menu before this, so adding one shadows
  nothing.
- The rename box guards mouse-down and click, but `.drag(DragGroup)` is still on
  the outer `base`, so a header should still be draggable mid-rename, committing
  on the blur. Untested, minor.

PR #739, opened 2026-08-26
fix(windows): detect in-pane ssh hops

cloudy-liu:
## Summary
- Detect interactive `ssh` / `ssh.exe` hops inside a Windows pane by walking the process tree and reading argv (Unix already had foreground-pgid detection).
- Clear local git sidebar grouping (branch / +N -M) once the pane is on an SSH hop, so the far side is not treated as this machine's repo.
- Wake the UI when `RemoteContext` arrives so the sidebar refreshes promptly.

## Test plan
- [ ] On Windows, open a local pane in a git repo and confirm sidebar shows branch / +N -M
- [ ] Run `ssh user@host` interactively in that pane; sidebar should drop local git status/grouping
- [ ] Exit the SSH session; local git status should return
- [ ] `ssh host htop` (non-interactive remote command) should not clear local git status
- [ ] `ssh-agent` / `scp` / `git` children should not be treated as hops

Made with [Cursor](https://cursor.com)


cloudy-liu:
Opened against upstream by mistake; intended only for cloudy-liu/tty7.

PR #761, opened 2026-08-31
feat(settings): add interface font family configuration in appearance typography

cybershape:
## Description

Adds an **Interface font family** configuration option under **Settings → Appearance → Typography**, allowing users to customize the UI font used across tabs, sidebars, modals, and settings panels.

### Changes
- **Core Config (`crates/tty7-core/src/core/config.rs`)**:
  - Added `pub ui_font_family: Option<String>` with `None` as default (system UI font).
  - Added sanitization to normalize blank font names to `None`.
  - Added unit tests for serialization and sanitization.
- **Theme (`src/ui/theme.rs`)**:
  - Applied `config.ui_font_family` to `gpui_component::Theme.font_family` in `apply_theme`.
- **Settings UI (`src/ui/settings.rs`)**:
  - Added `ui_font_select` dropdown under the Typography section directly after Interface font size.
  - Added search index entry for the new setting.
- **App (`src/ui/app.rs`)**:
  - Wired `ui_font_select` event subscriptions and added `commit_ui_font_family` to persist configuration and refresh windows.
- **i18n (`src/ui/i18n/`)**:
  - Added localization strings and search keywords for English, Simplified Chinese, and Japanese.

### Testing
- `cargo test --workspace` passed all tests.
- `cargo fmt --check` passed.


l0ng-ai:
Thanks — reviewed and merging. The wiring is right end to end (persist → apply → every open window → restart), so this only needed three small corrections, which I pushed to your branch as `d9847cf9` rather than sending you round again:

**Picking "Default" did nothing until the next launch.** `apply_theme` only assigned `Theme.font_family` when the config held `Some`, and `Theme::change` (running just above it) rewrites that field only when the active theme config names a face — which neither gpui-component's `default-theme.json` nor any tty7 preset does. So an override written once stuck for the life of the process: choose Courier, choose Default, `config.json` correctly goes back to `null`, and the whole chrome stays in Courier. Fixed by capturing the stock face in a `StockUiFont` global before anything can override it, then assigning `t.font_family` unconditionally in both directions, with a regression test.

**The first row of the dropdown said the wrong thing.** It reused `SettingsFontDefault` — "Default (match primary)" — which was written for the Bold/Italic selects, where "default" really does mean the primary terminal family. For the interface font it means the system UI font, so the row contradicted its own description two inches below it, in all three languages. Added `SettingsUiFontDefault` ("Default (system UI font)" / 默认（系统界面字体）/ デフォルト（システム UI フォント）) and made the shown label and the label `commit_ui_font_family` compares against come from the same function, so they can't drift apart later.

**`skip_serializing_if` on the config field.** `ui_font_family` was the only one of 23 `Option` fields that disappeared from `config.json` when unset; every other one is written as `null`, which is how someone hand-editing the file discovers a key exists. Dropped it, along with the redundant `#[serde(default)]` (the struct already carries one).

Also added the new key to `docs/reference/configuration.mdx` and `docs/customization/fonts.mdx`, which both enumerate every typography row.

Everything else checked out: the blank-name sanitization mirrors `link_file_command` and runs on the one production load path, old and new config files round-trip, the picker force-adds a saved-but-uninstalled family so the list and the config can't disagree, an uninstalled font degrades through gpui's fallback stack rather than panicking, the terminal grid's own font is untouched, and the choice survives a light/dark toggle and reaches windows opened later.

PR #766, opened 2026-09-01
feat(ssh): allow remote image clipboard writes

### LINKED ISSUE #765: Support OSC 5522 image clipboard writes over native SSH
### Type

### What’s going on?

Programs running on a remote host connected through tty7 native SSH currently have no standard way to place an image on the local machine’s system clipboard.

OSC 52 works well for text, but it does not carry MIME types, multipart image data, or a completion status. Kitty’s OSC 5522 clipboard protocol provides those semantics and allows a remote program to send PNG, JPEG, GIF, or WebP data as bounded chunks and wait for a result such as `DONE`, `EPERM`, or `EINVAL`.

The desired behavior is:

- Allow OSC 5522 image writes from native SSH panes.
- Keep the capability disabled by default and enable it per saved SSH host, because any remote program that writes terminal output could otherwise replace the local clipboard.
- Deliver completed image writes only to the controlling GUI, never observers.
- Keep clipboard protocol packets and image data out of scrollback and reconnect replay.
- Validate MIME type, image signature, decoded dimensions, and transfer size before updating the system clipboard.
- Return a status response to senders that include a request id.
- Clipboard reads can remain unsupported.

This enables workflows such as copying a remote screenshot or generated diagnostic image directly into a local chat, document, or image editor without downloading a temporary file manually.

### tty7 version

Upstream `main` before this capability; tested locally against tty7 26.8.3.

### Platform

macOS (Apple Silicon)

### Anything else?

A representative sender emits:

```text
OSC 5522 type=write id=<request-id>
OSC 5522 type=wdata id=<request-id> mime=<base64 MIME>;<base64 chunk>
...
OSC 5522 type=wdata id=<request-id>
```

and waits for:

```text
OSC 5522 type=write status=DONE id=<request-id>
```

A local implementation has been tested with a 16 MiB transfer limit, 4096-byte raw chunks, protocol capability probing, per-profile opt-in permission, and PNG/JPEG/GIF/WebP validation:

https://github.com/ayamir/tty7/commit/e59dc9e840925ca44777c5ca8cb0f8c2fcb28b60

Nested terminal applications also need to preserve this bidirectional protocol. The corresponding Herdr compatibility discussion is:

https://github.com/herdrdev/herdr/discussions/3484

ayamir:
## Motivation

Programs running over tty7 native SSH can use OSC 52 for text, but there is no MIME-aware, chunked, acknowledged path for copying a remote image to the local system clipboard. Kitty OSC 5522 provides those semantics.

Closes #765.

## Changes

- add a streaming OSC 5522 parser with capability probing, multipart image writes, request IDs, bounded chunks, and explicit status replies
- intercept clipboard packets in the daemon before scrollback/replay and forward completed writes only to the controlling GUI
- add a per-SSH-profile `remote_clipboard_write` permission, disabled by default and exposed under **Advanced → Security**
- validate PNG, JPEG, GIF, and WebP MIME/signature/dimensions before updating the local clipboard
- reject reads, primary-selection writes, malformed data, oversized transfers, stale-controller writes, and concurrent requests with explicit statuses
- bump the pane protocol to v6 and propagate the permission through spawn/attach/relink paths
- document configuration and include a remote Python sender that waits for `DONE`

## Safety

- maximum raw chunk: 4 KiB
- maximum encoded image transfer: 16 MiB
- maximum decoded dimensions: 16384 × 16384
- maximum decoder allocation: 256 MiB
- observers never receive clipboard side effects
- clipboard packets never enter scrollback or reconnect replay

## Validation

- `cargo test -p tty7-core clipboard --lib` (13 passed)
- `cargo test -p tty7 remote_clipboard --bin tty7-app` (3 passed)
- `cargo check --workspace --tests`
- local macOS arm64 bundle install and `codesign --verify --deep --strict`
- end-to-end native SSH PNG transfer to the macOS clipboard with a `DONE` response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Remote programs can copy images to the local clipboard over SSH.** A saved
+  host can opt into OSC 5522 clipboard writes under **Advanced → Security →
+  Remote clipboard images**. PNG, JPEG, GIF and WebP transfers are decoded out
+  of band, capped at 16 MiB, validated before they reach the system clipboard,
+  and never enter scrollback or replay after reconnecting. The permission is
+  off by default; clipboard reads and SVG writes remain unsupported.
+
 - **Documents dock beside the terminal** (#625). Opening a file, toggling the
   code panel or opening a diff no longer covers the workspace: the document
   takes a column to the right of the terminal — half the space between the
--- a/crates/tty7-core/src/client/pane.rs
+++ b/crates/tty7-core/src/client/pane.rs
@@ -192,6 +192,7 @@ impl PaneSession {
             // put a workspace back the way its user left it. Nothing that
             // scripts panes through this library has a screen to put back.
             restore: None,
+            allow_remote_clipboard_write: false,
         }
         .encode(&mut stream)?;
         let mut session = PaneSession::over(stream, 0)?;
@@ -224,7 +225,12 @@ impl PaneSession {
         size: WinSize,
         reply_wait: Duration,
     ) -> io::Result<PaneSession> {
-        ClientMsg::Attach { pane_id, size }.encode(&mut stream)?;
+        ClientMsg::Attach {
+            pane_id,
+            size,
+            allow_remote_clipboard_write: false,
+        }
+        .encode(&mut stream)?;
         PaneSession::checked(stream, "Attach", pane_id, reply_wait)
     }
 
--- a/crates/tty7-core/src/core/clipboard.rs
+++ b/crates/tty7-core/src/core/clipboard.rs
@@ -1,0 +1,699 @@
+//! Streaming OSC 5522 clipboard-write support.
+//!
+//! Clipboard packets are intercepted in the daemon before they reach the replay
+//! ring. Completed writes travel to the GUI as compact binary frames; only the
+//! GUI is allowed to touch the system clipboard.
+
+use base64::Engine as _;
+use base64::engine::general_purpose::STANDARD as BASE64;
+
+pub const MAX_CLIPBOARD_BYTES: usize = 16 << 20;
+const MAX_CHUNK_BYTES: usize = 4096;
+const MAX_OSC_PAYLOAD: usize = 8 << 10;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClipboardWrite {
+    pub mime: String,
+    pub data: Vec<u8>,
+    pub id: Option<String>,
+}
+
+impl ClipboardWrite {
+    pub fn encode_frame(self) -> Vec<u8> {
+        let mime = self.mime.as_bytes();
+        let id = self.id.as_deref().unwrap_or_default().as_bytes();
+        let mut out = Vec::with_capacity(4 + mime.len() + id.len() + self.data.len());
+        out.extend_from_slice(&(mime.len() as u16).to_le_bytes());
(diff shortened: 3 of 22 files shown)


l0ng-ai:
Thanks for this — it is careful work. I read the parser and the daemon path closely and could not break either: the byte limits really do bound allocation *before* it happens, the OSC buffer is capped and dropped to a scan state rather than grown, clipboard packets are stripped from `passthrough` before the ring so scrollback and reconnect replay stay clean, observers never receive `ClipboardWrite`, reads and primary-selection writes answer `ENOSYS`, MIME is matched against magic bytes before decode, and no tokenizer state fails to advance. The v6 bump is necessary and safe — `Attach`'s payload went from a 2-tuple to a 3-tuple and both peers gate on exact equality, so no old peer half-parses it, and `SshProfile`'s container-level `#[serde(default)]` means existing configs read the new field as `false` rather than silently granting it.

One blocker, though, which I've fixed on your branch as `c5700d60` so this can go in:

**The profile's grant was silently revoked on the first re-attach, so the feature died after any app restart.** The permission had two homes and both said no. `attach_subscriber_with_permissions` overwrote `st.allow_remote_clipboard_write` with the client-supplied bool unconditionally, discarding the value `spawn_native_ssh` derived from the authoritative `NativeSshSpec`; `PaneRoute::allow_remote_clipboard_write()` returns `false` for `PaneRoute::Local`, and a native SSH pane's route *is* `Local` (`spawn_native_ssh_once` dials through `connect()` and leaves it that way), so every `Attach` for such a pane sent `false`; and the client-side gate keyed on `self.ssh_spec`, which only `from_native_ssh_parts` ever sets, so a pane restored by attaching to its id never got one back. Enable the permission, connect, copy an image — works. Quit and reopen tty7, and because the pane outlived the window the restore takes the attach-by-id branch: from then on every copy answers `EPERM`, the switch still reads "on", and nothing is logged.

The fix pins the spec's answer in `PaneState` and routes both attach and detach through one `set_clipboard_permission`, so a pane carrying a spec keeps that spec's answer whatever an attaching client claims, and a pane without one — everything on a remote `tty7-server` — is exactly as permitted as its controller says. That also closes the reverse hole: a client can no longer attach with `true` to a pane whose profile said no. On the client side the gate now refuses only what the pane can see is forbidden and defers to the daemon otherwise.

Two smaller things in the same commit:

- `Parser::fail` set `failed = true` but kept `state.data`, and `feed_data` then returns `None` for every later packet including the terminator — so the `WriteState` was never dropped. A sender that fails its last chunk on purpose could park `MAX_CLIPBOARD_BYTES` in every pane it can reach, for as long as it liked. The reply now takes the bytes with it.
- The capability probe answered `CSI ? 5522 ; 2 $ y` unconditionally. Per DECRPM that is "recognised but reset", so a sender honouring its own probe concluded the feature was off even on a host the user had opted into, and never sent. It now reports the permission actually in force — 1 when granted, 2 when not — which keeps "not supported" (0) distinguishable from "not permitted here".

Non-blocking observations, left alone: a second `type=write` replaces the first with no reply at the daemon, so the superseded sender waits out its own timeout (the GUI does answer `EBUSY`); the stated 16384×16384 dimension cap never binds, because `max_alloc = 256 MiB` refuses anything past roughly 8192² RGBA first; validation fully decodes and discards the pixels and then hands gpui the encoded bytes to decode again; and `encode_frame` writing `id.len() as u16` is safe only because `MAX_OSC_PAYLOAD` bounds it, which is an implicit coupling between two constants.

Not verified here: the image actually landing on the macOS pasteboard over a real SSH link, and the Windows leg — CI covers the latter.

PR #777, opened 2026-09-06
fix(remote): keep the reason a remote server failed to start (#774)

l0ng-ai:
Closes part of #774.

A remote workspace could get stuck in a loop nobody could get out of: every reconnect
failed with `started but nothing was answering on the control socket after 15s`, the
status strip showed a copy bar frozen at 100%, and no button was offered. This fixes
the mechanism behind the loop, keeps the reason a failed start had to give, and stops
the strip and the progress bar from running off the window.

Scope is the connection failure and the UI that reports it. The Vim `:wq` cursor and
the btop re-attach items in #774 are not touched.

## What changed

| | Before | After |
|---|---|---|
| Daemon cannot bind the control socket, and nothing else is serving | Logs one line and **keeps running**, holding the single-server lock | Exits non-zero with the reason, releasing the lock |
| Deciding "is someone else already serving control?" | Never asked — any bind error was survivable | Asked by connecting; `AddrInUse` alone does not answer it |
| Daemon's stdout/stderr on launch | `> /dev/null 2>&1` | `<binary>.startup.log`, exit status in `<binary>.startup.exit`, both `0600` |
| Which recorded exit status counts as this launch's | — | Only one stamped with this launch's nonce, so a restart never reads the outgoing daemon's status as the incoming one's |
| A daemon that exits **0** (another server holds the lock) | — | Not a failure: keep probing for the server that is there |
| A home that cannot hold the two files | — | Redirect falls back to `/dev/null` and the daemon still starts, as it did before |
| Startup readiness probe result | `out.success()` only; status and stderr dropped | Failure reason kept and reported |
| A daemon that dies at startup | Polled for the full 15s, then reported the timeout | Fails as soon as the exit status is there (sub-second) |
| The error a failed start gives | `… nothing was answering … after 15s` | Exit status, the probe's stderr, the last lines the server logged, and the path to the full log |
| Caption after the upload reaches 100% | Stays on "Copying tty7's server… 8.8 MiB / 8.8 MiB" through the startup wait | Moves to "Starting tty7's server…" |
| Install progress after an **automatic** reconnect attempt | Never cleared — froze the bar and **suppressed the strip's button**, including Update Server | Cleared when the attempt starts and when it finishes, as the manual connect already did |
| Status strip with a long error | Card grew with the message: 1978px wide in a 1440px window, so the text and the retry button were off-screen | Card has its own width (560px, capped at 92% of the window); the message wraps, the button never shrinks |
| The two status strips (home, floating) | Two near-identical copies | One `status_card` / `status_row` / `status_message` |

## Why the daemon change is the root of the loop

`bind` reports `AddrInUse` for *any* path that already exists — a live server, a socket
a crash left behind, a directory in the way, a file this user may not remove. So the
errno cannot tell "someone else is serving" from "nobody is". Reading it as the former
kept a daemon alive that answered nothing, and a live daemon holds the single-server
lock, so every later `--daemon` stood down at once:

```mermaid
sequenceDiagram
    participant C as client
    participant D1 as first --daemon
    participant D2 as later --daemon
    C->>D1: launch
    D1->>D1: bind control.sock → AddrInUse
    Note over D1: before: logs it to /dev/null,<br/>keeps the lock, serves nothing
    C->>D1: --stdio --bridge
    D1--)C: Connection refused
    C->>D2: launch again
    D2--)C: "another server already serves<br/>this config dir" (exit 0)
    Note over C,D2: probe fails forever, nothing ever changes
```

After the change the first daemon exits, releasing the lock, so the next attempt is a
real attempt — and the reason reaches the client instead of `/dev/null`.

## Notes on the shell preamble

`umask 077` is scoped to a subshell, not written bare. Bare, it applies to the whole
SSH login shell, so `tty7-server --daemon` inherits it and so does every pane shell the
daemon forks — a `git clone` in a remote pane would produce files nobody but the owner
can read. Measured on a real box: bare gives `drwx------` / `-rw-------` for
pane-created paths, scoped gives the user's own `drwxrwxr-x` / `-rw-rw-r--`, and the two
diagnostic files stay `0600` either way.

The files are removed and re-created rather than only truncated, because `>` on an
existing file keeps whatever mode it already had. And the whole attempt sits in a
subshell with a `/dev/null` fallback: on a POSIX shell a redirection error on the
special builtin `:` ends a non-interactive shell outright, and `|| true` never runs.
Verified under both `sh` and `dash` with an unwritable log directory — the daemon still
launches, it just has no diagnostics.

## Testing

Unit tests (all green: `tty7-core` 1242, desktop 1711, `tty7-server`/`tty7-cli` suites, fmt clean):

- `a_daemon_that_died_at_startup_fails_at_once_and_in_its_own_words` — the exit status, the
  server's own words, the log path, and at most two probes for the whole wait.
- `a_daemon_that_never_answers_names_the_probe_and_the_log` — the timeout branch still names
  the probe's stderr, which used to be dropped for its exit code.
- `the_launch_command_keeps_what_the_daemon_says_and_how_it_ends` — no `/dev/null`, private
  files, previous status cleared first.
- `the_caption_moves_on_once_the_bytes_are_across` — the caption's last word is the phase
  actually in progress.
- `only_a_connect_tells_a_live_server_from_a_blocked_path` — a live server and a directory
  standing where the socket goes give the *same* `AddrInUse`; only a connect separates them.
- `a_status_from_the_launch_before_is_not_this_launch_s_answer` — the nonce, against a
  restart's late-writing predecessor.
- `a_daemon_that_stood_down_cleanly_is_not_a_failed_start` — status 0 keeps the probe going.
- `a_finished_attempt_retires_the_install_it_was_running` — red before the change.
- `a_long_failure_wraps_instead_of_stretching_the_card` — red before the change (card laid out
  1978px wide in a 1440px window).
- `the_copy_bar_stays_inside_the_home_strip` — a guard, not a reproduction; see the note below.

On a real Linux box (`java@java`, x86_64, isolated `--config-dir`, the machine's own daemon
untouched throughout):

- Reproduced the loop with the **old** build: the daemon stayed alive after the bind failed,
  the next `--daemon` printed "another server already serves this config dir" and exited 0,
  and `--stdio --bridge` returned `Connection refused`.
- Same setup with **this** build: exits 1 with `Address in use (os error 98)`, and the next
  launch reaches the bind and reports the real reason rather than standing down.
- Clean config dir: binds `control.sock` (`srw-------`), probe exits 0 — the happy path is
  unaffected.
- Another process already answering on the control path: the daemon stays up and serves panes
  only, saying why. This is the local shape where the GUI hosts the control listener.
- The generated launch command, run verbatim: `.startup.exit` holds this launch's nonce and `1`,
  `.startup.log` holds the three lines the server printed, both `0600` even when a predecessor
  left them at `0664`; a still-running daemon leaves no exit status, and one killed later
  records `143`. A status left by another launch is truncated away and, if it lands late,
  rejected on its nonce.
- The umask and fallback checks described above, under `sh` and `dash`.

A `/code-review high` pass over the diff found the umask leak, the status-0 false failure, the
restart status race, a progress entry the pane route would never clear, the `dash` redirection
trap, and `is_first_install` counting `.startup.log` as an installed server — all fixed above,
with the first re-measured on the box.

## Known limits

- #774's first screenshot shows the progress bar escaping its card. That does not reproduce:
  `w_full` resolved against the card at every window width tried, before this change and after.
  The card now has a width of its own, so the containment holds by construction and there is a
  test on it — but it is a guard, not the reproduction of a failure.
- The exact cause of the reporter's incident is still unproven. This fixes a mechanism that
  produces exactly its symptoms and that reproduces on a real machine; it does not claim to be
  what happened on theirs. The startup log this adds is what would settle it.

https://claude.ai/code/session_015q6HRem76HYy33T39bp34c

### DIFF
--- a/crates/tty7-core/src/daemon/install/mod.rs
+++ b/crates/tty7-core/src/daemon/install/mod.rs
@@ -301,8 +301,17 @@ pub fn install_confirm() -> Arc<dyn InstallConfirm> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum InstallPhase {
-    Downloading { done: u64, total: Option<u64> },
-    Uploading { done: u64, total: u64 },
+    Downloading {
+        done: u64,
+        total: Option<u64>,
+    },
+    Uploading {
+        done: u64,
+        total: u64,
+    },
+    /// The bytes are there and the far end is being waited on. Both a first
+    /// install and a restart end here, which is why its caption says starting
+    /// rather than restarting.
     Restarting,
 }
 
@@ -777,6 +786,18 @@ impl<'a> Installer<'a> {
                     let (confirmed, _) = self.install(asset, &paths)?;
                     report.installed = true;
                     report.confirmed = confirmed;
+                    // The upload's caption was the last thing reported, so the
+                    // strip sat on "copying… 100%" through the startup wait —
+                    // and when the wait failed, that is the frame the user was
+                    // left looking at.
+                    //
+                    // Reported here rather than in the wait itself: an
+                    // `ensure_daemon` with nothing to install has to stay
+                    // silent. A pane route quietly restarting a dead remote
+                    // daemon goes through that path, and nothing on it retires
+                    // a progress entry — one left there freezes the strip and
+                    // takes its button away for the rest of the session.
+                    install_progress().report(&self.host, InstallPhase::Restarting);
                 }
             }
         }
@@ -915,9 +936,21 @@ impl<'a> Installer<'a> {
         ReleaseDownload { fetch }.load_with_progress(&self.version, asset, &on_progress)
     }
 
+    /// Whether this machine has never had a tty7 server on it — the question
+    /// the consent prompt turns on.
+    ///
+    /// A launch leaves `tty7-server-c8p6.startup.log` and `.startup.exit`
+    /// beside the binary, and both begin the same way it does. Counting one as
+    /// a server would let someone who deleted the binary to uninstall tty7 get
+    /// a new one downloaded and written without ever being asked.
     fn is_first_install(&self, paths: &RemotePaths) -> bool {
+        let a_server = |name: &String| {
+            name.starts_with("tty7-server-")
+                && !name.ends_with(STARTUP_LOG_SUFFIX)
+                && !name.ends_with(STARTUP_EXIT_SUFFIX)
+        };
         match self.ops.list_dir(&paths.bin_dir) {
-            Ok(Some(entries)) => !entries.iter().any(|name| name.starts_with("tty7-server-")),
+            Ok(Some(entries)) => !entries.iter().any(a_server),
             Ok(None) => true,
             Err(_) => true,
         }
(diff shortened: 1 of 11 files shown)


shihuaidexianyu:
所以这个相当于是对我的那个修复的另一种更简单的实现？

PR #778, opened 2026-09-06
feat(ui): add bindable close-window action

### LINKED ISSUE #773: Feature: bindable CloseWindow action for the tray-retire path
## Summary

26.9.0's tray-retire model (commit `2e6103c`, #639) makes closing
the last window the "keep the daemon, drop the UI process weight"
gesture. That path is only reachable from the OS red close button
today — there is no `CloseWindow` action, no palette entry, no
keybind. `⌘W` is bound to `CloseActiveTab` (`Close Pane / Tab`,
`src/ui/keymap.rs:1219`), `⌘H` hides but keeps the window in
memory, and `⌘Q` stops the daemon.

For users who *want* the retire-to-tray path from the keyboard —
close the window, free its memory, leave shells running on the
daemon — there is currently nothing to bind.

## Proposed change

Add a `CloseWindow` action:

- Route: `CloseWindow` → the same code path `on_window_should_close`
  runs when the red button is clicked (`src/ui/app.rs:1397`) →
  `close_window_for(cx, workspace)` (`src/ui/windows.rs:640`), so
  tray-retire semantics are identical.
- Palette: register a `Command::localized(L10nKey::CmdCloseWindow,
  CloseWindow)` entry so it's discoverable and command-frecency
  can find it.
- Default binding: none (leave the slot free for the user). `⌘⇧W`
  is a natural fit and is unused on macOS; `⌘⇧K` is another
  option.

## Motivation

- Completes the keyboard vocabulary for the window-lifecycle model
  the release just introduced.
- Costs nothing to users who don't bind it — no default behavior
  change.
- The window-close path already exists (`close_window_for`); this
  is a name + palette entry + action dispatcher, not a new
  subsystem.

## Impact

Small: one action, one palette registration, one dispatcher arm.
No default-behavior regression.

bytehello:
Closes #773.

## What changed

- adds a bindable `CloseWindow` action with no default shortcut
- exposes Close Window in the command palette and Keybindings UI
- routes native and action-driven closes through the same session cleanup and tray-retire logic
- adds English, Simplified Chinese, and Japanese labels

## Validation

- `cargo fmt --check`
- `cargo test --locked keymap` (29 passed)
- `cargo test --locked palette` (20 passed)

The build emits only pre-existing unused-code warnings outside this change.


l0ng-ai:
Reviewed and merging. The action itself is right, including the two calls it would have been easy to get wrong:

- Routing through `on_window_should_close`'s logic rather than `close_window_for` (which the issue suggested) is the correct read — `close_window_for` recycles the last window onto a fresh workspace, which is what workspace deletion wants, not what closing a window wants. Extracting `prepare_window_close` so both paths share the tray-retire decision is exactly right.
- `remove_window()` cannot re-enter `should_close` on any backend: gpui drops the window at the end of the update trail, Windows reaches it via `DestroyWindow` → `WM_DESTROY` (not `WM_CLOSE`), and the macOS backend clears the delegate before closing. So no double `detach_workspace`. Also checked that `QuitMode::Explicit` keeps gpui's `quit_on_empty` out of the picture on Windows/Linux.

I pushed one follow-up commit (fa69956a) rather than send it back for a round trip:

1. **A test.** Everything about the action was a table entry, and every one of those can be filled in while the action reaches nothing. `close_window_action_tests` dispatches for real and asserts both that the window left gpui and that it left the `WindowRegistry` — the second is what says it ran the same cleanup the red button runs. Both assertions were mutation-checked.
2. **`docs/reference/keyboard-shortcuts.mdx`.** The "Actions with no default key" table is what a user reads before binding one, so an action missing from its Application row reintroduces the discoverability complaint #773 was about.
3. **Palette placement + subtitle.** Close Window was at the top of Application, above Settings, while `Quit` at the bottom carried the subtitle "shells keep running" — which is Close Window's promise and is flatly wrong on Quit, whose own confirmation says "anything running in your shells is terminated" and which calls `daemon::spawn::stop()` to make that true. Moved the pair together, gave the promise to the command that keeps it, and corrected Quit's.

One thing deliberately left out of scope: there's no File-menu entry, so on macOS "Close Window" is still absent from the menu bar where a Mac user would look for it first. Worth a follow-up if you're interested.

Thanks — clean change, and the comment you carried over into `prepare_window_close` intact is appreciated.

https://claude.ai/code/session_01LKMZVh6mUBxXAn6v7P6JC6

PR #784, opened 2026-09-07
fix(keymap): fold a shift-punctuation chord into the key the platform reports (#750)

### LINKED ISSUE #750: Settings key-recorder can't bind `cmd-shift-[` / `cmd-shift-]` for Prev/Next Tab on macOS
## Summary

On macOS, rebinding any action to a `cmd-shift-[punct]` chord in
Settings → Keybindings does not work. Tested with
`NextTab`/`PrevTab` bound to `cmd-shift-]`/`cmd-shift-[`: the chord
never fires the assigned action, regardless of what the config
stores.

## Repro

1. Settings → Keybindings → click `NextTab`, press `cmd-shift-]`.
2. Press `cmd-shift-]` in the app.

Observed: the workspace switcher opens (unclear why — no default
in `src/ui/keymap.rs` binds `cmd-shift-]` on macOS). The assigned
action never fires.

## Evidence

`~/.config/tty7/config.json` after recording — recorder captured:

```json
"NextTab": "secondary-}",
"PrevTab": "secondary-{"
```

Shift modifier dropped; shifted glyph kept. Hand-editing to every
plausible form fails to fire `NextTab`:

| spec                    | fires `NextTab`? |
|-------------------------|------------------|
| `secondary-}`           | no               |
| `secondary-shift-]`     | no               |
| `secondary-shift-}`     | no               |

## Root cause

Two bugs stack:

**1. `spec_from_keystroke` (`src/ui/keymap.rs:893`)** appends
`ks.key` verbatim. On macOS gpui reports `cmd-shift-]` as
`key="}", modifiers.shift=false` (character already reflects the
shifted layer), so the stored spec drops the modifier and keeps
`}`. The recorder writes `secondary-}`.

**2. Dispatch doesn't match any manual variant.** Since none of
`secondary-}`, `secondary-shift-]`, `secondary-shift-}` fire, the
dispatcher's canonicalization for shifted-punct keys on macOS
isn't reachable from the config format. That's the real blocker:
even hand-edited configs can't bind these chords.

The unexplained switcher-open on `cmd-shift-]` is a secondary
mystery — no default in `keymap.rs` binds these chords on macOS —
but not the main bug.

## Fix direction

- In `spec_from_keystroke`, use gpui's unshifted-key field (or
  invert-map shifted-punct chars back to the physical key + set
  `shift-`) so the spec always carries the physical key.
- Verify the parse side (`gpui::Keystroke::parse`) accepts that
  spec and that dispatch matches the raw event. Add a round-trip
  test alongside `spec_from_keystroke_round_trips_through_parse`
  (`src/ui/keymap.rs:1671`) covering shifted-punct on macOS.

## Impact

`cmd-shift-[/]` is the universal macOS tab-cycling shortcut
(Safari, Chrome, Finder, Terminal.app, VS Code). Right now no
config path binds them.

l0ng-ai:
Fixes #750

## What is actually wrong

The recorder is not the bug. Every gpui backend carries Shift as a modifier
for **letters only**; over the main row's digits and punctuation it spends the
modifier on the character and clears the flag:

- macOS `crates/gpui_macos/src/events.rs:469-480` — `else if shift { shift = false;
  chars_with_shift }`
- Windows `crates/gpui_windows/src/keyboard.rs:148-160` — `get_keystroke_key` +
  `need_to_convert_to_shifted_key` (`:175-203`, which lists `VK_OEM_*` and `VK_0..VK_9`)
- Linux `crates/gpui_linux/src/linux/platform.rs:997-1003` — "we only include the
  shift for upper-case letters by convention, so don't include for numbers and
  symbols"

So `⌘⇧]` is delivered as `key: "}"`, `shift: false`, and the recorder writing
`secondary-}` is writing down exactly what it was handed. What it cannot reach
is a binding spelled the other way.

Two of our own defaults are spelled the other way:

```rust
("FocusNextPane", per_platform("secondary-]", "secondary-shift-]")),
("FocusPrevPane", per_platform("secondary-[", "secondary-shift-[")),
```

Off macOS those install a binding nothing can press — **Ctrl+Shift+] and
Ctrl+Shift+[ have not cycled panes on Windows or Linux at all**. It is the
same reason the reporter's hand-edited `secondary-shift-]` changed nothing.

## The fix

`push_binding` lays down the spec **as written**, and the folded chord
**beside** it when the two differ. A US-layout table maps the unshifted key to
its shifted glyph and drops the flag.

Adding rather than substituting is the load-bearing part, because the
convention is *not* universal — see below. Since the fold only ever adds a
binding, it cannot unbind anything, whatever the backend does with Shift.
Still working after this:

| spelling              | source                     |
|-----------------------|----------------------------|
| `secondary-}`         | what the recorder writes for the main-row key |
| `secondary-shift-]`   | the default table, and hand-written configs |
| `secondary-shift-}`   | the third form people try; the flag is just redundant |
| `secondary-shift-/`   | what the recorder writes for Ctrl+Shift+numpad-`/` on Windows |
| `secondary-shift-t`   | shift over a letter — no fold |
| `shift-enter`, `shift-tab`, `shift-insert`, `ctrl-shift-up` | named keys have no shifted half — no fold |

`assign_keybinding` and the default-chord dedup test use `same_chord`, which
asks whether two specs claim one keystroke; comparing them as text, a recorded
`secondary-}` never displaces a config's `secondary-shift-]` and both survive
onto one keystroke, where which fires is arbitrary.

macOS behaviour stays behind the existing `secondary` abstraction —
`spec_from_keystroke` already writes it and `Keystroke::parse` already reads
it, so the fold round-trips through both without a new platform fork.

## Why the fold adds instead of substitutes

An earlier revision of this PR replaced the written spec with the folded one
and claimed that "can only turn a dead spec into a live one". That was too
broad, and review caught it. **Windows leaves Shift intact on the numpad's
`/ * + -`**: `need_to_convert_to_shifted_key` does not list `VK_DIVIDE`,
`VK_MULTIPLY`, `VK_ADD` or `VK_SUBTRACT`, so `get_keystroke_key` falls through
to `get_key_from_vkey` (`keyboard.rs:162-172`), and
`MapVirtualKeyW(VK_DIVIDE, MAPVK_VK_TO_CHAR)` returns `/` — verified by probe
on Windows 11. Ctrl+Shift+numpad-`/` therefore really is delivered as
`ctrl-shift-/`; the recorder writes `secondary-shift-/` and it dispatches
today. Substituting `secondary-?` for it would have unbound a working chord.

The other two backends differ again, so no glyph list can separate the cases:

- **macOS** clears the flag for numpad keys anyway — the numpad glyph is the
  same shifted or not, so `chars_with_shift` takes the branch that drops it.
- **Linux** names them `divide` / `multiply` / `add` / `subtract`
  (`platform.rs:970-971`), which are not one character and never reached the
  fold at all.

Installing both spellings sidesteps the ambiguity entirely.

## How this was verified

The chord cannot be pressed on the machine this was written on (Windows), so
it is verified by unit test rather than by hand. `cargo test --bin tty7-app
keymap` — **33 passed**; full `cargo test --bin tty7-app` — 1471 passed, 0
failed. Target dir private to the worktree.

Five new tests, none of them platform-gated:

- `a_shifted_punctuation_chord_dispatches_however_it_is_spelled` — builds the
  keystroke a backend really delivers (`}`, no shift), asserts
  `spec_from_keystroke` writes `secondary-}`, then asserts all three spellings
  dispatch `NextTab` for it. This is the round trip: record → serialize →
  parse → dispatch.
- `the_default_pane_cycle_chord_reaches_its_action` — the real
  `default_bindings()` table against the delivered keystroke. **This one fails
  on `main` on Windows**, which is the Ctrl+Shift+] regression above.
- `a_chord_the_platform_delivers_with_shift_intact_stays_bound` — the numpad
  case: `secondary-shift-/` still dispatches for a `ctrl-shift-/` keystroke,
  *and* for the main row's `?`. **Fails against the substituting fold**, which
  is how the review's finding was confirmed.
- `two_spellings_of_one_chord_are_recognised_as_one` — `same_chord`.
- `folding_leaves_every_other_spelling_alone` — the untouched column of the
  table, plus per-chord folding inside a `ctrl-b shift-5` prefix sequence.

Each behavioural test was confirmed to fail with its fix reverted and pass with
it restored.

## The unexplained part of the report

The reporter also saw the workspace switcher open on `⌘⇧]`. Nothing in
`default_bindings()` can catch that chord — `ToggleSwitcher` is
`secondary-shift-o`, which is a letter and so keeps its shift flag — and
nothing else in the tree binds `}` outside the tmux prefix sequence. That part
is not explained by this change and I did not invent an explanation for it.

### DIFF
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -6659,16 +6659,15 @@ impl Tty7App {
     }
 
     fn assign_keybinding(&mut self, action: String, spec: String, cx: &mut Context<Self>) {
-        // Compared as the keymap will build them, not as they are spelled: a
-        // recorded `secondary-}` and a config's `secondary-shift-]` are one
-        // chord, and only the folded form tells you so. Left unfolded, the
-        // displacement never fires and both bindings survive onto the same
-        // keystroke, where which one wins is arbitrary (#750).
-        let chord = crate::ui::keymap::dispatchable_spec(&spec);
+        // Compared as chords, not as spellings: a recorded `secondary-}` and a
+        // config's `secondary-shift-]` are one keystroke written two ways, and
+        // only `same_chord` sees it. Compared as text, the displacement never
+        // fires and both bindings survive onto that keystroke, where which one
+        // wins is arbitrary (#750).
         let displaced = crate::ui::keymap::effective_bindings(cx)
             .into_iter()
             .chain(crate::ui::keymap::extra_bindings(cx))
-            .find(|(a, k)| crate::ui::keymap::dispatchable_spec(k) == chord && *a != action)
+            .find(|(a, k)| *a != action && crate::ui::keymap::same_chord(k, &spec))
             .map(|(a, _)| a);
         // A trailing "…" on an action name marks a command that opens
         // something; it is not punctuation, and inside a sentence it reads as
--- a/src/ui/app.rs
+++ b/src/ui/app.rs
@@ -6659,10 +6659,16 @@ impl Tty7App {
     }
 
     fn assign_keybinding(&mut self, action: String, spec: String, cx: &mut Context<Self>) {
+        // Compared as the keymap will build them, not as they are spelled: a
+        // recorded `secondary-}` and a config's `secondary-shift-]` are one
+        // chord, and only the folded form tells you so. Left unfolded, the
+        // displacement never fires and both bindings survive onto the same
+        // keystroke, where which one wins is arbitrary (#750).
+        let chord = crate::ui::keymap::dispatchable_spec(&spec);
         let displaced = crate::ui::keymap::effective_bindings(cx)
             .into_iter()
             .chain(crate::ui::keymap::extra_bindings(cx))
-            .find(|(a, k)| *k == spec && *a != action)
+            .find(|(a, k)| crate::ui::keymap::dispatchable_spec(k) == chord && *a != action)
             .map(|(a, _)| a);
         // A trailing "…" on an action name marks a command that opens
         // something; it is not punctuation, and inside a sentence it reads as
--- a/src/ui/keymap.rs
+++ b/src/ui/keymap.rs
@@ -935,6 +935,88 @@ pub(crate) fn spec_from_keystroke(ks: &Keystroke) -> Option<String> {
     Some(spec)
 }
 
+/// The glyph a US keyboard prints on the shifted half of every key that is not
+/// a letter — `shift-]` types `}`, `shift-2` types `@`.
+const SHIFTED_GLYPHS: [(char, char); 21] = [
+    ('`', '~'),
+    ('1', '!'),
+    ('2', '@'),
+    ('3', '#'),
+    ('4', '$'),
+    ('5', '%'),
+    ('6', '^'),
+    ('7', '&'),
+    ('8', '*'),
+    ('9', '('),
+    ('0', ')'),
+    ('-', '_'),
+    ('=', '+'),
+    ('[', '{'),
+    (']', '}'),
+    ('\\', '|'),
+    (';', ':'),
+    ('\'', '"'),
+    (',', '<'),
+    ('.', '>'),
+    ('/', '?'),
+];
+
+/// Rewrites one chord into the shape the platform actually delivers, or
+/// returns `None` when it is already that shape.
+///
+/// Every gpui backend carries Shift as a modifier for letters only. Over the
+/// digits and the punctuation it spends the modifier on the character instead
+/// and clears the flag: that is the `chars_with_shift` branch on macOS,
+/// `need_to_convert_to_shifted_key` in the Windows mapper, and Linux's "we
+/// only include the shift for upper-case letters by convention". So the chord
+/// written `secondary-shift-]` arrives as `secondary-}` and the two never
+/// meet — the binding installs and is unreachable, which is why Ctrl+Shift+]
+/// stopped cycling panes off macOS and why hand-editing a config to that
+/// spelling changed nothing (#750).
+///
+/// Folding the spec the same way on the lookup side closes the round trip.
+/// The recorder already writes `secondary-}`, because that is what
+/// `spec_from_keystroke` was handed; a default or a config written the other
+/// way now installs the very same binding, so both spellings keep working and
+/// nobody's config comes undone. `secondary-shift-}`, which people also reach
+/// for, folds onto it too — the flag is simply redundant there.
+///
+/// The table is US-layout, and a chord it does not know is left alone rather
+/// than guessed at: this can only turn a dead spec into a live one.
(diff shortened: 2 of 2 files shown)


l0ng-ai:
Review is right, and the fix is pushed in 71f53c8a.

Verified against the fork before changing anything. Windows `get_keystroke_key` (`gpui_windows/src/keyboard.rs:148-160`) only strips Shift when `need_to_convert_to_shifted_key` (`:175-203`) claims the vkey, and that list has the OEM keys and `VK_0..VK_9` but not `VK_DIVIDE`/`VK_MULTIPLY`/`VK_ADD`/`VK_SUBTRACT`. Those fall through to `get_key_from_vkey` (`:162-172`) with the flag untouched, and I probed `MapVirtualKeyW` on this machine to confirm the character it lands on:

```
VK_DIVIDE    vk=0x6F -> MAPVK_VK_TO_CHAR = 0x002F '/'
VK_MULTIPLY  vk=0x6A -> 0x002A '*'
VK_ADD       vk=0x6B -> 0x002B '+'
VK_SUBTRACT  vk=0x6D -> 0x002D '-'
```

So Ctrl+Shift+numpad-`/` really is delivered as `ctrl-shift-/`, the recorder writes `secondary-shift-/`, and the substituting fold turned it into `secondary-?` — a working binding made dead. Exactly the class of bug the PR set out to fix.

The other two backends do it differently again, which is what rules out fixing this with a smarter glyph list:

- **macOS** clears the flag for numpad keys anyway — the numpad glyph is identical shifted or not, so `chars_with_shift` takes the branch that drops Shift (`gpui_macos/src/events.rs:469-480`).
- **Linux** names them `divide`/`multiply`/`add`/`subtract` (`gpui_linux/src/linux/platform.rs:970-971`) — not one character, so they never reached the fold.

No table over glyphs can separate main-row `/` from numpad `/`; gpui's own Windows mapper has the same ambiguity, since `get_vkey_from_key("/")` can only answer `VK_OEM_2`. So the fix stops choosing between the two spellings: `push_binding` installs the spec as written **and** the folded chord beside it. The fold can now only ever add a keystroke the dispatcher can reach, which makes the claim you challenged true as stated rather than merely narrower.

`assign_keybinding` and the dedup test moved to `same_chord`, which asks whether two specs claim one keystroke instead of comparing folded text.

Regression test added: `a_chord_the_platform_delivers_with_shift_intact_stays_bound` asserts `spec_from_keystroke` writes `secondary-shift-/` for that keystroke, that the spec still dispatches for it, and that the main row's `?` reaches the same action. It fails against the substituting fold and passes against this one — that is how I confirmed your finding rather than just taking it. `cargo test --bin tty7-app keymap`: 33 passed; full bin suite 1471 passed, 0 failed. PR body corrected.

PR #861, opened 2026-09-10
fix(update): explain GitHub API rate limits

ayamir:
## Summary

- recognize GitHub API rate-limit responses from both response headers and the JSON error message
- tell users approximately how many minutes remain before the limit resets
- include short, sanitized GitHub JSON messages for other HTTP failures while preserving the status code
- cap error-body reads at 8 KiB and retain the original HTTP error if the body cannot be read

## Motivation

The update checker currently reduces every non-successful GitHub response to a message such as `GitHub returned HTTP 403`. In particular, unauthenticated GitHub API requests share a 60-request hourly allowance per public IP, so users behind a shared proxy or NAT can hit the limit without knowing why or when retrying will work.

## Validation

- `cargo test --bin tty7-app core::update::tests` (28 passed)
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/src/core/update.rs
+++ b/src/core/update.rs
@@ -1904,7 +1904,9 @@ fn github_http_error(
     let message = serde_json::from_slice::<GitHubError>(body)
         .ok()
         .map(|error| sanitize_github_message(&error.message));
-    let rate_limited = status == 403
+    // 403 is what the REST API has always answered a spent quota with; 429 is
+    // what it increasingly answers instead, and both carry the same headers.
+    let rate_limited = matches!(status, 403 | 429)
         && (rate_remaining == Some("0")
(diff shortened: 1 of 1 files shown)


l0ng-ai:
Thanks — this reads much better than the bare status code, and the retry line is the part that actually helps.

I pushed one commit on top: GitHub's REST API answers a spent quota with 403 **or** 429 depending on the endpoint, and both carry the same `x-ratelimit-*` headers. Only the 403 spelling reached the retry advice, so a 429 said the quota was gone without saying when it comes back. `matches!(status, 403 | 429)` plus a test for it.

Merging once CI is green.

https://claude.ai/code/session_01E4EPKzHg1fm9HMmHkUYpER

PR #873, opened 2026-09-11
feat(agents): add Qoder CLI integration

WhiteDG:
Qoder CLI is now a first-class agent alongside Claude, Copilot, Grok, and the others. The integration covers hook lifecycle, session commands, status tracking, and UI presentation.

## What changed

**Hook events** map Qoder's lifecycle to tty7's state machine: session start, prompt submit, permission requests, MCP tool elicitation (an authorized MCP tool can still pause for user input mid-call), tool completion, stop, and session end. Compaction events are filtered out — Qoder emits a session-start after compacting the active turn, which would reset the status line to Idle without this filter, even though the turn is still running.

**Settings path resolution** respects `QODER_CONFIG_DIR` for local installs, falling back to `~/.qoder/settings.json`. Remote targets ignore the override (a local env var must not redirect remote hooks).

**Session commands** support `--resume` and `--fork-session`. The resume command strips conflicting flags (`--resume`, `-r`, `--continue`, `-c`, `--session-id`, `--worktree`, `--fork-session`) from the original launch argv before appending the new session id. The `-w`/`--cwd` flags survive (Qoder's `-w` means `--cwd`, not `--worktree`). Both commands require session persistence: when `--no-session-persistence` is present, there is no saved conversation to reopen, so the commands return None.

**Localization** complete for en/ja/zh. Icon embedded, search keywords wired.

## Testing

- `cargo test -p tty7-core --lib core::agent_hooks::tests` — 30 passed, 0 failed. Three new tests: `qoder_compaction_preserves_the_active_turn` (verifies the compaction filter), `qoder_mcp_elicitation_waits_for_user_input` (validates MCP tool question handling), and `qoder_config_dir_controls_local_hook_lifecycle` (multi-case test confirming env var behavior with isolation).
- `cargo test -p tty7-core --lib core::cli_agent::tests` — 30 passed, 0 failed. Three new tests covering resume/fork command generation, worktree handling, and persistence requirements.
- `cargo fmt` clean.


l0ng-ai:
Reviewed and pushed one fix onto this branch.

**Detection missed the documented entry point.** The npm package installs two binaries, and `qoder` — not `qodercli` — is the one the docs tell people to run. `qoder` is a dispatcher: a bare invocation, a leading flag, or a prompt all go to the CLI, and only `ide`/`chat`/`serve-web`/`tunnel` or an existing path hand off to the IDE. Both binaries are `#!/usr/bin/env node` scripts, so what the pty carries is `node <path to the shim>`; the dispatcher's child, which is where `qodercli` shows up on the path, is not the process group leader and `foreground_argv` never reads it. Verified locally that a symlinked shim comes through as `["node", "/…/bin/qoder", …]`.

So `&["qodercli"]` detected nothing for a session started the documented way — no avatar, no entry in `tty7 agents`, no launch argv for resume or fork. The alias list is now `&["qoder", "qodercli"]`, with a test covering both spellings, bare and node-wrapped. The cost is that an IDE launch wears the CLI's avatar for as long as the launcher takes to exit.

Two smaller things in the same commit: `--session-id` restores a session rather than naming a new one, so the comment where it is stripped now says that (stripping it is still right — it collides with the appended `--resume <id>`); and the `Notification` step in the elicitation test went through a hook map with no seat for it, so it asserted nothing. It now asserts the absence directly, which is the real property — Qoder says outright when it is blocked, so a non-blocking alert must not be able to strand the pane on "waiting".

Everything else checked out against the docs: the eight hook event names all exist, `SessionStart` really does carry `source: "compact"`, `~/.qoder/settings.json` is the user-level file, `QODER_CONFIG_DIR` is a real environment variable, and `-w` is `--cwd`. `cargo test -p tty7-core --lib` passes, 1314 tests.

One thing I could not verify without running Qoder, left as-is: `--remote`, `--remote-session`, `--teleport` and `--remote-control` are documented as mutually exclusive with `--resume`, and they are not in the strip list. A pane started with one of them would get a resume command that Qoder rejects. Those sessions live in the cloud, so the right answer may be to offer no resume at all rather than to strip the flag — worth a look when you have one to test against.

Thanks for the integration.

https://claude.ai/code/session_01LAqfzqELnoDWU56LBXS1Nh

PR #900, opened 2026-09-19
fix: prevent blocking link operations and improve agent interactions

wsyangzy:
File-link detection can block the UI on slow filesystems, and opening
  remote localhost URLs can wait synchronously for port forwarding.

  This change moves file probes and loopback forwarding off the UI thread,
  preserves the clicked target across asynchronous operations, deduplicates
  pending requests, and adds read/write timeouts to forwarding control requests.

  It also adds independent agent notification settings, deduplicates terminal
  focus reports, and improves project-based tab labels.

  Validation:
  - 52 targeted application tests passed.
  - 11 targeted core tests passed.
  - Formatting and diff checks passed.

  The previously reported intermittent scrolling freeze has not yet been
  reproduced and verified with a newly packaged application.


l0ng-ai:
Thanks for this — the async rework is solid and the gpui coverage around the new deferred paths (`settle_link_opens`, `settle_link_probe`, the forwarder dedup test) is genuinely good. CI is green and the branch is current with `main`. A few things I'd like resolved before this goes in.

## 1. `explicit_file_candidate`'s dot rule is too broad

`src/terminal/view.rs`:

```rust
candidate.looks_like_a_path(style) || candidate.line.is_some() || candidate.path.contains('.')
```

`looks_like_a_path` only accepts `~`, `/`, or `\`. Adding "contains a dot" promotes any dotted token in terminal output to an explicit file candidate: `3.14`, `v1.2.3`, `e.g.`, `1.5s`.

Two visible consequences:

- Modifier-clicking such a token now starts a host `stat` and, on the miss, `report_unresolved_link` pushes a notification along the lines of "`3.14` is not under `/some/cwd`". On `main` this returned `false` silently and the terminal kept the click.
- Right-clicking one grows the whole file block in the context menu (label `3.14`, Open Link, Reveal, Copy Link Path), because `record_menu_link` now also latches `LinkAt::Unresolved { pending: true }`. And it only does this while the probe cache is cold — once the `stat` lands as a miss, the entries disappear. The same word gives two different menus depending on cache state.

The goal is clearly to make a bare `notes.md` clickable, which is worth having. Could this be narrowed to something like "a dot followed by a plausible extension" rather than any dot anywhere in the token?

## 2. First click on an extension-less filename can be swallowed

`Makefile`, `Dockerfile`, `LICENSE` have neither a separator nor a dot, so they are not explicit candidates, and `open_link_at` falls back to requiring an already-cached hit. But `ANSWER_TTL` is 2s and `expire()` only runs inside `hover_link_at`, which is only reached on mouse-move. So: rest the pointer on `Makefile` for more than two seconds, modifier-click — nothing happens; the second click opens it.

Pressing the modifier does re-arm the hover and kick off a fresh probe, so in practice that usually wins the race against mouse-down. But it is a race rather than a guarantee, and it is new — local panes resolved inline via `local_probe` before. Is the 2s TTL load-bearing for the network-share case, or could it be considerably longer?

## 3. The tab-label change belongs in its own PR

`project_name()` in `tab_strip.rs` turns `TabLabel::Cwd` into the last path segment: `~/repo/tty7` becomes `tty7`, `.../crates/tty7-core/src` becomes `src`.

The stated rationale is that the project directory distinguishes tabs running the same agent, but a bare basename does the opposite: `~/a/src` and `~/b/src` are now the same label. The diff acknowledges this by deleting the previous "a tab named after a directory says nothing more on hover" rule and restoring a tooltip on every cwd tab — which is the admission that the label alone no longer identifies the tab.

This changes what every user's tab strip says, and it is unrelated to either of the things the PR title names. I'd rather discuss it on its own, with the tab-strip behaviour as the subject instead of a clause at the end of a link-responsiveness PR.

## 4. The single-batch gate in `take_wanted` can wedge link detection for good

`src/terminal/link_probe.rs`:

```rust
pub fn take_wanted(&mut self) -> Vec<PathBuf> {
    if !self.in_flight.is_empty() {
        return Vec::new();
    }
    ...
}
```

`in_flight` is only cleared by `land()` and `retarget()`. `HostOps::run` skips its land callback entirely when `off_thread` yields `None` (the pooled job panicking, for instance), and the new epoch/generation guards inside the callback also return before `land()`. If that ever happens, every subsequent batch is empty, every probe stays `Unknown`, and link detection in that pane is silently dead for its lifetime — with no log line.

Low probability, high blast radius, and cheap to defend: clear `in_flight` on the guard's early return, and/or give in-flight entries a deadline after which they are reclaimed.

## 5. Smaller things

- `open_url_with_forwarder` checks `loopback.is_none()` and then does `loopback.expect("a forwardable loopback URL")` a few lines later. `let Some(loopback) = ... else { cx.open_url(url); return; }` collapses both.
- Hitting the caps (`pending_file_links.len() >= 8`, `pending_loopback_urls.len() >= 8`) silently discards the click — and `resolve_file_link` returns `true`, claiming it was handled. At minimum this deserves a `log::debug!`.
- `FileLinkAction::Reveal => {}` is unreachable: the menu row is `disabled` on a non-local host, and gpui-component neither attaches a click handler to nor allows keyboard selection of a disabled item. Harmless as defence, but if it were ever reached the user gets total silence; a `warn_file_open_failed`-style path would be better than `{}`.
- The commit message is in Chinese. Repo convention is English for commit messages (the PR body here is already English, thanks).

## Scope

This is one commit carrying six independent changes: link async-ification, the agent-notification setting, focus-report deduplication, the tab-label policy, shell/agent title filtering, and the `Notification` state-machine change in core. Items 1 and 3 are what I'd like changed; 2 and 4 I'd like your read on. The notification setting and the focus dedup look right to me on their own and I'm happy to take those as-is.

